PR #1044, opened 2014-07-17
Ability to drag while the camera is moving.

jeroenverfallie:
Right now  when you drag a sprite, and the camera moves, the position gets messy.

For example, if your camera moves when you're near the edge, now you can do that while dragging a sprite.

However, a drag update still requires a mousemove, so there probably should be an update when the camera moves.. I didn't make this change yet as I'm not sure on the best approach.

What I did, might not have been the best approach either, so please review it.. its a small change.

### DIFF
--- a/src/input/InputHandler.js
+++ b/src/input/InputHandler.js
@@ -1012,12 +1012,12 @@ Phaser.InputHandler.prototype = {
         {
             if (this.allowHorizontalDrag)
             {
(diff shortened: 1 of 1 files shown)


fyyyyy:
:+1: thanks

photonstorm:
Thanks for submitting this PR (even if it was years ago!). We have fixed this, and the fix has been pushed to the `dev` branch.

PR #1249, opened 2014-10-25
Add basic positional audio to the sound system.

woutercommandeur:
Use: game.sound.listener = someSprite; to set the listening position to follow that sprite. (Usually your player character). And then use someSound.following = otherSprite; to make the sound follow that sprite. set following to false to stop the sound following a sprite.

Even if this is not the way to go, it might serve as an example on how to approach this :)


photonstorm:
I'm really sorry but I'm not going to merge this in. I like the feature it provides, but the way it's hooked to velocity properties and has hard coded values make it a bit too restrictive. I do also wonder if it's the wrong way around - i.e. I think I prefer the concept of attaching a sound to a sprite (rather than a sprite to a sound) and being able to control the degree of the pan, i.e. screen based, world based, etc.

This has given me lots to think about though. I'm going to close it off so we don't have the PR hanging around (from your side or mine).

Issue #1355, opened 2014-11-23
Safari's bizarre caching of CORS resources breaks preloading


toolness:
The following "game" just presents a one-eyed fly blinking, with a "blop" sound playing once per second. The spritesheet and sound are retrieved from a separate domain with CORS enabled.

```html
<!DOCTYPE html>
<meta charset="utf-8">
<title>Safari CORS Lameness Example</title>
<div id="phaser-holder"></div>
<script src="//cdnjs.cloudflare.com/ajax/libs/phaser/2.1.3/phaser.min.js"></script>
<script>
var game = new Phaser.Game(288, 216, Phaser.CANVAS, 'phaser-holder', {
  preload: function() {
    game.load.crossOrigin = "anonymous";
    game.load.spritesheet("fly", "https://s3.amazonaws.com/safari-cors-lameness/fly-flying.png", 80, 92);
    game.load.audio("blop", "https://s3.amazonaws.com/safari-cors-lameness/Blop-Mark_DiAngelo.mp3");
  },
  create: function() {
    var fly = game.add.sprite(120, 57, "fly");
    var blop = game.add.audio("blop");

    fly.animations.add("flying", [0,1,2,3,4,5,6,7,8,9], 10, true);
    fly.animations.play("flying");
    game.stage.backgroundColor = 0xf0f0f0;
    game.time.events.loop(1000, blop.play.bind(blop));
  },
});
</script>
```

The above code, which you can try out live at http://jsbin.com/yolola/2/, works fine on Firefox 33, Internet Explorer 11, and Chrome 38. It also works on Safari 8.0 **for the first page load**, after which subsequent page refreshes fail. When it fails, the following appears in the Safari developer console:

```
[Error] Cross-origin image load denied by Cross-Origin Resource Sharing policy.
[Warning] Phaser.Loader error loading file: fly from URL https://s3.amazonaws.com/safari-cors-lameness/fly-flying.png (phaser.min.js, line 16)
[Error] XMLHttpRequest cannot load https://s3.amazonaws.com/safari-cors-lameness/Blop-Mark_DiAngelo.mp3. Origin http://jsbin.com is not allowed by Access-Control-Allow-Origin.
[Warning] Phaser.Loader error loading file: blop from URL https://s3.amazonaws.com/safari-cors-lameness/Blop-Mark_DiAngelo.mp3
```

Viewing the page successfully again requires manually clearing Safari's cache.

[...]

Issue #1330, opened 2014-11-15
Loader Async


austinh:
Right now I believe the Loader loads asset synchronously. It waits for .onFileComplete() and then calls .nextFile().

It would be more performant if the loader queued all the files to be loaded asynchronously. This would be great for loading a lot of assets, especially for web servers utilizing SPDY for fast async downloads.

I think this can be done without changing any API or needing to wait for 3.0, yes?

And then for 3.0, as a feature request, loading via XHR2, for file progress support for more accurate loading bars, with a setback to tag loading (I believe this is already on the roadmap)

photonstorm:
Closing as it's now in 2.3 :)

PR #1356, opened 2014-11-24
Loader - parallel / extensible / documentation

pnstickne:
Updates to the Loader to support parallel downloads (disabled by default) as well as adding future extensibility points with a pack/file unified filelist and an inflight queue. General documentation cleanup and jsdoc refinement is also included.

There are no known incompatibilities with the current version; all local testing (including emulated Android) "pass". With parallel downloading enable the order of the events may vary, which changes the results of the "Load Events" demo.

Parallel downloading (ref. https://github.com/photonstorm/phaser/issues/1330) can be enabled with `enableParallel` and the concurrency limit is controlled through `maxParallelDownloads`. Under simulated slower network connections parallel loading was a good bit faster than sequential loading. Even under a direct localhost connection parallel loading was never slower, but benefited most when loading many small assets (loading large assets is more limited by bandwidth throughput); both results are fairly expected.

Synchronization points -  an asset marked as a synchronization point must be loaded (or fail to load) before any subsequent assets can be loaded - can be added using the `withSyncPoint` and `addSyncPoint` methods. Packs ('packfile' files) and Scripts ('script' files) are treated as synchronization points by default. This allows parallel downloads in general while allowing synchronization of select resources if required (packs, and potentially other assets in the future, can load-around synchronization points if they are written to delay final 'loading').

Structurally the same process is used, with packs being merged into the filelist (as a 'packfile') format and the actual file objects, instead of indices, being threaded throughout the process. The separate state-loops have been combined into a single `processLoadQueue` but otherwise maintains the same flow and logic as previous.

Additional error handling / guards have been added, and the reported error message has been made more consistent. Invalid XML (when loading) no longer throws an exception but fails the particular file/asset that was being loaded. (There is still no timeout support, which might be a good "fallback error" to have.)

Some public methods/properties have been marked as protected, but no (except in case of a should-have-been-private-method) public-facing interfaces have been removed. Some private methods have been renamed and/or removed.

A new XHR object is created for each relevant asset (as there must be a different XHR for each asset loaded in parallel). Online searches indicated that there was no relevant benefit of XHR (as a particular use-case) re-use; and time will be dominated with the resource fetch. With the new flight queue an XHR cache could be re-added, at the cost of some complexity.

The URL is always transformed through `transformUrl`, which can make adding some one-off special cases like https://github.com/photonstorm/phaser/issues/1355 easier to deal with.

This also incorporates the fast-cache path for Images tags that can _greatly_ speed up the responsiveness of image loading - this uses the same "complete" and "width/height" check as used by PIXI, only it also orders the events to not suffer from PIXIs race condition (https://github.com/GoodBoyDigital/pixi.js/pull/1165). As such this should likely be treated as mutually exclusive with the other PR.

Feedback welcome!


pnstickne:
@photonstorm 
Bump. Updated for clean merge with dev and a few fixes. Also enabled parallel by default - which could be changed back although except for a few potential "ordering" expectations parallel / unordered resource fetching is likely better all around.

artyprog:
Unsubcribe

src/loader/Loader.js
```diff
@@ -988,440 +1118,658 @@ Phaser.Loader.prototype = {
...
-            extension = extension.substr((Math.max(0, extension.lastIndexOf(".")) || Infinity) + 1);
+                if (url.indexOf("?") >= 0) // Remove query from URL
+                {
+                    url = url.substr(0, url.indexOf("?"));
+                }
 
-            if (extension.indexOf("?") >= 0)
-            {
-                extension = extension.substr(0, extension.indexOf("?"));
+                var extension = url.substr((Math.max(0, extension.lastIndexOf(".")) || Infinity) + 1);
```
nickryall: @pnstickne  This line here will throw an error 'extension is undefined'.  I'm not sure it it's correct by I fixed it on my local copy with: 

var extension = url.substr((Math.max(0, url.lastIndexOf(".")) || Infinity) + 1);

photonstorm:
I've got the perfect project to test this on! Let's see what happens.

photonstorm:
So far, so good! Will need some wider community testing, but looking great in our current client projects at least (which as massively heavy on the loader).

PR #1430, opened 2014-12-09
Tweens - delay/repeat/yoyo relations

pnstickne:
There are some breaking changes, but such changes should be mostly
confined to case with "unexpected" or "erroring" behavior.

None of the changes apply when an index is explicitly applied (to
`delay/repeat/yoyo`), as might have been done in existing complex cases.

- `delay` applies to the _next_ child and `repeat/yoyo` apply to the
  _previous_ child, if any.  Warnings are emited if methods are used in an
  invalid pattern, consistent with attempting to add a child after the
  tween has started.

- As a special case, if `delay` is specified after all children it is
  applied to the first child; a warning will be emitted. This is a
  special-case for backward compatibility with "work abouts".

- `loop` renamed to `loopAll` (a proxy method is kept for backward
  compatibility, and both methods remain in the documentation, so this is
  not a breaking change); this is for uniformity with `repeatAll` and to
  avoid confusion with `repeat/yoyo`.

- `yoyo` defaults to true and `repeat` defaults to 1; for uniformity with
  `loop` and a positive-fluent design. This is a minor breaking change; it
  will only affect code that assumed `yoyo()` to act like `yoyo(false)` or
  `repeat()` to act like `repeat(0)` - which was already invalid code for
  not specifying a required parameter.

- Updated documentation and intra-links.

### DIFF
--- a/src/tween/Tween.js
+++ b/src/tween/Tween.js
@@ -7,9 +7,12 @@
 /**
 * A Tween allows you to alter one or more properties of a target object over a defined period of time.
 * This can be used for things such as alpha fading Sprites, scaling them or motion.
-* Use `Tween.to` or `Tween.from` to set-up the tween values. You can create multiple tweens on the same object
+*
+* Use {@link #to} or {@link #from} to set-up the tween values. You can create multiple tweens on the same object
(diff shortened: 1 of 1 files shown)


photonstorm:
Sorry this didn't ever get merged, but I'm closing it off to tidy-up outstanding PRs.

Issue #1377, opened 2014-11-28
phaser v2.2.0-RC12 camerafollow bug with arcade physics (sprite jitters)


jpcloud:
Hi Phaser-Team,

I am testing phaser v2.2.0-RC12 (Built: Thu Nov 27 2014 21:10:45) and I am experiencing the following problem with camera follow:

Set the camera to follow a sprite, move this sprite via setting it's arcade body velocity.

Expected behavior:
As long as the cam has not reached any bounds I would expect the sprite to be absolutely still in the middle of the screen.

what actually happens:
the sprite around jitters by a pixel

jsfiddle: 
http://jsfiddle.net/jppresents/ab4r7ekv/

(hold cursor right to see the effect, there is no actual sprite, but the green-error-box jitters instead of beeing still).

also: if you move the sprite by setting x or y every frame, then there is no jitter.

I would consider this bug critical, because it makes camera follow with arcade physics unusable.

Another user has found this problem too in a phaser 2.x version:
http://www.html5gamedevs.com/topic/10377-bodyvelocity-camerafollow-is-choppy/
(unresolved, only workarounds posted)

photonstorm:
I've spent some time looking into this today - and the solution proposed in the PR doesn't actually solve the issue. Pixi *by default* works on sub-pixel boundaries when rendering. It does not compensate for this, and setting `camera.roundPx = true` won't make any difference in the jsfiddle above, because the Sprite itself is still at a non-integer position.

The solution is to tell Pixi to not render sub-pixels. This only needs doing once in the renderSession, and will stop all the 'jiggling' going on without having to round any values internally (physics, camera or otherwise):

`game.renderer.renderSession.roundPixels = true`

Add that line to the jsfiddle and notice the difference! I'm going to close this issue off as the above should be the way this is handled. I'm half-tempted to turn it on by default, but that's a considerable change and doesn't suite all art styles, so maybe not for now. I will however mention it in the Camera part of the docs.

jpcloud:
thanks for looking into it and providing a solution :)

hocuspucus:
We started a game in Phaser two years ago and had to leave the project because this. It's a shame that two years later nothing changed at all, even with `game.renderer.renderSession.roundPixels = true;` Check this out: http://jsfiddle.net/7LLqnxw1/16/

Issue #1141, opened 2014-08-26
camera follow judders if target has non integer position


beeglebug:
If you set the camera to follow a sprite which is moving via velocity or some other method which causes it's position to end a tick on a non integer position, the camera will look very juddery.

This is because of the line `this.view.floor()` in the camera checkBounds method (https://github.com/photonstorm/phaser/blob/a6bc859246ca6cab7cd9fd06a8055a99fb6ab8e0/src/core/Camera.js#L308), which always snaps the camera position to integers. 

If I comment out that line my game suddenly looks a lot better, but i'm hesitant to do a pull request because i'm not sure of the wider ramifications.

PR #1444, opened 2014-12-12
Update Camera.js

RemcoTukker:
Set default Camera.roundPx value to false, in order to prevent almost-1-px jitter  (Pixi does subpixel rendering on (almost?) everything else in Phaser, so rounding here is not okay). 

@ next cleanup: remove this bool completely if there's no problems with the new default value.

This should solve issue #1377  and is how issue #1141 was supposed to be solved

Only caveat: this may have undesired side effects in some situations, as I dont know why this rounding was added in the first place..


photonstorm:
See my reply in #1377

PR #1447, opened 2014-12-12
Local AudioSprite JSON data object support

jounii:
Tested with modified audio sprite phaser-examples example. If ok, modified example could be another PR


photonstorm:
Thanks for this! I have implemented my own take on providing this feature which is now in dev, but thanks for the suggestion and example json.

PR #1470, opened 2014-12-18
Wrapped play() calls to "fix" IE Windows Phone random crashes

jounii:
Wrapped play() calls to "fix" IE Windows Phone random "Unexpected call or property access" crashes.

The play() calls may fail even if using proper audio formats and html !DOCTYPE html.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,7 @@ Android browser does not support Full Screen.
 * The RandomDataGenerator could be seeded with an array of values. However if the array contained a zero it would stop seeding from that point (thanks @jpcloud @pnstickne #1456)
 * Added extra checks to Sound.play to stop it throwing DOM Exception Error 11 if the `sound.readyState` wasn't set or the sound was invalid. Also wrapped `stop()`` call in a `try catch`.
 * Time.reset would incorrectly reset the `_started` property, now maps it to `Time.time` (thanks @XekeDeath #1467)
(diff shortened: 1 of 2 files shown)


photonstorm:
Oh God, this is horrible. Trying to think of another way to handle this!

jounii:
I tried quick patch using the pendingPlayback (afair) but couldn't get it work straight away, as it wasn't ment to fix this kind of issues. Right now my solution is to disable audio with windows phone IE and/or if nothing else, have third layer of audio abstraction to not to use audiosprites but limited set of individual sounds with WP.

I created abstraction audio manager to overcome audiosprite issues with IE. I categorize audio to named groups and group can consists any number of audiosprites. Interface a bit similar to tweens to configure playback parameters and choose named sound/marker which is then resolved to correct audiosprite instance. On Webaudio you can have one audiosprite, but on IE you have group sounds so that concurrency playback is spread across different audiosprites. Using multiple instances of same audiosprite caused tons of problems (IE doesn't cache loads, etc).

photonstorm:
I'm really sorry, but as this is such a super edge-case, yet the change impacts performance for *everyone*, I think I'm going to have to insist it's handled in user-land code (i.e. over-write the Sound.play call with your own prototype replacement perhaps). Sorry, I wish I could merge it, but it just feels 'wrong' :(

hilts-vaughan:
Maybe a documentation note should be made, at least? We use a custom
version of Phaser with a try catch baked in for some our games. :)

Issue #1479, opened 2014-12-22
Recent updateTransform optimisation causes glitch


spayton:
I have tracked down a visual glitch to this update 028943baad477fa81ce5d446e9e47119832c4bc4 and may be relevant to this issue #1424 

I was seeing sprites momentarily appear at the wrong position after setting visible=true despite also setting the correct xy at the same time.  I noticed this was only happening on low power devices that had to skip frames to keep up.  I also noticed that the wrong position was the last position the sprite had before visible was set to false.

The updateTransform optimisation added a _updateTransform bool, which gets set after the 1st update and this is then used to skip further transform updates within this update.  the problem occurs on a slow device that is skipping frames on a subsequent state.update(), i made a sprite visible and set its position, the position transform is not updated but its visibility is picked up by the renderer - thus drawn at the wrong position.

As it is, this optimisation is unsafe, it needs some more work or removing.

photonstorm:
Addressed in the latest merged changes.

PR #1493, opened 2014-12-30
Run updateTransform() to update world coordinates after arcade physics update

kabuhr:
Alternate patch for issue #1479.  Note that this is an *alternative* to the 028943ba branch, *not* to be merged with it.


pnstickne:
@kabuhr The problem is "eventually" is not after each logic update (which means that two logic updates called in a row will produce incorrect results if the scene graph is not "synced" between them and objects are moved).

Issue #1439, opened 2014-12-10
TileMaps Broken in Safari (Phaser 2.2.0/1)


neurofuzzy:
TileMaps are very broken in Safari, and it seems to be due to 1 line of code which, if I comment it out, makes everything work fine. I'd create a PR but I do not know exactly how this particular rendering flag is used:
 
The line is here:
 
https://github.com/photonstorm/phaser/blob/master/src/tilemap/TilemapLayer.js#L677
 
In method: `Phaser.TilemapLayer.prototype.shiftCanvas`
    
    context.globalCompositeOperation = 'copy';

Example:
 
http://jsfiddle.net/zq8hn23g/6/
 
This works fine in all browsers except Safari. If you run the example, then click and drag on the canvas, you'll see the tilemap flicker in and out of visibility.
 
Please let me know if you need any more info.

photonstorm:
Hmm I've merged in the latest PR but I now see the following happen in all tilemap examples:

[map screengrab](https://www.dropbox.com/s/fqz5ghzcj44342a/maps.png)

Maybe something didn't merge correctly? This happens in Chrome, FF, etc.

DanielHudson:
enableScrollDelta = true fixes that in Safari and Chrome

photonstorm:
If that's the case it probably needs to be the default setting. Which I'm assuming will bork the optimisation it introduced.

photonstorm:
Actually I've just checked and `enableScrollDelta` is already enabled by default and it still happens in Chrome (on Windows, not tested OS X yet).

photonstorm:
Setting it to `false` fixes the issue. The property was there and set to `true` as default in 2.2.1 though and didn't exhibit this issue, so it's a regression most likely from the Safari double-copy fix.

pnstickne:
@photonstorm Does `layer.renderSettings.copyCanvas` contain a DOM canvas element?

pnstickne:
@photonstorm That is bad - it is caused because the previous area is not cleared, and it "smells" like something got merged bonkers. I'll take a look at it when I get home tonight.

Since it is Chrome on windows, that should be using the "direct" method. Looking at `theLayer.renderSettings.copyCanvas` will be undefined if it is a direct bitblt.

photonstorm:
Yes `copyCanvas` is undefined, so it's definitely using the bitblt code path. Even so, in Chrome it appears as in the screen shot posted above.

[...]

PR #1498, opened 2014-12-31
Fix / double-copy for Safari tilemap bug when rendering with delta scrolling

pnstickne:
Should fix https://github.com/photonstorm/phaser/issues/1439#issuecomment-68322244 - with automatic selection based on browser detection.


photonstorm:
Rather than ua sniffing this should be quite simple to test for by creating a canvas, drawing, performing the copy onto itself and checking the resulting pixel values I'd have thought?

pnstickne:
@photonstorm That is indeed the "correct" approach. I'm just shy on a device on hammer out the exact details of the failing case.

photonstorm:
Sure no worries, I'll add the check in.

pnstickne:
That update should make it default to a double-copy *except* when using a known-to-work browser not on iOS.

A proper feature check would be better, but I'm not entirely certain of the _exact_ minimal reproduction case that I can verify/test locally.

Issue #1496, opened 2014-12-31
Device.windowsPhone not detected with Windows Phone 8.1


jounii:
Device.windowsPhone is set in "else if" path after Android (and other) checks, however this causes atleast "Windows Phone 8.1" IE11 to be detected as Android + iPhone. IE team had it's reasons to lie like this, but obviously not exactly correct if you want ie. to disable something on Windows Phone device what you want to be enabled in iOS device not matter how much IE team wants to believe their browser is like iPhone.

This monster comes up with IE11 in Windows Phone:

"Mozilla/5.0 (Mobile; Windows Phone 8.1; Android 4.0; ARM; Trident/7.0; Touch; rv:11.0; IEMobile/11.0; NOKIA; Lumia 925) like iPhone OS 7_0_3 Mac OS X AppleWebKit/537 (KHTML, like Gecko) Mobile Safari/537"

http://msdn.microsoft.com/en-us/library/ie/hh869301(v=vs.85).aspx

pnstickne:
Well, that's a can of worms..

pnstickne:
YMMV with the PR.
Tests and TLC / updates welcome!

enriqueto:
I use to detect windows phone:

 if(navigator.userAgent.match(/iemobile/i)){ ....}

photonstorm:
Ok that should now work properly - but it'd be good if someone could test!

PR #1503, opened 2015-01-02
UA updates for lies from IE "edge" versions

pnstickne:
These updates are _not tested_ and could lead to regression cases if
another browser has claims to be "Edge" or "Trident" or "IEMobile". YMMV.

The support for plain Windows (not NT/Phone) has also been removed.

Ref. https://github.com/photonstorm/phaser/issues/1496


pnstickne:
(I want to stress this is _not tested_ and needs some TLC / review.)

photonstorm:
Closing off old PRs that are now fixed, no longer required, locally merged or were just for testing.

PR #1506, opened 2015-01-02
Fixed alpha values for generated canvas styles

wayfu:
fillStyle/strokeStyle colors in "rgba(r, g, b, a)" format take alpha as
a value between 0 and 1. A few utility functions created strings with
alpha values between 0 and 255.

### DIFF
--- a/src/utils/Color.js
+++ b/src/utils/Color.js
@@ -78,7 +78,7 @@ Phaser.Color = {
         }
 
         out.color = rgba;
-        out.rgba = 'rgba(' + out.r + ',' + out.g + ',' + out.b + ',' + (out.a / 255) + ')';
+        Phaser.Color.updateColor(out);
 
         if (hsl)
         {
@@ -117,7 +117,7 @@ Phaser.Color = {
         out.b = ((rgba & 0x0000ff00) >>> 8);
         out.a = ((rgba & 0x000000ff));
 
-        out.rgba = 'rgba(' + out.r + ',' + out.g + ',' + out.b + ',' + (out.a / 255) + ')';
+        Phaser.Color.updateColor(out);
 
         return out;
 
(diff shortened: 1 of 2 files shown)


src/utils/Color.js
```diff
@@ -441,7 +441,7 @@ Phaser.Color = {
 
         var out = { r: r || 0, g: g || 0, b: b || 0, a: a || 1, h: h || 0, s: s || 0, l: l || 0, v: v || 0, color: 0 };
 
-        out.rgba = 'rgba(' + out.r + ',' + out.g + ',' + out.b + ',' + out.a + ')';
+        out.rgba = 'rgba(' + out.r + ',' + out.g + ',' + out.b + ',' + (out.a / 255) + ')';
```
pnstickne: What about `Phaser.Utils.Color.updateColor(out)`?

wayfu:
My gut feeling is that the internal color.a value should always be 0..255 and that it should only be converted to 0..1 for the rgba() string. So perhaps it is best to change createColor to switch to 0..255 for the internal value as well. 
The rgba property of a color, as far as I can tell, isn't used even once in Phaser itself. 
updateColor takes a color object, not raw values. I'm hesitant to make an API breaking change here, and equally hesitant to introduce a new function.

pnstickne:
@wayfu One of the problems is "a color object" isn't well-defined.

My first step when approaching a problem like this is first updating the documentation to reflect the current behavior. This may defining what "a color object" is better in context.

Then the current code can be compared vs the documented (as in what the function does) behavior - does the current code hold? If so, then don't change the function in a way incompatible with the documentation (and this may be "accepts 0..1 or 0..255" but you'll have to carefully identify it).

I believe `rgba` is primarily (only) used in a few cases, such as for background colors (eg. when the color setting bleeds down to pixi.js). More generally it *may* be useful for CSS - although that's outside of core.

wayfu:
The only places Phaser uses the rgba property of a color object is to pass
colors to a callback in bitmapData per-pixel functions. These use rgba(r,
g, b, a) with a = [0..1]. My changes do not break anything there.

Inconsistencies in the Color set of methods can be explained by different
authors. Some functions were clearly written as helpers for the specific
bitmapData.processPixel use case. There, the explicit purpose of providing
that string is that it will be useful for styles.

My PR updates the documentation.

2015-01-05 5:25 GMT+01:00 pnstickne <notifications@github.com>:

PR #1507, opened 2015-01-02
Fix small typescript definition issue with Emitter.setAlpha()

Schmavery:
The setAlpha() method accepted type "number" for the ease parameter, which obviously didn't work when passing in an easing function.  A pretty small change, but I figured it couldn't hurt.


photonstorm:
:+1:

PR #1511, opened 2015-01-04
Mark all easing methods as static.

johnfn:
Hi there! The docs were treating these methods as non-static, but they definitely are static (as far as I can tell...) so I went in and cleaned that up. Thanks for the fantastic library!


johnfn:
Oops, I totally messed up the Github PR interface. Closing and trying again.

PR #1512, opened 2015-01-04
Add documentation to Typescript definition file. 

johnfn:
Accomplished via [this python script](https://gist.github.com/johnfn/d7907fd100261beda045) I wrote. I didn't think it made sense to include it in the PR. 

This adds documentation to most of the functions in the phaser.d.ts file, so that when you mouse over keywords or use autocomplete in an IDE that supports it, you can now immediately see the Phaser docs. It's pretty handy! 

There are currently a few minor problems worth considering in the future:
* I didn't do p2 or PIXI docs, so there is some missing documentation at this point. p2 and PIXI their comments in a way less conducive to scraping than Phaser does, so I figured I'd just do Phaser for now. p2 and PIXI are both possible - I might do it someday - but they are more annoying and the payoff is smaller. 
* I haven't tested it super-extensively. It's a 20,000 line diff, I might need a little help. :) 
* The @params, @returns etc look a little ugly in Visual Studio since there's no special handling for them. I don't think it's a huge issue, but if anyone has suggestions for better formatting, I'd love to hear it. 

Anyway, let me know what you think! 

P.S. I also changed easing.js to properly indicate that those methods are static. It was messing up my script ;-)


photonstorm:
Interesting :) I can't merge this for a couple of reasons, but I like the idea of it:

1) It's against master, not dev. PRs go against dev.

2) This is a bit too experimental to totally replace the TS defs at this stage. Also the docs are changing constantly, so we need the Python script as part of the jsdoc generation process otherwise this becomes a little useless the moment we update the docs. Also I'm worried it'll mess-up what @clark-stevenson works so hard to maintain.

In that light I think this needs to generate a new defs file, not replace the current one (maybe phaser-docs.d.ts?) and we need the script merged into our workflow somehow.

clark-stevenson:
This is cool @johnfn  :+1: 

It is a bit over my head this ticket so just to add my 2 cents. 

From the point of view of the master branch, this is awesome because that is an additive process to something which is never maintained. 

But if it added 20k+ lines to the phaser.d.ts file on dev, that is a different context which I would avoid as it essentially doubles maintenance times.

johnfn:
Thanks @photonstorm, those changes seem easy enough. :) 

@clark-stevenson can you clarify what you mean by 'doubles maintenance times'? If I understand correctly, by @photonstorm's suggestion you could continue to maintain phaser.d.ts as usual, and when it's time for a release, you could run my little script and generate a phaser-docs.d.ts file alongside phaser.d.ts. 

Does that make sense, or am I missing something here? I don't fully understand the process by which these files are maintained.

clark-stevenson:
"You'd never have to deal with the 20k+ bloated file."  -  This is all I wanted to know and was my only concern. Great work!

photonstorm:
@johnfn any suggestions on how to get your script run as part of our jsdocs grunt task?

johnfn:
@photonstorm It should actually be super simple. I want to hook it up with some command line arguments to make your guys' life easier. I'll try to get that done over the weekend.

Bilge:
:+1: Do want this.

vulvulune:
I added a grunt task to generate automatically the .d.ts files with comments in this pull request: #1552

PR #1515, opened 2015-01-05
Game Object Unification - Request For Comments

pnstickne:
This is a WIP branch (not designed to be merged *just* yet) for the unification of Game Objects (Images, Sprites, Text, Graphics, etc.) which was one of the 2.3x milestone features.

It is designed to minimize any breaking changes and to not impose a performance impact. On Desktop Chrome it appears a few percents "faster" (although really it's the same speed) than the individualized implementations - at least for Sprites / Particles, of which there are likely the most instances.

The benefit of course, is consistent (one place) documentation and consistent (one place) implementations. Re-using the same functions also has allowed fixing some inconsistencies and applying tweaks universally - hundreds of lines of code and dozens of methods have been unified.

There are a few small changes to be made, but I think it's a pretty solid approach that doesn't sacrifice performance<sup>1</sup> or (if needed) specialization.

---
<sup>1</sup> I test performance in Chrome and other Windows browsers; mobile and/or non-JIT implementations may react differently, but care has been taken to make the unified functions and objects "well optimizable", if the JS engine is capable of doing so - *specific* performance issues, if found in these mobile environments, can be addressed as required.


pnstickne:
Here is the class description from GameObjects, for an overview:

```
/**
* @classdesc
* Game objects are {@link PIXI.DisplayObject display objects} with additional Phaser features.
*
* Because there is not a single prototype hierarchy a "GameObject" mixin system is used to 
* consistently add the appropriate game object features to different Phaser types.

* All game objects are expected to inherit from PIXI.DisplayObject or otherwise 
* implementation the methods and properties it defines.
*
* ---
*
* A mixin approach is used to avoid altering the prototype inheritance while providing maximum
* feature compatibility and code/documentation sharing between existing types.
*
* This approach has been designed for maximum code-resue and consistency with no/minimal performance
* impact; the shared functions and easy of applying general optimizations may even improve overall performance,
* even if the functions themselves are not specialized.
*
* Judicial guards - lifted into a prototype or instance as appropriate - are used to handle
* varied features and care is taken to not introduce extra work.
* (Support for various Traits can be triggered after an object is created by adding properties.)
*
* The following game object mixins are available:
*
* - {@link Phaser.GameObject.CoreMixin} - Included in every game object.
* - {@link Phaser.GameObject.CullingMixin}
* - {@link Phaser.GameObject.TextureMixin}
* - {@link Phaser.GameObject.InputMixin}
* - {@link Phaser.GameObject.EventsMixin}
* - {@link Phaser.GameObject.PhysicsMixin}
* - {@link Phaser.GameObject.LifeMixin}
*
* Using separate types also allows documentation generation and inheritance tracing. In JSDoc this
* is handled with the use of multuple @extends doclets.
*
* - Even though no MI is used, multiple @extends JSDoc doclets result in correct output that shows the
*   mixin member with the type and links to the inherited base.
*   (JSDoc 3.3 should include @interface/@implements and better @mixin support.)
* - Google Closure Compiler may currently support @interface/@implements.
*
* @class Phaser.GameObject
* @protected
*/

photonstorm:
Very nice! I was going to release the new version today, because we need to get the Safari tilemap fixes and updateTransform stuff into the wild asap. But after that I'll merge this into dev for a play.

This would be a good time to get rid of the _cache array as well. Array access being much slower than either single _cache vars or a _cache object in this instance. Also I'm sure that for `if` statement at the start of `preUpdate` could be sorted out too - iirc it looks like a transform sync, but I reckon there must be an easier way to do that now.

photonstorm:
I suspect this drops the overall file size down a bit too?

I'm tempted to deprecate the Rope class while we're at this. It's quite buggy (Pixi side) - so think it might be worth leaving out of the main build.

pnstickne:
It is indeed important to get out a point-release.

The update logic could probably be resorted. I haven't looked at it context of the world-sync stuff that has come up; the code is simply trying to unify the previous methods without adding regressions. Improvements or suggestions of such are always welcome.

I did a few tests with the cache array vs. using normal properties and it was on the fence in Chrome so I left it alone; but Chrome is magic all about. I really need to setup a cross-browser performance marks to be able to get some better numbers - especially on mobile environments - to make sure it's not just Best on Chrome (TM).

It will reduce the file and run-time sizes a tiny bit; I don't think it'll be a big overall change.

I've never used Rope, but it's easier to say "not supported" then "it's supported but doesn't work".

pnstickne:
Relevant changes appears to be in dev.

photonstorm:
Most of them yes, I haven't yet finished up the jsdocs side of things though, but am still very happy with the way things are now split out into components instead (which are no more than tiny mixins). Will make it far easier for devs to modify / include / exclude just key bits in their own object types.

PR #1521, opened 2015-01-07
Patch to fix a compatibility problem between Opal and Phaser for opal-phaser

ylluminarious:
I'm working on an Opal wrapper for Phaser, to allow folks to write Phaser games entirely from Ruby code, and I ran into a problem trying to replicate a Phaser Example. I was working with the `atlasJSONHash` loader method, and there was an error coming from AnimationParser.js, due to something that Opal wasn't doing quite right. Phaser was iterating through a `frames` object (for JSON frames), with `for (var key in frames)`, but as Opal had something inside the object, it tried to add a frame for the "OpalClass".

A friend of mine created an issue about this on the Opal repository: https://github.com/opal/opal/issues/680
Fortunately, the creator of Opal came up with a solution that fixed this problem. I was hoping my change could get merged with Phaser, as I don't see it causing any problems.

### DIFF
--- a/src/animation/AnimationParser.js
+++ b/src/animation/AnimationParser.js
@@ -190,39 +190,41 @@ Phaser.AnimationParser = {
 
         for (var key in frames)
         {
-            var uuid = game.rnd.uuid();
+            if (frames.hasOwnProperty(key)) {
+                var uuid = game.rnd.uuid();
 
-            newFrame = data.addFrame(new Phaser.Frame(
-                i,
-                frames[key].frame.x,
-                frames[key].frame.y,
-                frames[key].frame.w,
-                frames[key].frame.h,
-                key,
-                uuid
-            ));
+                newFrame = data.addFrame(new Phaser.Frame(
+                    i,
+                    frames[key].frame.x,
+                    frames[key].frame.y,
+                    frames[key].frame.w,
+                    frames[key].frame.h,
+                    key,
+                    uuid
+                ));
 
-            PIXI.TextureCache[uuid] = new PIXI.Texture(PIXI.BaseTextureCache[cacheKey], {
-                x: frames[key].frame.x,
-                y: frames[key].frame.y,
-                width: frames[key].frame.w,
-                height: frames[key].frame.h
-            });
+                PIXI.TextureCache[uuid] = new PIXI.Texture(PIXI.BaseTextureCache[cacheKey], {
+                    x: frames[key].frame.x,
+                    y: frames[key].frame.y,
+                    width: frames[key].frame.w,
+                    height: frames[key].frame.h
+                });
 
-            if (frames[key].trimmed)
-            {
-                newFrame.setTrim(
-                    frames[key].trimmed,
-                    frames[key].sourceSize.w,
-                    frames[key].sourceSize.h,
-                    frames[key].spriteSourceSize.x,
-                    frames[key].spriteSourceSize.y,
-                    frames[key].spriteSourceSize.w,
-                    frames[key].spriteSourceSize.h
-                );
+                if (frames[key].trimmed)
+                {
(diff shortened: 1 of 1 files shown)


pnstickne:
I would probably use `if (!frames.hasOwnProperty(key)) { continue; }` - and I don't use continue very often.

photonstorm:
A well formatted `atlasJSONHash` file should never have an array inside it. I'm curious where the Opal properties are coming from too - this is supposed to be an externally loaded json file from the likes of Texture Packer. It shouldn't have been munged by anything first. If the Opal property has been truly added to the object then `hasOwnProperty` won't skip it. If the json is coming from some deeper internal object is it picking it up from there?

gabrielrios:
@photonstorm I'm confused on this, how can it never have an array, isn't frame an array here https://github.com/photonstorm/phaser-examples/blob/master/examples/assets/sprites/running_bot.json ?

photonstorm:
@gabrielrios Phaser supports 2 texture atlas formats: JSON Hash and JSON Array. The one you linked is an array format. You'll find lots of hash format json files in the assets folder too. The hash format files don't use any arrays internally, which is what I was referring to in my comment.

If you try and load an array format json file using `atlasJSONHash` it will error big time (the opposite is also true!).

That is all I meant about the fact that if properties or arrays have been added somewhere into the json object then it must have been tampered with by something outside of Phasers loader. Is there a reason you need to parse the json before giving it to Phaser? (or aren't using Phasers Loader to ingest it in the first place). I don't know anything about Ruby/opal so am curious to learn.

ylluminarious:
Well, what you just said is interesting, @photonstorm. See, like I said in the initial post, I got this issue while replicating a Phaser Example. Specifically, this example: http://examples.phaser.io/_site/view_full.html?d=basics&f=05+-+load+an+animation.js&t=05%20-%20load%20an%20animation&jsbin=http://jsbin.com/zagob/10/edit?js,output
As you can see in the example, it's using an `atlasJSONHash` to load this JSON file, but the file that it's loading is the aforementioned https://github.com/photonstorm/phaser-examples/blob/master/examples/assets/sprites/running_bot.json.
I may be missing something, but doesn't that mean that this particular example is not using the correct method to load that JSON file?

gabrielrios:
@photonstorm ok, this explains a lot. So this file is indeed JSON Array, but we're trying to load as JSON Hash. 

Just to clarify, we're not parsing the JSON before giving to Phaser, we're using the Phaser Loader, but as its a Opal wrapper its javascript run inside of Opal context and so the Array get all those properties. 

But so I guess the issue is just loading a JSON Array file using the JSON Hash code.

photonstorm:
I guess the first thing to test then is if you load the file using the correct format, what happens? :) Does it still error?

gabrielrios:
Yes, you're right, using atlasJSONArray worked flawless. I'm just curious about the example @ylluminarious pointed that uses the atlasJSONHash to load that file and works ok. maybe we should change that example to reflect this? ie: use atlasJSONArray on the example instead of atlasJSONHash

photonstorm:
I'm guessing that if you say "this is a hash" but it's really an array, that 'for in' still works to iterate over it - to a degree. But the opposite breaks entirely, hence why the example appears to be ok, but should really be swapped for the other format.

ylluminarious:
Yeah, that's pretty much what's happening, @photonstorm. So, I'm also wondering the same thing as @gabrielrios: should we change that example to use `atlasJSONArray` instead of `atlasJSONHash`?

ylluminarious:
By the way, I made a pull request to change this example here: https://github.com/photonstorm/phaser-examples/pull/76

photonstorm:
Yeah, will merge it shortly. Going to close this specific one off for now, but feel free to re-open if you feel there's still something here :)

PR #1528, opened 2015-01-11
Bitmapdata

wayfu:
Adds a function to render a Sprite or Image exactly as it would be
rendered to the main game canvas, respecting scale, rotation and skew.

### DIFF
--- a/src/gameobjects/BitmapData.js
+++ b/src/gameobjects/BitmapData.js
@@ -879,10 +879,6 @@ Phaser.BitmapData.prototype = {
             source = this.game.cache.getImage(source);
         }
 
-        var src = source;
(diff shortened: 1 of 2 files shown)


wayfu:
I don't know why the earlier commits about Color.js are in there, I think github just hates me.

pnstickne:
I would remove the `instanceof` check.

If it isn't a correct type per the documentation it should explode. Or at least not silently "do nothing".

photonstorm:
I'm really sorry but I'm not going to merge this in. I feel it can be easily achieved with the existing `BitmapData.copy` method as long as the correct arguments are provided. Rather than leave the PR hanging I think it's best to close it off.

PR #1536, opened 2015-01-12
Fixed wheelDelta type on Typescript definition file

belohlavek:
Changed type `WheelEventProxy` to `number` for property `wheelDelta` in `phaser.d.ts` as specified in the Phaser Documentation: http://docs.phaser.io/Phaser.Mouse.html#wheelDelta


clark-stevenson:
:+1:

PR #1541, opened 2015-01-14
Add comments in the typescript definition

vulvulune:
The comments has been added with some scripts using docjs and
TypeScript.Parser.
The comments format is compatible with the autocompletion of Visual
Studio.


clark-stevenson:
Hi @valvalune  

Can you review this ticket here #1512  and let me know your thoughts?

vulvulune:
I didn't see there was already a pull request with almost the same work. :-(

Some differences between the johnfn job and mine:
- I did the job on the dev branch and not the master one.
- I had a comment on the class and on the constructor (it's the same, except I didn't add the params in the class comments.)
-There is also comments on the P2 classes.
- I didn't put the @method, @static in the comments, because the context already gives these information. (and these data are not managed by the Visual Studio IntelliSense.
- I didn't put the type of type in the param comments, because already in the method signature, and I added textually "- Default = xxx" at this end of the param comment to get the comment value.
-I have comments on the enum.
-I used jsdoc and the Parser class of TypeScript to parse the files, so like this the process is robust, and I support comments in the TS files for example, and use some of the plugin uses by jsdoc to get the best quality of data from js files.

I agree that we should add the automatic generation of the comments in the build process.

My scripts are usable, but not enough clean and structured to add them now in the build process.

Because @johnf has started the job first, I let him continue it and try to add his script in the build process.
If he don't want to continue, and the community is interessted I can add my scripts as a grunt task, and the idea a to have 2 .d.ts files (one with the comments, and one without) seems very good.

photonstorm:
I'd very much like to include either this or @johnf version but I absolutely need it as part of the build script, or it will be out the date the moment after I merge it :(

vulvulune:
OK, I will try to add my script in the build process and generate a copy of the .d.ts file with the comments.

vulvulune:
I closed this pull request (#1552) because I did a new one with a grunt task to generate the commented files.

PR #1548, opened 2015-01-18
adjust mass bug fix

tomlarkworthy:
calling adjust mass desyncs the debug graphics from the real positions of bodies


tomlarkworthy:
Oops wrong branch

PR #1549, opened 2015-01-18
Bug fix: adjust mass and debug body gfx 

tomlarkworthy:
calling adjust mass desyncs the debug graphics from the real positions of bodies


photonstorm:
Thanks! :+1:

PR #1559, opened 2015-01-22
Complete "tsdocs" tasks to download plugins + improve ts doc + grunt:jsdoc

vulvulune:
A little fix:
- If there was a comment formatted with "//" before a member name in the d.ts file, the property name was not extracted correctly to get the comments from jsdoc

-The most important part:
Complete the tsdocs task to download automatically the phaser-plugins (with gitclone) and scan their comments to add them in the phaser.comments.d.ts where the plugins classes are defined.
To have more comments from the plugins, the pull request 8 (photonstorm/phaser-plugins#8) on the project phaser-plugins should be merged.
I didn't put the updates of the phaser.comments.d.ts file in this pull request. Tell me if you want I include it ,but I think it's better to refresh it periodically.

-Another import thing:
 I replace builddoc and exportdocjson grunt tasks with grunt-jsdoc.
To avoid an error I added a fake comment to a @todo param in the p2.js file. I didn't find a way to configure grunt-jsdoc to ignore the warnings, and even with a look in the source code there is no option.


photonstorm:
:+1:

PR #1565, opened 2015-01-24
Fix CORS loading BitmapFonts with IE9

jeppester:
This fix enables phaser to load BitmapFonts under CORS-conditions with IE9 (when enabling useXDomainRequest).
Most of the code has been copy-pasted from the XML-load fix that was already in the code, but only used when loading XML-files directly.


photonstorm:
Nice one, thanks :+1:

PR #1570, opened 2015-01-25
Compact overloaded constructor using union syntax.

Bilge:


clark-stevenson:
One of the things here which I considered recently was that even though union types are added. What does that mean for an open source project?

Does TypeScript 1.4 become a requirement for using phaser?

I have no problem with it, but I just wonder if anyone else will?

Bilge:
TypeScript is never a requirement for using Phaser. Those who wish to use TypeScript have no compelling reason not to upgrade.

vulvulune:
My script to generate the comments.d.ts files supports the new unions introduced in typescript 1.4

I agree that TypeScript is still experimental and everybody use always the last version, and there is no reason to not upgrade from 1.3 to 1.4.

clark-stevenson:
This sounds good to me! I also agree and on 1.4 

However, If this is the case, then the webgl.d.ts should be removed as it has no relevance any more, along with a readme.md update to say that you must use 1.4

vulvulune:
Who want to update the readme.md file? I can do it if you want.

Bilge:
@vulvulune What are you talking about? What's wrong with the readme?

vulvulune:
@bildge specify that the version 1.4 is necessary. See the last message of Clark in this topic.

Bilge:
Oh sorry, I missed that message. That makes sense then :+1:

clark-stevenson:
Sorry guys, I have been afk for a few days.  Yeah @vulvulune that would be great!

PR #1571, opened 2015-01-25
Safari scroll delta final fix

pnstickne:
This fixes an issue with Safari 7-based implementations where the scroll delta would result in black regions at the bottom when scrolling up. It reverts some code that could - if only Safari worked correctly - reduce the secondary copy requirements.

The change increases the size of the copy canvas, but it is still shared between all layers.

This is *tested* on iOS 8 on an iPad mini 2 - and it works there. (It was only tested with one layer, in CANVAS mode, however.)

The performance is still much better than without scroll delta on this device; I am not sure how older devices compare - notably there may be more pixels copied overall, even if the total work done is less per pixel.

Ref. https://github.com/photonstorm/phaser/issues/1439#issuecomment-70591294

### DIFF
--- a/src/tilemap/TilemapLayer.js
+++ b/src/tilemap/TilemapLayer.js
@@ -938,7 +938,7 @@ Phaser.TilemapLayer.prototype.renderDeltaScroll = function (shiftX, shiftY) {
         this.context.clearRect(((left * tw) - scrollX), 0, (right - left + 1) * tw, renderH);
 
         var trueTop = Math.floor((0 + scrollY) / th);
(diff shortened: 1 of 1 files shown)


pnstickne:
Bah, I take tossing code. But the N-slice was specifically for Safari which is still busted with it. Removed.

pnstickne:
http://codepen.io/anon/pen/vEZdEQ demo (Phaser hosted here)

photonstorm:
Ok let's see what this does to Safari on OS X.

photonstorm:
Confirming this works on Safari on OS X with all the Phaser tilemap tests in both canvas and webgl. Phew! :)

PR #1576, opened 2015-01-27
Correct comments

vulvulune:
Correct comments:
-Phaser.Camera: checkWorldBounds =>checkBounds
-Phaser.RetroFont: Set correct @name for text and smoothed
-Phaser.DOM: inViewport => inLayoutViewport
-Phaser.ArrayUtils: Replace Phaser.Math => Phaser.ArrayUtils in the comments
-Phaser.Stage: Replace @name => @method on some functions
-Phaser.Physics.P2.BodyDebug=> fix @method names


photonstorm:
Nice one :+1:

PR #1580, opened 2015-01-28
fix #1551 - detect mouseup on tilesprites

### LINKED ISSUE #1551: TileSprite mouse click detection issue
When detecting mouse clicks on a `TileSprite` like this:

    var ground = game.add.tileSprite(0, 0, 600, 600, "ground");
    ground.inputEnabled = true;
    ground.events.onInputUp.add(mouseUp, this);

mouse clicks are registered only on part of the screen, which is usally close to the resoluion of the image used as tile. For instance on a 256x256 tile-sprite, clicks are detected only in the area from (0, 0) to approximately (256, 256), even though both the world bounds and game resolution are larger.

integricho:
Although the git diff might look a bit misleading (because the two branches share some pieces of code), all that is done is actually just change the order of the branches, so that `displayObject` is first checked if it's an intance of a `Phaser.TileSprite` (if so execute it's branch which remained unmodified), and if it isn't a `Phaser.TileSprite`, check if it's an instance of `PIXI.Sprite` (and do the same, code unchanged within the branch).
I guess the issue happened because `Phaser.TileSprite` is a subclass of `PIXI.Sprite`, so it got catched in the if check.

### DIFF
--- a/src/input/Input.js
+++ b/src/input/Input.js
@@ -844,10 +844,10 @@ Phaser.Input.prototype = {
         {
             return (displayObject.hitArea.contains(this._localPoint.x, this._localPoint.y));
         }
-        else if (displayObject instanceof PIXI.Sprite)
+        else if (displayObject instanceof Phaser.TileSprite)
         {
-            var width = displayObject.texture.frame.width;
-            var height = displayObject.texture.frame.height;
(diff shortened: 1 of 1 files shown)


photonstorm:
Well spotted, thanks.

PR #1582, opened 2015-01-30
Fix doc comment on Phaser.Physics.Arcade#getObjectsAtLocation.

micahjohnston:
When the doc comment was copied from getObjectsUnderPointer and edited, the @param for pointer was not changed to @params for x and y, so I just fixed that.


vulvulune:
You should change the name of the property in the comment:
* param {number} pointer - The x coordinate to check. =>
* param {number} x - The x coordinate to check.

micahjohnston:
Oh no, I thought I had done that! Fixed.

photonstorm:
Thanks :)

PR #1583, opened 2015-01-30
Add original position in onDragStart

vulvulune:
When the onDragStart event is trigerred, the position of the sprite has
already changed if the lockCenter is activated. So it's impossible to
know the original position of the sprite.
The new parameter give the position in all the cases.

See http://www.html5gamedevs.com/topic/12025-get-original-position-when-dragging-with-lockcenter/


photonstorm:
I'm going to do this slightly differently, but nice idea all the same, so consider it merged locally.

PR #1585, opened 2015-01-31
Update TilemapLayer.js

Leekao:
getTile now returns a copy of the results and not a reference.


photonstorm:
I can't merge this because it's against master not dev, but have made the change locally, thanks.

PR #1588, opened 2015-01-31
Fix: Path to Phaser in RequireJS project template.

ptlis:
Does exactly what it says on the tin :)

### DIFF
--- a/resources/Project Templates/RequireJS/src/main.js
+++ b/resources/Project Templates/RequireJS/src/main.js
@@ -6,7 +6,7 @@
         
         paths: {
             //  Edit the below path to point to where-ever you have placed the phaser.min.js file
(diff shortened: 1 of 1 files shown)


photonstorm:
Hmm, this line is meant to be modified to fit your project (not to actually point to the repo itself).

ptlis:
I might be missing something - all I wanted to do was run the example project to see how it hung together. I navigated into the RequireJS template project & ran bower install, but the example wouldn't run due to the path to Phaser being incorrect and this change fixed that template for me.

photonstorm:
I understand, but it's a project template and isn't meant to be run from that folder. You're supposed to copy it into your own project, tweak it and set things up from there. Either way I can't merge this as it's against the wrong branch, but will consider doing a dev branch local change.

PR #1589, opened 2015-01-31
fixed scrollTo action triggered by click on TOC

abderrahmane-tj:
since the navbar is fixed, its height is not taken into account when scrolling to a specific h4. So I am subtracting the height of the navbar plus the margin-top of an h4 to keep things nice :^)

### DIFF
--- a/docs/scripts/toc.js
+++ b/docs/scripts/toc.js
@@ -13,8 +13,10 @@ $.fn.toc = function(options) {
       e.preventDefault();
       var elScrollTo = $(e.target).attr('href');
       var $el = $(elScrollTo);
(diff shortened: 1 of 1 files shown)


photonstorm:
Thanks for this. Just 2 issues: 1) the PR needs to be against the dev branch and 2) This needs to edit the docstrap template itself, not the output in the docs folder. I believe the file you'd need to edit is resources/docstrap-master/template/static/scripts/toc.js

vulvulune:
This problem has been corrected in this commit of docstrap:
https://github.com/terryweiss/docstrap/commit/79c46e9cfad9c0337e870c7c831707fabe7fe3d3

photonstorm:
Nice, have updated to use the docstrap file.

PR #1594, opened 2015-02-03
Allow max velocity of 0 in Arcade physics computeVelocity

zekoff:
Previous method of setting default max did not allow falsey values.

Allows movement to be constrained to one axis. This is useful when using methods like accelerateToPointer for mouse control over e.g. the paddle in a Breakout-style game, player ship in Galaga-style game, etc.


photonstorm:
Lovely, thanks.

PR #1595, opened 2015-02-03
fixed scrollTo action triggered by click on TOC

abderrahmane-tj:
Since the navbar is fixed, its height is not taken into account when scrolling to a specific h4. So I am subtracting the height of the navbar plus the margin-top of an h4 to keep things nice :^)


photonstorm:
Thanks for this - I've merged in the latest toc.js from the docstrap repo which fixes this, but you're still credited in the readme :)

PR #1601, opened 2015-02-06
use XDomainRequest in Loader.xhrLoad

draconisNoctis:
XDomainRequests wasn't used for atlas json loading. So i moved xdr into the xhrLoad method to use it for all request if required.


photonstorm:
:+1:

Issue #1603, opened 2015-02-06
Tilemap layer camera and collisions broken when scale is changed


JRogaishio:
When changing the tilemap scale, the camera will be stuck to the top-left and the collision detection will not be updated to the new sized tiles.   
Using the below code will resize the layer but cause these issues.
var layer = map.createlayer('myLayer');
layer.scale.set(2,2); //Double the scale
  
Perhaps a "Scale" parameter could be added to the tilemap member of the loader or even the tilemap.createLayer. I haven't looked at the phaser source but in theory if you scale the tiles before it hits the collision detection then it may work and be easier to implement. Something like:

mickez:
I actually implemented proper scaling to tilemap layers today, that makes both rendering and collision work, without having to pre-process any files/textures.

I'll be more than happy to submit a PR of the code if that feature would be appreciated

JRogaishio:
>I'll be more than happy to submit a PR of the code if that feature would be appreciated
Thank you! I just snagged the .min file from your fiddle and it works great in my implementation!  
This issue was a major show-stopper for me because my tile map is 16x16. Without proper scaling it looked terrible. Hopefully this gets merged in to develop soon!

mickez:
We had the same problem (up until today that is!) and we used the naive approach to pre-process all files to a size of 64x64 (which was reaaaaaaally bad for performance). I'm glad I could help! :)

photonstorm:
It's now in the dev branch and will be part of the 2.3 release later this week (or early next week depending how things go)

darde:
Thank you @rehnarama , your PR helped me as well.

Issue #735, opened 2014-04-18
After scaling the layer of a tilemap camera seems to get buggy


eugenioclrc:
I was working on a small map that i scale to make the player feels that its a big (and pixelated) world while its just a small and simple tilemap.

While doing this i've noticed that the camera its not working quite well. I have prepared a demo using one of the examples from http://examples.phaser.io

Demo using mario
http://jsbin.com/xicav/3/

Source;
http://jsbin.com/xicav/3/edit?js,output

JRogaishio:
Is there any update to this? I'm trying to scale my tilemaps as well but everything is coming out wonky. Doing the below will re-scale the layers but the collisions are not fixed even if scaling is before the map.setCollisionsBetween.
```
var layer = map.createlayer('myLayer')
layer.scale.set(2,2); //Double the scale
```

photonstorm:
This feature is now merged in the dev branch and will be part of the 2.3 release.

PR #1605, opened 2015-02-07
Added scaling capability to TilemapLayer

mickez:
See feature request  #1603 and #735

Usage is as following:
```
var layer = map.createLayer('layerName');
layer.setScale(4);
```

A common use case would be to make sure your pixel art game renders crisp at high resolution, without having to pre-process all files (which would cause an undesirable performance hit anyway).

JSFiddle example! http://jsfiddle.net/mickez/zp3f33hv/

it works by iterating through the underlying block data in the corresponding `map.layers` and set the `scale` property inherited from `Phaser.Image`. The rest of the code changes is to make sure the camera is moved accordingly and the correct size is rendered.

I'm unsure if all functions that would be affected by the maps size has been updated accordingly, but my tests have shown that rendering and collision works properly.

### DIFF
--- a/src/tilemap/TilemapLayer.js
+++ b/src/tilemap/TilemapLayer.js
@@ -330,8 +330,8 @@ Phaser.TilemapLayer.prototype.postUpdate = function () {
 
     //  Stops you being able to auto-scroll the camera if it's not following a sprite
     var camera = this.game.camera;
-    this.scrollX = camera.x * this.scrollFactorX;
-    this.scrollY = camera.y * this.scrollFactorY;
+    this.scrollX = camera.x * this.scrollFactorX / this.scale.x;
+    this.scrollY = camera.y * this.scrollFactorY / this.scale.y;
 
     this.render();
 
@@ -358,7 +358,7 @@ Phaser.TilemapLayer.prototype.postUpdate = function () {
 */
 Phaser.TilemapLayer.prototype.resizeWorld = function () {
 
-    this.game.world.setBounds(0, 0, this.layer.widthInPixels, this.layer.heightInPixels);
+    this.game.world.setBounds(0, 0, this.layer.widthInPixels * this.scale.x, this.layer.heightInPixels * this.scale.y);
 
 };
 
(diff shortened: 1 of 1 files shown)


photonstorm:
Nice. I'll have a play with this later. I'd like to see a way to just over-ride layer.scale instead of adding another method to it (as I bet it would really mess things up if you set them both by mistake :))

mickez:
That would be ideal and more consistent!

It wouldn't really mess things up per se, it's just that if you set the scale without the added method the data used for collision doesn't get updated.

There is however one problem, if you create more than one TilemapLayer of a single actual layer, and scale one, but not the other, the collision of one of the TilemapLayer's will be messed up! The only real solution to this would be to make a copy of the data instead of using the current approach using a reference.

I did play around without adding a new method. The data used for collision (in arcade physics at least) is the left/top/right/bottom-properties in the Tile-class. If every instance of Tile has a reference to the TilemapLayer it belongs to it would be trivial to just do the calculation inside the get function. I opted against it in this PR as the Tile data is populated with the creation of the TileMap and not with the creation of the TilemapLayer and I didn't really want to re-structure too much of the code :).

mickez:
Another approach to solve both problems described above would be to calculate the scaling of the TilemapLayer inside the arcade physics collision checks itself. This would remove the need to edit any data (that is to just simply use the scale property) and in turn still have the ability to add two TilemapLayers which uses the same data

photonstorm:
For some reason your jsfiddle doesn't work in WebGL mode, but maybe that's just the build you ran off? I get nothing rendered, just a grey screen.

mickez:
Must be the build, if not, surely the character and text should have rendered! I'll test webgl later today with my commits rolled back
________________________________
Från: Richard Davey<mailto:notifications@github.com>
Skickat: ‎2015-‎02-‎08 22:28
Till: photonstorm/phaser<mailto:phaser@noreply.github.com>
Kopia: Michael<mailto:rehn.michael@outlook.com>
Ämne: Re: [phaser] Added scaling capability to TilemapLayer (#1605)

For some reason your jsfiddle doesn't work in WebGL mode, but maybe that's just the build you ran off? I get nothing rendered, just a grey screen.

---

mickez:
It was just cross-domain complaints! I changed the images to base64 encoded images so I could have them inline. The jsfiddle is now updated and works with WebGL: http://jsfiddle.net/mickez/zp3f33hv/1/

photonstorm:
Perfect, I didn't have time to investigate fully but thanks for confirming. Ok let's get this merged and see what happens!

PR #1608, opened 2015-02-08
Added support for tiles rotated and/or flipped in Tiled

nkholski:
Added support for rotated and/or flipped tiles in Tiled tilemaps. 

Demo based on an example from examples.phaser.io:
http://dev.niklasberg.se/phaserRotateTiles


photonstorm:
This is a nice addition, thanks. I'll have to make it super-clear in the docs though that rotated tiles are absolutely not supported when it comes to collision though!

PR #1609, opened 2015-02-10
Read ObjectLayer Objects .type property for Tiles

mikaturunen:
Why isn't the .type already read from the objects details for Tile objects as it's there? Latest release version of Tiled 0.11.0 has this field for ObjectLayers Tile Objects and it's quite extensively used by other frameworks.

Well, I need this and I suppose there are other people too who might be interested in having this in here too. Currently I go around this by creating a custom property of "type" and setting my value on it but that's just extremely silly. :)

### DIFF
--- a/src/tilemap/TilemapParser.js
+++ b/src/tilemap/TilemapParser.js
@@ -448,6 +448,7 @@ Phaser.TilemapParser = {
 
                         gid: json.layers[i].objects[v].gid,
                         name: json.layers[i].objects[v].name,
+                        type: json.layers[i].objects[v].type,
                         x: json.layers[i].objects[v].x,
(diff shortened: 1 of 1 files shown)


photonstorm:
Because it doesn't exist prior to 0.11 - I don't mind merging this PR but please change it to only set the type property *if it exists*, or provide a fallback value if it doesn't.

mikaturunen:
Ah, due to historic reasons. That explains it. Thanks mate! Will do.

mikaturunen:
That should do it. Defaults to empty string in case of the .type property is missing from the Object.

photonstorm:
Perfect, thanks.

PR #1610, opened 2015-02-10
On Ceiling Method Addition

yigitozdemir:
Hi Guys,

I found out, there is a shortcut for controlling on floor but there is no shortcut for onCeiling control, i tought it might be useful to implement a same way for controlling if body collides ceiling or not.

### DIFF
--- a/src/physics/arcade/Body.js
+++ b/src/physics/arcade/Body.js
@@ -661,6 +661,16 @@ Phaser.Physics.Arcade.Body.prototype = {
     onFloor: function () {
         return this.blocked.down;
     },
+    
+    /**
+    * Returns true if the top of this Body is in contact with either the world bounds or a tile.
+    *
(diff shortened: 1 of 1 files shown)


yigitozdemir:
Today i asked about how to use grunt and tested my code with

```javascript
if(this.activeSprite.body.onFloor() | this.activeSprite.body.onCeiling() ){
        this.changeSprite("dead");
}
```

And its working. I tought if there is a function to control is body onFloor, there should be a function to control is body onCeiling.

pnstickne:
Make sure the jsdoc is correct, including the method name.

Curious question: why not just using `.blocked.up` directly?

Garbanas:
@pnstickne `.blocked.down|up` only works for TileMaps but not for other Arcade bodies - you'd have to check with `touching` or `wasTouching`, although these values change with every single collision if I remember correctly.

photonstorm:
This may be old, but why not.

PR #1611, opened 2015-02-11
Added support for animated tiles, as exported from Tiled map editor

nkholski:
I added support for animated tiles, as exported from Tiled map editor (including individual duration property per frame). I think this feature was added quite recently.

- It will animate all tiles with the same gid in sync (all animated tiles will be on the same frame on any given moment).
- It will, to my understanding, not affect the performance for tilemaps without animated tiles. High number of different animated tiles with unsynced timing for frames might do. One solution could be to check if a layer have animated tiles within the camera view before requesting a redraw. A good idea to avoid unnecessary redraws is to use only even intervals of 100 ms.
- It works well also with flipped and rotated tiles in my tests: http://dev.niklasberg.se/phaserRotateTiles/ (this version miss an important bug fix that probably hit perfomance)

If the pull request is not accepted and merged I will move the code into a plugin with the aim to add the feature as seemless possible. I think it would make a good addition to Phaser though.


pnstickne:
Something like this (or at least the work for it) is quite closely tied in with the entire implementation; so I think core is appropriate.

Anyway, notes:

- `time.now` is not guaranteed to be "any specific unit" so it should likely not be used for such; to time off of milliseconds *world time* (without pauses counted) use `time.time`; to time off of milliseconds *physics elapsed* use `time.physicsElapsed` (how many "ms of time to process" this time `update` was called).

- Setting the `dirty` of the layer to true invalidates the entire canvas and prevents it from being able to apply selective rendering updates. This is a larger issue that could be tackled.

- If I see `this.map.animatedTiles[gid]` another time going to get RSI just from looking at it :>

nkholski:
Thanks for the reply!

Comments on your notes:
* Should be an easy fix.
* I couldn't find any other solution. Is there any built-in way to initiate an selective render? The best would be to be able to re-render tiles within camera view that themselves have a property "dirty" set to true. I could of course add this function myself.
* :-) It should probably be an object too, not an array.
* My though was that sprites are more demanding than tiles, and that it wouldn't be such a performance boost. Shouldn't be to hard to add this. Don't know how to decide how many a few are though.

However, before I do any more work on this I would like to hear from Rich if there is a chance that this could make it into core.

pnstickne:
My vote is in - core, yes.

nkholski:
It's been a month since I initiated the pull request without any word from Rich, so I guess it wont be merged even if I improve it.

When I find time I will try to add the feature to my plugin https://github.com/nkholski/phaser-tiled-extras instead.

photonstorm:
It may have been a few months, but there has been no Phaser release since then (that is what you should gauge if I'll merge or not on)

I'd be willing to merge this into the core class, yes. It will need updating as per the points @pnstickne made above though (the time one being the most important). As it's a completely new feature it also doesn't matter terribly if it doesn't make the 2.3 release, and as things have changed a lot in there it may be worth doing for the one after.

nkholski:
Sorry if I sounded confrontative, didn't mean that. Happy to hear that it could make it into the core! I'll do my best to improve it. 

The second point will be the most challenging. Is there any built-in way to re-render just single tiles? One idea is to make a method that re-renders all tiles with a certain gid (or gid within a given array of gids) within the camera view or specified area. Another is to make the method check tile properties and re-render tiles with "dirty" set to true. Both approaches could be supported within one method: methodName(area (Phaser.Rectangle / NULL = camera view), gid (string / array / NULL = only by property)). The method could also be used by putTile too instead of "this.layers[layer].dirty = true;", and perhaps elsewhere.

photonstorm:
Don't worry about point 2 for now - just make it painfully clear in the jsdocs you add that the use of animated tiles causes this side effect and there's no current work-around.

nkholski:
So, I did a couple of changes (and struggled a lot with Github, and learned to use "npm test" among other things):

* It now uses time.physicsElapsedMS instead of time.now.
* Instead of selective rendering I added a test if animated tiles is within camera view before setting layer.dirty to true.
* map.animatedTiles is now an object instead of array.
* I realized that tilelayer.update is called on all layers, and only one call per main loop is desired. To solve this I added boolean map.animatedTiles.updated that will be set to true by first layer update call per game loop. (tilemap.update is never called, which would be nicer otherwise)
(I havn't tried to add sprites instead of animated tiles.)

Everything works well in my tests.

photonstorm:
When writing the jsdoc blockdoc for the new properties (like animatedTiles) you can reference any limitations in there and / or add to the TilemapLayer class docblock. Don't edit the html files, just edit the jsdoc comments in the source.

nkholski:
I added a note about performance. Might be a bit too long, and since English is not my native language (I actually failed in English) it might need some improvement.

So, given that I will not try to add support of replacing animated tiles with sprites or selective rendering of tiles I guess I'm done for now.

(I also changed a "=" to a "+=" which will help keeping animations in sync, and corrected a copy-paste misstake from Image to refer to TilemapLayer.)

photonstorm:
I'm sorry it's been so long since this was looked at, but I'm going to put this poor PR out of its misery and close it off. It's been left for too long, and merging now would be a nightmare, especially with the new WebGL renderer in place. We're going to have a look at a different way to support animated tiles.

PR #1612, opened 2015-02-11
Dev

spayton:
Fixed Phaser.Ellipse.contains() by reverting to previous code


spayton:
Eh, this doesn't look right, it's only 1 files, I'll try again.

PR #1616, opened 2015-02-12
For better or worse, drawPolygon and Polygon actually take any number of arguments of any type.

Bilge:


photonstorm:
Thanks for this, but am closing this PR off as these changes are already in (via the union types and other recent PRs).

Bilge:
@photonstorm You are correct for `phaser.d.ts`. However, `pixi.d.ts` is still wrong.

clark-stevenson:
I will be doing a review asap on Pixi soon too thanks.  I am not sure I 100% understand the relationship any more though. 

@photonstorm  is phaser using its own branch of pixi now?  I am concerned that if I continue to maintain pixi.d.ts based on goodboy digital commit history, that I get out of sync with phasers version of it.  If mat released 2.9.9 tomorrow, would you plan on using it or have we passed that point?


@Bilge  Can I ask your advice on an unrelated matter?

Would you say that these commits are Interfaces in a definition file? That for example `class Sprite extends Pixi.Sprite implements Component.Angle`  

https://github.com/photonstorm/phaser/commit/bf4c1d062080a774bea3c10b67cac63cca5bf5f4

photonstorm:
@clark-stevenson I think there are 2 things here, so in order:

1) I'm not using Pixi v3 at all, so if you are updating the defs to track that branch then they will be incorrect compared to Phaser. I am still using Pixi v2, I've just removed various things that aren't needed by Phaser and have started applying my own fixes locally. From what I can tell Pixi v2 is pretty much dead now. They claim it's in maintenance lock down, but it isn't really, all the bug fixing is taking place in v3 only, other than a couple of small things. So it's no longer beneficial for me to track it as we're ahead of it in terms of fixes.

2) I personally don't believe you need to worry about the Components in 2.3 re: the TS defs - they are mixins, applied as a one off hit when the game starts, they're not run-time components (i.e. the dev shouldn't be adding them on the fly to objects as they're created in-game). Therefore all the properties that Sprite currently has in the defs are correct, as that is how they're exposed to the devs.

I guess if you can find a 'clean' way to manage them as components re: interfaces then it would make sense to, but it's not essential - the way the defs are structured now are still correct.

clark-stevenson:
@photonstorm  thanks as always for your advice. 

1) Great! We are on the same page then. I am not making any pixi3 at this time. The last time I checked pixi was Jan 1st.  I wondered if they were going to break it down into modules etc or as you said, have basically shelved it. I will track only phaser (which means just a review of unions).

2) Awesome!  

Clark.

PR #1622, opened 2015-02-14
Fix for webkitAudioContext warning.

cainrus:
Fixed console warning with message: 'webkitAudioContext' is deprecated. Please use 'AudioContext' instead.

### DIFF
--- a/src/system/Device.js
+++ b/src/system/Device.js
@@ -897,7 +897,7 @@ Phaser.Device._initialize = function () {
     function _checkAudio () {
 
         device.audioData = !!(window['Audio']);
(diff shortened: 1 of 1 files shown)


photonstorm:
With your 'fix' in place I now get warnings in Chrome, Canary and Firefox. With the current Phaser code they all work fine with no warnings.

cainrus:
I'm sorry. It was really bad idea(did not sleep all night).
I think that I still receiving that annoying warnings because of mac + Canary.

photonstorm:
I'm using Canary on OS X too without seeing the warnings. I wonder if something else is triggering them.

Issue #1591, opened 2015-01-31
Make the Loader accept Headers parameters


Mefteg:
Hello,
I want to know if you consider the Loader to accept extra parameters for getting assets.
I need to specify a particular Header to get my assets from GitHub and I can't load them directly with Phaser for the moment.

I've already posted on the forum (http://www.html5gamedevs.com/topic/12093-phaser-loader-contribution-to-add-special-headers/) but I have no reply.

I know how I can modify the Phaser Loader to make it works but I want to know if you want this feature or not.
I see the 2.3 version will update the Loader and if this feature interests you, it can be the right moment to add it.

Thank you for your time, I hope we can work together.

P.S. : sorry for my english :)

photonstorm:
Show me what would be needed to change in the Loader to support this, and we can consider it for sure.

photonstorm:
In order to keep the Github issues clear this feature request has been moved to the [Phaser 3 wishlist](http://www.html5gamedevs.com/topic/7949-the-phaser-3-wishlist-thread/).

PR #1639, opened 2015-02-24
Allow custom creation of response object

pnstickne:
Per https://github.com/photonstorm/phaser/issues/1591

A fairly-generic method of intercepting request creation and specifying custom request objects (and configurations therein). Works for XHR / Image / Audio as well as "fine tuning" the URL or whatnot.

Usage for custom XHR headers looks like:

    loader.onCreatingRequest.add(createRequest, this);

    function createRequest (loader, r)
    {
        if (r.requestType === 'xhr' && shouldCustomizeRequestForUrl(r.url))
        {
            var xhr = loader.createXMLHttpRequest(r.url);
            // any relevant modification to XHR / request can be done here
            // (including providing a fake proxy object or using a different URL entirely)
            // but there are some restrictions and the xhr.responseType will be overridden
            xhr.setRequestHeader('custom-header', 'value');
            r.requestObject = xhr;
        }
    }


pnstickne:
"npm install" failed for Travis CI :<

photonstorm:
How safe is this to merge this close to release? At least I've extensively tested the current 2.3 Loader through several large client projects, but I can see lots of underlying structure changes here.

pnstickne:
I would say 'fairly safe'. Most of the structure changes come from unifying XDR+XHR code (of which I have sadly not tested for XDR, but XHR works as expected).

The building (and extraction of) the request configuration doesn't really affect the structure and semantic differences should only be observed _if_ adding an `onCreatingRequest` handler and altering the supplied values.

Some of the image/audio callback code is updated for consistency and I have not run into any regression cases.

photonstorm:
I'm afraid I'm going to leave this one be. I'm happy with how the Loader currently works, and will leave features like this for Lazer.

Issue #1641, opened 2015-02-26
Tilesets with margin/spacing render incorrectly


zekoff:
A rendering issue occurs when using a tileset with the following properties:

* 432x396 image size
* 32x32 tile size
* 12x11 distinct tiles
* 2px margin on all sides
* 4px spacing between tiles

The important properties for purposes of this issue are margin and spacing. The tileset image is included below, annotated with red for margin pixels and green for spacing pixels.

![bugreport](https://cloud.githubusercontent.com/assets/1486137/6386236/b72fa63e-bd3b-11e4-942a-2be42d044b61.png)

When creating a tilemap based on this tileset, `Phaser.Tileset` warns that "image tile area is not an even multiple of tile size" in the function `updateTileData`. Then, when the tilemap is loaded, it displays the wrong tiles.

Where Tiled renders map as this:

![tiled_rendering](https://cloud.githubusercontent.com/assets/1486137/6386305/5851f0da-bd3c-11e4-925f-3736183840f5.png)

Phaser renders map as this:

![phaser_rendering](https://cloud.githubusercontent.com/assets/1486137/6386306/6109d2ec-bd3c-11e4-8761-8fa778d7b4db.PNG)

**Minimal example to reproduce bug:**
https://github.com/zekoff/phaser-tileset-issue

photonstorm:
Closing this off as you solved it yourself and it's now merged!

Issue #1371, opened 2014-11-26
Tileset with gaps produces incorrect .json


Christian-Broomfield:
Original issue filed in Tiled:  https://github.com/bjorn/tiled/issues/831

Project used to reproduce problem described below: https://github.com/SoulBeaver/broken-tiled-tilemap2

Tiled screen:

![Tiled screen of tilemap](http://i.imgur.com/iMhRXrm.png)

Rendering using phaser-tiled plugin:

![Rendering tilemap using phaser-tiled plugin](http://i.imgur.com/WRoHn9l.png)

Rendering using vanilla phaser:

![Rendering tilemap using vanilla phaser](http://i.imgur.com/neWBnMD.png)

[The .json for this tilemap](http://pastebin.com/v7sfKZDK)

It turns out that Tiled has the correct ids in the tile layer, but Phaser is displaying an incorrect sprite.

**Example:**

The first tile in the tilemap is a corner bush with id 857; the tiled .json has 858, which is okay because the firstgid is 1. Phaser instead renders the broken bush, which is 842. We're off by 15. This makes my previous assessment of phaser missing one index per row. The bush in the tilemap is in the 15th row, so we've omitted 15 IDs.

[I also have a second project in TypeScript using a smaller Tilemap](https://github.com/SoulBeaver/broken-tiled-tilemap).

PR #1642, opened 2015-02-26
Fix row/col count formula.

zekoff:
**This fixes the rendering issue for tilesets with margin/spacing as described in photonstorm/phaser#1641.**

The formula at 217/218 as previously written only accounts for the left/top margin pixels, but adding the right/bottom margin is important for checking whether the image is sized correctly or not. Furthermore, when counting rows and columns, the use of Math.floor at 227/228 does prevent loading of partial tiles if the image is in fact sized correctly, but in this case (where the margin is smaller than the spacing) it will actually cut off the last row and column of tiles.

With the previous formula, rowCount is:
`var rowCount = (432 - 2) / (32 + 4) = ~11.95`
This causes the "even multiple" warning. The `Math.floor` call at 227 then forces this value down to 11 rows, when in fact there are 12. This in turn causes Phaser to assign tile IDs incorrectly, which leads to the wrong tiles being displayed.

Changing the formula at 217/218 to the following will handle this case, while preserving the current version's ability to discard extraneous pixels:
`var rowCount = (imageHeight - this.tileMargin * 2 + this.tileSpacing) / (this.tileHeight + this.tileSpacing);`
Thus:
`var rowCount = (432 - 4 + 4) / (32 + 4) = 12`
...which is the correct value, as confirmed by Tiled.

Since colCount works the same way, all of the above applies there.

See also photonstorm/phaser#1371, which resulted in the current version of the code. The change in the associated pull request fixed the issue, but didn't account for these margin/spacing issues.

Additional "test cases":
* 4 rows, 16px tiles, 0 margin, 0 spacing. Image height would be 64px. `(64 - 0 + 0) / (16 + 0) = 4 rows`
* 4 rows, 16px tiles, 2 margin, 1 spacing. Image height would be 71px. `(71 - 4 + 1) / (16 + 1) = 4 rows`

Image showing result of this change to Phaser:
fixed_phaser_rendering

**Link to minimal project with a Phaser build that includes this fix:**
https://github.com/zekoff/phaser-tileset-fixed

### DIFF
--- a/src/tilemap/Tileset.js
+++ b/src/tilemap/Tileset.js
@@ -214,8 +214,8 @@ Phaser.Tileset.prototype = {
     updateTileData: function (imageWidth, imageHeight) {
 
         // May be fractional values
-        var rowCount = (imageHeight - this.tileMargin) / (this.tileHeight + this.tileSpacing);
-        var colCount = (imageWidth - this.tileMargin) / (this.tileWidth + this.tileSpacing);
+        var rowCount = (imageHeight - this.tileMargin * 2 + this.tileSpacing) / (this.tileHeight + this.tileSpacing);
+        var colCount = (imageWidth - this.tileMargin * 2 + this.tileSpacing) / (this.tileWidth + this.tileSpacing);
 
         if (rowCount % 1 !== 0 || colCount % 1 !== 0)
         {


photonstorm:
Nice one, thanks :+1:

PR #1647, opened 2015-03-01
Moved component installation out of constructors

pnstickne:
Components / prototypes are now installed once-per-type instead of every time a new object is constructed.

This also applies some minor fixes such that the above works cleanly.


photonstorm:
Nice, this was on my list so at least it'll free-up some more time. I also intend to move the component arrays out of the classes and into a single file somewhere so they can be easily overridden before Phaser starts up.

pnstickne:
I hadn't thought of that scenario.

Best approach I can think of if such is desired so to then have something like:

   `components = PhaserGlobal.DefaultComponents.Sprite || <default_components>`

It would be required to be run a script before Phaser.js.

photonstorm:
I was going to make it part of the game configuration object to be honest. You couldn't create a game object until the game has booted anyway, so it should be safe to do at that stage.

photonstorm:
Hmm something about this change has stopped objects from correctly inheriting properties.

In short - getters, settings and functions that exist on the mixin prototypes are applied to the instances just fine, but properties are ignored.

For example if you log out `Phaser.Sprite.prototype` you'll see all the correct properties like _exists, alive, fixedToCamera, etc. All correct. Now create an instance of Phaser.Sprite and view it, all of those things are missing. The only things that exist on the instance are functions.

Maybe it's the mixin method itself not copying them over.

(just thinking out loud here)

adelciotto:
This is weird, I've done some investigation and it does not seem to be a problem with the mixinPrototype or install function, its something else.

If i log some of the missing properties within the Component.Core.init function upon a GameObjects construction, it seems to log them fine, but if I simply log the **this** context, they are missing.

```javascript
Phaser.Component.Core.init = function (game, x, y, key, frame) {

 .....
    console.log('_exists: ' + this._exists);
    console.log('components: ' + this.components);
    console.log(this);

};
```

screen shot 2015-03-07 at 4 04 59 pm

pnstickne:
@photonstorm  I just pulled from dev and this does not appear to the be case anymore.

The prototypes appear correctly for me when doing a `new Phaser.Sprite()`. Is this still a concern?

photonstorm:
Nope, I spent a good while working through it and tracking it down, should be all good now.

Aiming to release 2.3 this week, just going to tidy up the docs and add padding to the Text object.

Issue #1666, opened 2015-03-15
Tile.alpha is not working


nkholski:
It seems that any tile with alpha != 1 will cause a bug for the whole layer, making it flicker or go blank.

Snippet to try in the console (global var map):
for(var x=0;x<map.width;x++){for(var y=0;y<map.height;y++){var tile = map.getTile(x,y); if(tile){tile.alpha = 0.5;}}}

Example:
http://dev.niklasberg.se/phaser-tiled-extras/example/
Run in the console for flickering layer:
for(var x=0;x<map.width;x++){for(var y=0;y<map.height;y++){var tile = map.getTile(x,y,layers[1]); if(tile){tile.alpha = 0.5;}}}

pnstickne:
Which browser(s) does the problem show up on?

It looks fine in Chrome.

nkholski:
Chrome 41.0.2272.89 (64-bit/Ubuntu)
Is it fine even when the tilemap is scrolling? That's when the problems start for me.

pnstickne:
Okay, I get the flickering now. The code should be ..map.getTile(x,y,**layers[0]**).. to set the tiles.

pnstickne:
@nkholski I've pushed a dev patch. As a work-about this should do the trick:

```
Phaser.TilemapLayer.prototype.renderOrig = Phaser.TilemapLayer.prototype.render;
Phaser.TilemapLayer.prototype.render = function () {
    this.context.save();
    this.renderOrig();
    this.context.restore();
}

pnstickne:
Also, I got Mario to fall off the map :crying_cat_face:

nkholski:
Great work! I know the falling of map-problem. It's happens sometimes with one-way-collision-tiles. I had a hard time reproducing the exact problem, but most of the times it will occur in the "tower" full with arrows and never on a tile with collisionUp set to true. When I know more I will post an issue or PR.

photonstorm:
Closing as this is now fixed in dev :)

PR #1671, opened 2015-03-17
Tilemap rendering sets initial state better

pnstickne:
- #1666 - previous alpha state could bleed through.
  Now the initial alpha and composite modes are set explicitly.


pnstickne:
Does not fix problems fully

pnstickne:
But that does fix it..

PR #1674, opened 2015-03-18
Fix debug drawing for p2 physics bodies

englercj:
There were a couple of problems preventing the debug drawing from being correct:

1. The shape check in `draw()` needed to check for Convex last, since other shapes (like Rectangle) inherit from Convex.

2. The `updateSpriteTransform()` function needed to be called from the ctor. Otherwise bodies with no sprite (so no postUpdate call) would never be moved to draw in the correct position.


photonstorm:
:+1:

PR #1675, opened 2015-03-18
Split p2 definitions into its own file

englercj:
The p2 ts definitions were embedded into the phaser typescript file. This splits it into a separate file that the main phaser file references. This makes it consistent with the way PIXI's definition file is treated.


clark-stevenson:
Good catch, it was pasted in there originally from my p2.d.ts repo which may be improved since it was included in phaser.  I will check it out if Rich accepts your PR.

photonstorm:
If you're happy @clark-stevenson then I'm happy.

mikaturunen:
:+1:

PR #1683, opened 2015-03-21
Update input.js hitTest method for TileSprite

nextht:
Phaser.TileSprite is SubClass of PIXI.Sprite, so in hitTest we should check Phaser.TileSprite before PIXI.Sprite


photonstorm:
This is already fixed in the dev branch (where all PRs should go btw)

Issue #1679, opened 2015-03-19
Destroying a group with a sprite containing an animation on the onComplete signal causes a JS error.


trueicecold:
Example code:

```
this.enemy_sprite = this.group.create(x, y, 'zombie');
this.enemy_sprite.scale.set(1);
this.enemy_sprite.scale.x = -1;
this.enemy_sprite.smoothed = false;
this.enemy_sprite.anchor.set(0.5, 0.5);

this.walk_anim = this.enemy_sprite.animations.add('walk');
this.walk_anim.onComplete.add(this.destroy, this);
this.walk_anim.play(10, false);

this.destroy = function() {
    this.group.destroy(true);
}
```

causes a JS error:

```
Uncaught TypeError: Cannot read property 'getFrame' of null phaser.js:53464 
Phaser.Animation.update phaser.js:53464 
Phaser.AnimationManager.update phaser.js:52802 
Phaser.Sprite.preUpdate phaser.js:37442 
Phaser.Group.preUpdate phaser.js:21860 
Phaser.Group.preUpdate phaser.js:21860 
Phaser.Stage.preUpdate phaser.js:20367 
Phaser.Game.updateLogic phaser.js:26737 
Phaser.Game.update phaser.js:26684 
Phaser.RequestAnimationFrame.updateRAF phaser.js:46614 
Phaser.RequestAnimationFrame.start._onLoop phaser.js:46598 
```

Note: delaying the "group destroy" by using setTimeout seems to fix it, but it's not a proper solution...
My guess is that the animation keeps running for a couple more frames even when the onComplete signal was sent?

Using Phaser 2.2.2

pnstickne:
It doesn't keep running. The problem is group.destroy triggers a destroy of the animation from within the `update` - but later on, in the same update, the (now invalid) `_frameData.getFrame` is called.

pnstickne:
There is a patch for the issue.

In the meantime the best work-about is probably to keep using `setInterval(.., 0)`.

photonstorm:
Fix merged in dev.

trueicecold:
Downloaded Tarabon yesterday, the bug still occurs. But setTimeout seems to work just fine.

pnstickne:
@trueicecold Make sure to build locally (and not just a build pulled from git) when testing these 'latest and greatest' in dev.

[...]

PR #1685, opened 2015-03-23
Animation - guard around destroy

pnstickne:
- Allows sprites (and/or the animation) to be destroyed from within
  the onUpdate, onLoop, or onComplete animation events

Addresses https://github.com/photonstorm/phaser/issues/1679


photonstorm:
:+1:

PR #1687, opened 2015-03-23
Update Sprite.js for renderCanvas method

nextht:
Default when we change Sprite.width and height, we are changing Sprite.scale acutally. I need to change the size of Sprite but NOT changing its scale, so i modify Sprite.texture.frame size. I test changing Sprite.texture.frame size all right in WEBGL, but when i switch to CANVAS, i found the Sprite's size doesn't change. 
I found the Sprite._renderCanvas's code drawImage not using texture.frame'size but using texture.crop's size. I think we should use frame's size for drawImage.

### DIFF
--- a/src/pixi/display/Sprite.js
+++ b/src/pixi/display/Sprite.js
@@ -399,8 +399,8 @@ PIXI.Sprite.prototype._renderCanvas = function(renderSession)
                                 this.texture.crop.height,
                                 dx / resolution,
                                 dy / resolution,
-                                this.texture.crop.width / resolution,
-                                this.texture.crop.height / resolution);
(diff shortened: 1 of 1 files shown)


photonstorm:
This change breaks sprites created from trimmed texture atlases. It should always use the crop value, not the frame size.

photonstorm:
Also changing the width/height *should* change the scale. That's exactly how Pixi works internally (copied from how Flash does it). There will be no removal / changing of this in any version of Phaser 2, it's an API change beyond anything a point release is entitled to.

nextht:
Thanks @photonstorm for pointing out, i forgot trimmed case.

PR #1688, opened 2015-03-23
jsdoc update

cuixiping:


photonstorm:
PRs to dev branch only (and angularVelocity is already updated there btw).

PR #1695, opened 2015-03-24
Minor Updates RC1

clark-stevenson:
Some really minor updates as usual. 
Cleaned up the readme and got rid of the webgl.d.ts file as you need 1.4 so its useless.


photonstorm:
Thanks man.

PR #1697, opened 2015-03-25
Update Input.js

uhe1231:
fix the hitTest function:

line 837 : else if (displayObject instanceof PIXI.Sprite)
line 853 : else if (displayObject instanceof Phaser.TileSprite)

if displayObject is tilesprite  instance. "displayObject instanceof PIXI.Sprite" is true

so the "displayObject instanceof Phaser.TileSprite" is not run.

i think that is bug and i change code order fix the bug!


photonstorm:
This is already fixed in the dev branch. Please always check dev first!

PR #1710, opened 2015-04-01
Fixes neverending tween when duration set to zero (caused by division with zero)

hardalias:
Passing a value of zero as the value of duration causes the
 calculations performed in Tween#update to divide by zero when
 calculating the next step for the tween. This causes the tweened
 property value to be set to NaN having undesireable results and
 also, causes the tween to never end since the ending criteria are
 never met.


photonstorm:
:+1:

PR #1711, opened 2015-04-01
Fix some errors in the definitions files for Phaser.Group

isuda:
There was some errors in the definitions for Phaser.Group causing Typescript to error when using bringToTop and moveDown.

It appears to have been broken here: https://github.com/photonstorm/phaser/commit/7994427583edfad1454be42b58f9e4c837dc8e01


photonstorm:
:+1:

PR #1719, opened 2015-04-03
Typescript definition fixes for Graphics

TimvdEijnden:
Fixes incorrect parameters for generateTexture.

Both parameters should be optional because 
[resolution is set to 1 if undefined](https://github.com/photonstorm/phaser/blob/v2.3.0/src/pixi/primitives/Graphics.js#L641)
[scaleMode is set to PIXI.scaleModes.DEFAULT](https://github.com/photonstorm/phaser/blob/v2.3.0/src/pixi/textures/BaseTexture.js#L53)

I wanted to override generateTexture in Phaser.Graphics so they're also optional in Phaser documentation, but haven't found a good way to do this. I was thinking about this:

src/gameobjects/Graphics.js

```javascript
Phaser.Graphics.prototype.superGenerateTexture = Phaser.Graphics.prototype.generateTexture;

/**
 * Useful function that returns a texture of the graphics object that can then be used to create sprites
 * This can be quite useful if your geometry is complicated and needs to be reused multiple times.
 *
 * @method generateTexture
 * @param {Number} [resolution=1] - The resolution of the texture being generated
 * @param {Number} [scaleMode=PIXI.scaleModes.DEFAULT] - Should be one of the PIXI.scaleMode consts
 * @return {Texture} a texture of the graphics object
 */
Phaser.Graphics.prototype.generateTexture = function(resolution, scaleMode) {

    if (typeof resolution === 'undefined') { resolution = 1; }
    if (typeof scaleMode === 'undefined') { scaleMode = PIXI.scaleModes.DEFAULT; }
    this.superGenerateTexture(resolution, scaleMode);

};

### DIFF
--- a/typescript/phaser.d.ts
+++ b/typescript/phaser.d.ts
@@ -1405,6 +1405,7 @@ declare module Phaser {
         destroy(): void;
         drawTriangle(points: Phaser.Point[], cull?: boolean): void;
         drawTriangles(vertices: any[], indices?: number[], cull?: boolean): void;
+        generateTexture(resolution?: number, scaleMode?: number): PIXI.RenderTexture;
         postUpdate(): void;
         preUpdate(): void;
         update(): void;


photonstorm:
Shouldn't this be addressed in the pixi.d.ts file instead though? cc @clark-stevenson

clark-stevenson:
@photonstorm  the PR is correct for the d.ts file for sure, I think @TimvdEijnden is looking for advice if the documentation should be fixed also.

photonstorm:
I'm not sure it is in the right defs file though, as Phaser.Graphics doesn't have a generateTexture method (it just uses the one defined in Pixi)

TimvdEijnden:
That's true, what can I do to get it merged? It's in the right defs file if I add generateTexture to the Graphics, but I didn't found how to call the super.generateTexture.

clark-stevenson:
Does this look right @TimvdEijnden ? It was fairly out of date that function. 

I suppose my confusion is that I do not know if you want to add new functionality to the actual library itself, or if this is a definition only problem?

TimvdEijnden:
Yes that looks good, I only wanted to update the definition cause it caused compiler warnings. 
Good job :+1:

photonstorm:
Thanks for sorting this out guys, closing this one off :)

PR #1722, opened 2015-04-06
Minor change to Tilemap class

GGAlanSmithee:
Changes Tilemap.getTileWorldXY() to take an additional parameter - nonNull - to make it behave as Tilemap.getTile() does.


photonstorm:
:+1:

PR #1723, opened 2015-04-06
Fix for Phaser with requirejs

dgoemans:
Built a fix into p2.js for using requirejs with phaser. This has also been put as a pull request to p2.js. It was to do with a change in p2.js that stopped defining p2 as a global via browserify. Note that the build i put in this pull request was from p2's tag v0.6.1 and not master, so this should be good to go :)


dgoemans:
To see the changes, I've forked p2 here:
https://github.com/dgoemans/p2.js

photonstorm:
Awesome, thank you.

PR #1734, opened 2015-04-11
Looking up object layers based on id, uid or name

mandarinx:
The description of the method stated that you could pass a string to look up an object layer based on its name, but the original code only looked for a gui property, regardless of type.

I'm using Tiled 0.11.0 on OSX, and my exported json-file doesn't contain a gui property on the object layers. Here's a snippet of my file:
```
{
    "layers":[
        {
         "draworder":"topdown",
         "height":0,
         "name":"Objects",
         "objects":[
                {
                 "height":48,
                 "id":1,
                 "name":"pier",
                 "properties":
                    {

                    },
                 "rotation":0,
                 "type":"",
                 "visible":true,
                 "width":64,
                 "x":208,
                 "y":240
                }],
         "opacity":1,
         "properties":
            {
             "type":"pier"
            },
         "type":"objectgroup",
         "visible":true,
         "width":0,
         "x":0,
         "y":0
        }]
    }]
}
```

The property in my file is called id and not uid. This might be due to changes in one of the later releases of Tiled. I kept the check for uid to avoid possibly breaking compatibility with older versions of Tiled.

I also added scaling of the sprite.

Feel free to make formatting changes, or rewrite the function to better fit the programming style of Phaser.

### DIFF
--- a/src/tilemap/Tilemap.js
+++ b/src/tilemap/Tilemap.js
@@ -390,10 +390,32 @@ Phaser.Tilemap.prototype = {
         }
 
         var sprite;
+        var found = false;
 
         for (var i = 0, len = this.objects[name].length; i < len; i++)
         {
-            if (this.objects[name][i].gid === gid)
+            if (typeof this.objects[name][i].gid !== 'undefined' &&
+                typeof gid === 'number') {
+                if (this.objects[name][i].gid === gid) {
+                    found = true;
+                }
+            }
+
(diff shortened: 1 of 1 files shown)


mandarinx:
I realise I probably should have made this pull request to the dev branch. Should I create one on the dev branch?

photonstorm:
Please re-submit against dev branch, thanks.

PR #1745, opened 2015-04-15
Button now has useHandCursor set to true by default

alamboley:
I don't see any reason why it isn't set to true by default... ?


photonstorm:
Hmmm, I guess it could be. It's unnecessary on mobile though which is where Phaser is used most.

alamboley:
Well, does it have any inpact on mobile performance? It shouldn't, and if it's the case, Phaser should not set it to true on mobile.

I'm on mobile web & desktop + CocoonJS, each time I create a button I've to add that damn line. Obviously, I can extend Button...

photonstorm:
Ah what the hell, it's in 2.4.

alamboley:
:+1:

PR #1772, opened 2015-04-29
Stubbed Net and Debug so they can be properly excluded during a custom build.

soldoutactivist:
In Game.World#boot Phaser.Net is instantiated without regard to its inclusion. So I created a a stub for the Net module.

I also stubbed the Debug module.


photonstorm:
Awesome, thanks. Have been meaning to do this for a while.

PR #1777, opened 2015-05-02
Load audiosprite from asset pack

boniatillo-com:
The loader process the asset pack but to load the audio sprite, it uses the method "audio()" instead of "audiosprite()". The fix is to call "audiosprite()" with the right arguments.


photonstorm:
:+1:

PR #1781, opened 2015-05-06
Use `mute` instead of `_muted` to check whether or not a sound should be muted

joelrobichaud:
This fixes a problem with IE11 where the value of the mute property of the SoundManager does not affect sounds. This problem was brought up a couple months ago [in the forums](http://www.html5gamedevs.com/topic/10840-soundmanager-mute-on-ie11/). Looking at the code, this problem probably affects all of the browsers that don't support WebAudio.

### DIFF
--- a/src/sound/Sound.js
+++ b/src/sound/Sound.js
@@ -385,7 +385,7 @@ Phaser.Sound.prototype = {
     /**
     * Called automatically by the AudioContext when the sound stops playing.
     * Doesn't get called if the sound is set to loop or is a section of an Audio Sprite.
-    * 
+    *
     * @method Phaser.Sound#onEndedHandler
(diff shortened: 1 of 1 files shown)


pnstickne:
The forum post is using `mute` already.. anyway, perhaps it is a different problem of `_muted` not being set correctly on the SoundManager? As the `mute` property will check `_muted` and then the global game mute.

(It might be more appropriate to use `mute` anyway, but it would be nice if the underlying cause was more clearly identified.)

photonstorm:
Sorry but I need to see a test case of how this fixes anything before I can merge it (as I don't believe it will actually make any difference). I appreciate this PR is very old now though, so don't worry if this is long dead.

PR #1783, opened 2015-05-06
Update TypeScript State.init signature to allow multiple optional parameters.

jamesgroat:
The typescript signature of State.init should allow multiple optional parameters passed in from StateManager.start.


photonstorm:
Thanks for the PR, but please submit against the dev branch. Also you don't need to edit the comments defs file, as it's auto-generated based on the jsdocs and ts defs file :)

jamesgroat:
FYI -- Looks like this was addressed in phaser.d.ts in dev already: https://github.com/photonstorm/phaser/commit/a0a8f34c7b1fa53d1402e5848f51b62e636bf6fb

Issue #1782, opened 2015-05-06
Tween#generateData with no arguments causes infinite loop, memory issues.


rblopes:
Hi. I think you did a little oversight there.

In the Phaser' API manual, the `Tween#generateData` method is documented as having two optional arguments, whose default parameter values are, respectively, a default frame rate of 60 FPS and an empty Array to store the generated data.

Looking through the class source code, though, I noticed that only the last parameter is assigned the default value, possibly leaving the `frameRate` parameter undefined.

When that parameter is omitted and not assigned the appropriated value, its value gets coerced by JavaScript to `NaN`, due to the calculations done by the `TweenData#generateData`, which doesn't handle the value in any way either, an makes the program run in an infinite loop.

When I tried that method without a frame rate argument, misled by the documentation, my web browser became unresponsive, causing high CPU and memory usage. It took me a while to figure out why. So much for a missing line of code, I guess :)

The issue could be reproduced as in the following code snippet but, really, I don't recommend that.

```js
var t = this.make.tween({ foo: 0 });
t.to({ foo: 10 })
t.generateData(/* no arguments */); // Actually, don't try this :)
```

This issue is present in both release (`v2.3.0` and earlier builds) and development code bases.

----
2015-05-08: Edited.

photonstorm:
Merged to dev, thanks.

PR #1785, opened 2015-05-09
Assign the default value of 60 FPS to the optional `frameRate` parameter, when ommited.

rblopes:
Related to #1782.


photonstorm:
:+1:

PR #1794, opened 2015-05-13
Add a Heal Method to Phaser.Components.Health

stephandesouza:
Ideal when used in games with health packs, bonus items, or healing sprites.

In fact: Doing a negative damage causes the same effect, but as a new method is better for documentation.


photonstorm:
Yes makes sense :)

PR #1797, opened 2015-05-15
Use Pixi.js v3

englercj:
I know, I know. Pixi v3 has a while to go before it is bug-free enough for you to want to merge something like this, but I have been using this for a new project and just wanted to share how it could be done.

Basically this switches from using a copied in-place pixi.js to using pixi.js from npm. The build methods stay the same, it just uses the prebuilt pixi.js that is packaged in npm.

Feedback welcome.


photonstorm:
Aside from quite a few local changes we've made to Pixi 2 I'm curious how many things would this add back in that we don't ever use? (resource loader? event handling? filters? etc?) - I know if Phaser was properly modular it could chose not to use them, but that isn't happening in v2s lifetime.

englercj:
@photonstorm Totally, and it probably adds a bunch. Alternatively a build script can run `gulp build -e loader -e interaction -e filters` to build the pixi file that is then imported.

That keeps you able to easily use the pixi package, but minimize your footprint? Just a possible solution.

georgiee:
Oh thanks for this. I'm currently prototyping for an upcoming project where I need Phaser instead of a Pixi only environment. I always wondered how many places I would have to change in Phaser to use Pixi.js V3. To my surprise this looks more or less like the default Pixi v2 to v3 converting process. I'm looking forward to the moment where pixi is bug free enough to get merged.

photonstorm:
Closing off old PRs that are now fixed, no longer required, locally merged or were just for testing.

PR #1799, opened 2015-05-16
Fix Rope.js

gionatan7:
Rope creation key parameter didn't work, because PIXI.Rope requires a texture, not string.
Changed it like Sprite.


photonstorm:
Good catch, thanks (shows how often people use Rope!)

PR #1808, opened 2015-05-21
Update sample asset pack file.

rblopes:
This commit brings up some needed updates to the `assetPack.json` sample file,
including:

-   The addition of the missing sample entries for the following file types,
    added in recent versions of Phaser:

    -   a XML file;
    -   an Audiosprite file;
    -   a Video file entries.

-   Fix the example Bitmap Font entry.


photonstorm:
Excellent, thank you.

PR #1810, opened 2015-05-24
Sprite vs Tilemap can now check for overlap, without trying to separate

Preece:
Currently, the ability to check for an overlap between a Sprite and Tilemap does not exist. This is because the overlapOnly property from collideHandler stops getting passed down the chain (or is passed, and ignored). This update brings the overlapOnly property down to the "separateTile" function. Within that function, an intersection between the Sprite and Tilemap will return true and not proceed with the separation.


Preece:
Anything else I can do to get this merged? It would help greatly with some current issues I am facing.

PR #1811, opened 2015-05-24
Improvements to Text word wrap functionality.

soldoutactivist:
- Improved word wrapping functionality:
    - Now breaks long words if they are the first of a line.
        - And repeats the process as necessary.
- Possible breaking changes:
    - White space is condensed (e.g., consecutive spaces are replaced with one).
    - Lines are trimmed of white space before processing.
    - Throws an error if the user was smart enough to specify a wordWrapWidth less than a single character.


photonstorm:
This has now been merged in a slightly different way.

PR #1825, opened 2015-05-29
fixed the responsive example

rsadwick:
spriteMiddle not defined and uses game variable game instead of this.game was breaking the entire example


photonstorm:
Nice :+1:

PR #1826, opened 2015-05-29
Custom build system refactoring

rblopes:
Hi again. Just so you know, I have been messing with the build tasks for quite a while now. Don't worry, though, it's good stuff :smiley:

To start, that custom task code you've written earlier was restructured and split in two related tasks.

The first one is a `custom` task, for the users. Essentially, same stuff as before, just a few subtle differences:

  * It accepts a new `--usage` flag, displaying the CLI options and listing the optional modules available. (Note: the flag `--help` is reserved by Grunt and can not be overridden.)

  * The `--exclude` flag is now optional. So, even if the user don't ever want to remove any module from his custom build, now he can do that. In contrast, it was only allowed to run the `custom` task if the user offered a set of optional modules to be excluded or the task would fail, which seems quite counterintuitive.

The second one is a newly created, Grunt `build-phaser` multi-task, based on your previous work. This is the main task, used internally for Phaser releases. I found that this kind of specialized task can benefit from much of Grunt's features, much like how the 'contrib' plugins do.

So now, along with the custom task, the following build tasks are now available too:

  * `build-phaser:standard` builds the main `phaser.js` release file.
  * `build-phaser:arcadephysics` for the AP-only custom build.
  * `build-phaser:nophysics` with no physics engines included.
  * `build-phaser:minimum` including only the required Phaser modules.
  * `build-phaser:full`, which embeds all available Phaser modules.

Some notes here:

  * I did some interpretation of what a 'full' or a 'standard' build is, so I named the tasks such way to avoid confusion.

  * Due to its nature, the 'full' task is not ran by default, only the former four.

  * The first four tasks are aliased to simpler names, with the `build-phaser:` prefix omitted.

So, much of the work done here ensured that the distribution tasks worked the same as before. For you, that means you'll use Grunt to manage Phaser releases the same way you're used to, no big surprises.

Much of the improvements, though, were made in the (you guessed it) tasks internals, and I tried to tackle much of the limitations you may have struggled to solve. The main issues I've found dealt with how the various Phaser modules are managed, including dependency tracking, and how the users choices need to be processed.

The first issue I tried to alleviate was how modules are declared and managed. Before, you had to edit code in three different places to manage modules. In the `Gruntfile`, there was an object accounting for the module ordering, and information like a module description and whether it's required or optional. The `manifests` JSON files where just mere file listings. The `concat` sub-tasks were the worst part, making for most of the code bloat. There were also the stub files, should a module being excluded from a given build still being required somehow. And two minor cases: the support for Tilemaps AP being an 'opt-in'; and the entry for the 'Creatures' module missing in the `concat` configuration file.

To simplify that process, all the information required for a module was combined in a single place — inside each manifest file. So, I ended up devising the following format for the JSON files:

```js
{
    "description": "A module's description",
    "optional": /* a boolean value */,
    "dependencies": {
      "modules": [ /*an array of module names this module depend on*/ ],
      "reason": "Some explanation why the above modules are necessary"
    },
    "files": [ /*the source file listing*/ ],
    "stubs": /*an array listing stub files, or a 'null' value meaning none*/
}
```

An `index.js` file inside the `manifests` directory is responsible of tracking module ordering — away from the task file. Also, manifests now carry information about module dependencies.

That greatly simplifies how modules are selected and bundled together. Modules are no longer concatenated and its partial results placed in a temporary directory. Instead, Module files are read directly, avoiding the necessity for `concat` sub-tasks -- which is one file less to edit.

The main benefit here is that you have less things to worry about. If you introduce a new feature in Phaser, just edit a new JSON file, like above, declare it in the `manifests/index.js`, and it's done.

The next issue is about dependency tracking. When processing users choices, some problems may arise. What if s/he tries to remove a required module, like 'gameobjects'? I noticed that there was no way to avoid that, the required module could just go missing, resulting in a broken build. What if a new module depended on some other, optional module? More code was required to link these pieces and handle the dependency properly.

At the expense of writing some more code, some routines were developed to guarantee that the task could run properly, producing the desired results and aiding on automating this process. One to filter invalid user inputs, be it unknown modules or not optional modules. So, for example, if users misspell some module names, the filtering invalidates the input, and the reasons why are reported back to them. Then, the dependencies of each optional module are evaluated, and if one is missing, the module gets removed too — same behavior as the original task code.

As for the stability of the newly written tasks, I used them to build Phaser v2.3.0, with almost no differences -- only the version tag and the time stamp.

Lastly, I upgraded some of the Grunt's plugins to their latest releases.

To summarize, this branch adds the following files to the tasks directory:

```
tasks/
├── build-phaser.js           # The `build-phaser` multi-task.
├── custom.js                 # The `custom` task.
├── helpers/                  # Helper modules for:
│   ├── selectBuildFiles.js   # * Files selection.
│   └── validateExcludes.js   # * Validation routines and filters.
├── manifests/
│   └── index.js              # A CommonJS module index for the JSON files.
└── options/
    └── build-phaser.js       # The multi-task configuration.
```


photonstorm:
Wow, that's quite some work here! I'll merge it and start testing with 2.4 later today, but it sounds great from what I've read above, so thank you.

photonstorm:
Hmm I can't merge this from Github for some reason. I could do it locally of course, but if you want to fix the PR so it doesn't conflict any more let me know and I can merge directly.

rblopes:
Well, I've tried to reproduce the issue and merge the changes locally just to be sure what the conflict is — I keep an up to date copy of the Phaser repo, just in case. As I've suspected, the only divergent file is the PIXI manifest file (`tasks/manifests/pixi.json`), due to upstream changes (photonstorm/phaser@fb4709fb1a7cab930c7550dacf88fb482904966e). I did one last commit to track that change a while ago, I guess it's safe to replace the current file with the one in the PR. I'm afraid you'll have to do this in your local copy though.

One way to solve it locally is, after failing the merge, issuing `git co --theirs -- tasks/manifests/pixi.json` in a command prompt/terminal, then `add`ing the file again before commiting.

rblopes:
Ok, never mind that previous silly comment :)
I rebased the branch, so it's OK to merge again, without conflicts.
Please, before merging these into `dev`, do some tests with it to see if everything goes OK for you.

rblopes:
Oh man, it's broken again. :cry:
Rich, still interested in delivering this feature along with Phaser yet?
It may take a while for me to fix the conflicting changes this time, though. What you think?

photonstorm:
It's probably just broken because I had to add in 2 new modules to the build task, nothing else changed there (I literally just added in the rope and tilesprite modules). I'm still interested in merging this, but it was just too much to take on for the 2.4 release.

rblopes:
I'm closing this pull request for now. It seems that the requirements of the custom task have changed quite a bit since last time and are not compatible with what I have done earlier back in May. That'll require more time to review and doing a `git rebase` like earlier is impossible at the moment. I'm not quitting, though, some of these commits will be reviewed and pushed as smaller patches later instead.

PR #1827, opened 2015-05-30
Fix cache.getKeys() definition.

jamesgroat:
Fix typescript definition of Cache.getKeys to match the underlying js.


photonstorm:
:+1:

PR #1831, opened 2015-06-02
Minor TS Update

clark-stevenson:
You already ninja'd the 2 open tickets :+1:  so just adding snapPoint and bumping the review date.


photonstorm:
I probably ought to make snapPoint actually do something now :)

PR #1833, opened 2015-06-03
Fix docstring of drawCircle

danxexe:
The `drawCircle` method is missing from the docs, probably because the docstring is not properly formatted: http://phaser.io/docs/Phaser.Graphics.html/Phaser.Graphics.html


photonstorm:
:+1:

PR #1837, opened 2015-06-06
JSON support for BitmapFont atlases

Feenposhleen:
Since some environments (CocoonJS, I'm looking at you!) does not support parsing XML through DOMParser, this allows to bypass that by providing a pre-parsed JSON atlas instead. From what I've gathered there are quite a few people who have had this issue, and who have had to monkey patch their implementations in scary ways to resolve it.

### DIFF
--- a/src/gameobjects/BitmapText.js
+++ b/src/gameobjects/BitmapText.js
@@ -5,13 +5,13 @@
 */
 
 /**
-* BitmapText objects work by taking a texture file and an XML file that describes the font structure.
+* BitmapText objects work by taking a texture file and an XML or JSON file that describes the font structure.
 * It then generates a new Sprite object for each letter of the text, proportionally spaced out and aligned to 
 * match the font structure.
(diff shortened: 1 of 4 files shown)


photonstorm:
Nice one, thank you :) I think it would be worth adding something to the docs about HOW to actually convert the XML into JSON.

MaksJS:
```javascript
    jsonBitmapFont: function (game, json, cacheKey, xSpacing, ySpacing) {
        var data = {
            font: json.font.info._font,
            size: parseInt(json.font.info._size, 10),
            lineHeight: parseInt(json.font.common._lineHeight, 10) + ySpacing,
            chars: {}
        };
```
font: json.font.info._font should be font: json.font.info._face, as it is in the XML formated file.

Also you should try for "kernings" key existence cause it's not ever present in the XML.

For the doc, http://codebeautify.org/xmltojson seems to be a good tool.

PR #1838, opened 2015-06-06
new method Phaser.Tilemap#addTilesetBitmapData

unstoppablecarl:
added `Phaser.Tilemap#addTilesetBitmapData` method
allows setting a bitmapData object as an image of a Tileset. The bitmapData object must be cached. 

example

```js
var bitmapTileset = this.add.bitmapData(tileset.image.width, tileset.image.height, 'test', true);
// draw on bitmapTileset
tileMap.addTilesetBitmapData('test');
```

### DIFF
--- a/src/tilemap/Tilemap.js
+++ b/src/tilemap/Tilemap.js
@@ -359,6 +359,85 @@ Phaser.Tilemap.prototype = {
     },
 
     /**
+    * Adds a bitmapData object to the map to be used as a tileset. A single map may use multiple tilesets.
+    *
+    * @method Phaser.Tilemap#addTilesetBitmapData
+    * @param {string} [key] - The key of the Phaser.Cache bitmapData used for this tileset.
+    * @param {number} [tileWidth=32] - The width of the tiles in the Tileset Image. If not given it will default to the map.tileWidth value, if that isn't set then 32.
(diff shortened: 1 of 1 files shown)


unstoppablecarl:
I am not sure if this method can or should exactly match the signature and behavior of `Phaser.Tilemap#addTilesetImage` https://github.com/photonstorm/phaser/blob/master/src/tilemap/Tilemap.js#L254 .
Please advise.

photonstorm:
Thank you for the PR - it inspired me to modify the existing addTilesetImage to allow passing a BitmapData as an argument instead of a key. Sorry I didn't get to use your PR in this instance, but I wouldn't have made the update without it.

unstoppablecarl:
Thanks!

I think the argument and behavior variance of `addTilesetImage` is kinda getting out of hand. This function does too many different things. I found it confusing before and it took a fair amount of tinkering to figure it out. I would prefer separate functions that do more specific things like add a new tileset or update the image of existing one. 

At the very least I would like to see the newTileset creation  https://github.com/photonstorm/phaser/blob/14733e9b90e93399ae080b517142055f67f2882f/src/tilemap/Tilemap.js#L311 moved to a separate function. I would have easily been able to implement what I needed if that were separate.

PR #1851, opened 2015-06-12
Changes for nw.js-like environment detection code

rblopes:
This pull request proposes two small changes in the Device class:

1.  An alternative way to detect when Phaser is running under a NW.js (formerly Node-WebKit) environment, using feature detection, instead of relying on a guarded `require` statement.

    The former way is the source of a known incompatibility with `browserify` an similar tools.

2.  Add a complementary check for GitHub Electron.

References:

1.  [Changes related to node — nw.js Wiki][1]
2.  [Process object — Electron Documentation][2]

----
[1]: https://github.com/nwjs/nw.js/wiki/Changes-related-to-node#process
[2]: http://electron.atom.io/docs/v0.27.0/api/process/

### DIFF
--- a/src/system/Device.js
+++ b/src/system/Device.js
@@ -88,6 +88,12 @@ Phaser.Device = function () {
     this.nodeWebkit = false;
     
     /**
+    * @property {boolean} electron - Is the game running under GitHub Electron?
+    * @default
+    */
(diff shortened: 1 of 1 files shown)


photonstorm:
Looks good to me!

PR #1857, opened 2015-06-15
Implemented Keyboard.AddKeys function.

Mourtz:
Similar to createCursorKeys function but with user defined hotkeys and keycodes.T


photonstorm:
Nice, thanks :)

PR #1862, opened 2015-06-20
Fix tile index being off by one

noidexe:
With this.firstgid defaulting to 0 the tile index of every tile rendered is one more than it should be. With this.firstgid defaulting to 1 the problem seems fixed. 

It is true that the first tile id of a Tiled tileset is 0 but for phaser the same tile was always referenced by tile index 1.


noidexe:
This was fixed in 2.4

PR #1864, opened 2015-06-21
Fix small error in the Timer documentation

matugm:
Not sure if this is the correct branch for this change. The contrib section says that you should never PR against master, but I'm not sure if that applies for documentation as well.


photonstorm:
Hi - documentation is generated automatically from the source code jsdoc blocks, so corrections like this one need to be made in the source, not the output (otherwise they are lost when we next export).

PR #1867, opened 2015-06-22
Round linePositionX, to prevent placement of text on subpixel level which may cause blurriness

slashman:
When a multiline text object has align as 'center', text on lines other than the first one might be positioned at subpixel level, causing it to look blurry in low resolution rendering.


photonstorm:
This isn't always desirable behaviour, so I have placed it under a boolean flag.

Issue #1886, opened 2015-07-03
Attempting to clear a P2 collision callback with P2.Body.createGroupCallback fails


standardgaussian:
There is a minor typo in P2.Body.createGroupCallback which attempts to find a member "_groupCallbacksContext", whereas the actual member is "_groupCallbackContext", so a runtime error occurs when not passing a callback into the function.

photonstorm:
Fixed in dev.

PR #1889, opened 2015-07-05
 #1886

Langerz82:
https://github.com/photonstorm/phaser/issues/1886


photonstorm:
Thanks for this fix - but no PRs against master branch.

PR #1891, opened 2015-07-07
Handful of updates.  Fixes #1877 thanks @sanchopancho13

### LINKED ISSUE #1877: Game.renderer has the wrong typescript type
The Game "renderer" property is currently defined as a number type in the typescript definition file, however it is declared like this in the jsdoc:

{PIXI.CanvasRenderer|PIXI.WebGLRenderer}

In typescript 1.4 this can be declared as a union type.  Prior to that, I think the correct way to support that is with the 'any' type.

clark-stevenson:


photonstorm:
Thanks man. Am getting close to 2.4 final, so will keep trying to help out filling in the defs as I near it.

PR #1900, opened 2015-07-10
Feature: Phaser.RandomDataGenerator now can get/set its state

luckylooke:
Generator state can be saved now and it can continue after loading the game.

```javascript
var gen = new Phaser.RandomDataGenerator('myseed');
console.log(gen.rnd()); // first 0.4762128551956266
console.log(gen.rnd()); // second 0.21732503129169345
var stateOfGen = gen.state();
console.log(gen.rnd()); // third 0.5326935655903071
gen = new Phaser.RandomDataGenerator('myseed');
console.log(gen.rnd()); // first 0.4762128551956266
gen.state(stateOfGen);
console.log(gen.rnd()); // third 0.5326935655903071
```

### DIFF
--- a/src/tween/TweenData.js
+++ b/src/tween/TweenData.js
@@ -240,6 +240,13 @@ Phaser.TweenData.prototype = {
 
         this.startTime = this.game.time.time + this.delay;
 
+        //  DEBUG
+        this.realStart = Date.now();
+        this.totalElapsedMs = 0;
+
+        console.log('startTime', this.startTime);
+        console.log('realStart', this.realStart);
+
         if (this.parent.reverse)
         {
             this.dt = this.duration;
@@ -374,6 +381,13 @@ Phaser.TweenData.prototype = {
         }
 
         this.percent = this.dt / this.duration;
+
+        this.totalElapsedMs = (this.totalElapsedMs || 0) + this.game.time.elapsedMS;
+
+        if (this.percent === 1)
+        {
+            console.log("ElapsedTime", this.totalElapsedMs, Date.now() - this.realStart);
+        }
 
         this.value = this.easingFunction(this.percent);
 


luckylooke:
I almost forgot.. it can be also initialized with state
```javascript
var gen = new Phaser.RandomDataGenerator('myseed', stateOfGen);
```

pnstickne:
It doesn't seem 'right' to use `new Phaser.RandomDataGenerator('myseed', stateOfGen)` - that is, doesn't the state *encompass* the seed in a PRNG? I would expect something like `new Phaser.RandomDataGenerator(stateOfGen)` to suffice.

luckylooke:
True.. you can PR or wait few days I will change it ;)

luckylooke:
Hi @pnstickne, I have made your suggested change.. I also have tried to resolve conflict, but there is still "This branch has conflicts that must be resolved", I don't know why. It is because other files of my branch are outdated?

pnstickne:
@luckylooke There is a conflicting change currently between the current dev branch and the PR [dev] brnach.

Simply fetch the current dev branch and merge in changes. There will be conflicts that need to be resolved; fix the conflicts and push the PR branch again.

Anyway, since as you can see this fetch+merge could muck around with other local changes, I always do a `git checkout -b wip-name` from the latest dev before starting changes. (That creates a new branch and checks it out.)

luckylooke:
Creating new branch and place changes into it sounds easier... I will do it in a while.

pnstickne:
@luckylooke Also, what about returning / accepting the state as an opaque string value?

luckylooke:
@pnstickne please try to explain more.. I think I don't understand.

pnstickne:
ie. instead of returning an object a string would be returned. 'opaque' meaning the user should not care or have any guarantee about what the contents are, except that it can be used to restore the state. For instance maybe the result would be `"rng:DJSNN#$H!!@*!*#SDXVAASDFF"`.

luckylooke:
Aha.. I thought you mean that, but I don't understand the algorithm so deeply to return such string.

pnstickne:
@luckylooke The string can be whatever you want - as long as it's written and read back usefully. It doesn't even have to be cryptic, eg.

    stateStr = "!" + [s0, s1, ..].join("|')

    statStr = JSON.encode([s0, s1, ..])

    stateStr = base64encode(JSON.encode(..))

Which are all different - reversible - encodings I just made up. Ideally it would be come compact representation but that is less important than ensuring the state itself is kept as an opaque structure.

luckylooke:
@pnstickne how you distinguish new seed from state?

pnstickne:
@lucasmonteverde By using some form of value detection.

Currently seeds must be an array (of any value); thus the state might simply be a string - and a string that follows a particular encoded value format or 'magic' header.

For example:

    if ((typeof seeds === 'string' || seeds instanceof String) && seeds.match(/^!rng/) {
       // seeds is a string of the captured internal state

luckylooke:
Ok, I will rewrite it that way hopefully in few days.

luckylooke:
@pnstickne please tell me the advantage of using string representation of state. Thanks

pnstickne:
@luckylooke It creates an *opaque value* - think of a GUID. There is no point caring about what it is (or what it contains); only what it represents is important. This small opaque value is - should ideal be created such, already - densely encoded and can appear unobtrusively in other serialization contexts where a primitive value is accepted.

For example, a save file might look like:

````
{
    days: 200,
    worldGenerator: "!bc67268941194c5e8f565b402ef8c90c"
}

luckylooke:
@pnstickne Ok, perfectly understand now.. I will update code soon. Thanks for lesson :+1:

PR #1905, opened 2015-07-11
AABB vs. AABB collision in Ninja

standardgaussian:
Nearly all of the work required to get AABB vs. AABBs to work in Ninja was already present.

reportCollisionVsWorld already worked, and contained all of the logic required to resolve a collision once the appropriate vectors had been established. reportCollisionVsBody was refactored to use that function (now generically named reportCollision), and now AABBs can collide properly, including bouncing and friction. reportCollisionVsWorld is now just a wrapper around reportCollision to maintain compatibility.

You can grab a testfile based on the phaser tutorial demo [here](https://github.com/standardgaussian/phaserDemos/blob/master/phaserDemoNinja.js). All the entities are Ninja AABBs.


photonstorm:
Lovely, thank you.

PR #1923, opened 2015-07-21
Make PIXI available for Phaser when using require

mkristo:


photonstorm:
Thanks - would you mind sending me a tiny example of how you'd include Phaser using require now so I can update the resources folder. Just the html file will do - doesn't matter what the 'game' does, even if it just loads a single image that will be enough.

mkristo:
We are using webpack, so a simple html file wouldn't be enough. But I'll try to create a minimal example for this.

mkristo:
Here's a pull request of a simple webpack boilerplate. https://github.com/photonstorm/phaser/pull/1924

Using the full Phaser bundle in the webpack boilerplate crashes with the same error for p2 as earlier for PIXI.

PR #1929, opened 2015-07-22
Fix cache.GetKeys() - Use _cacheMap to map from constant to _cache.

jamesgroat:
Fix cache.getKeys(), correctly use _cacheMap instead of _cache with the passed in constant.


photonstorm:
:+1:

PR #1934, opened 2015-07-24
Fix unmatched context.save() & context.restore() calls

MortimerGoro:
Fix unmatched context.save() & context.restore() calls in the redraw optimization return

```
    if (!redrawAll &&
        shiftX === 0 && shiftY === 0 &&
        mc.renderWidth === renderWidth && mc.renderHeight === renderHeight)
    {
        //  No reason to redraw map, looking at same thing and not invalidated.
        return;
    }
```

As a optimization, I have moved the save after the return instead of adding a restore before the return because there are no state changes until later.

### DIFF
--- a/src/tilemap/TilemapLayer.js
+++ b/src/tilemap/TilemapLayer.js
@@ -997,8 +997,6 @@ Phaser.TilemapLayer.prototype.render = function () {
         return;
     }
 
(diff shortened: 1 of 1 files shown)


photonstorm:
:+1:

Issue #1935, opened 2015-07-24
cache.hasFrameData ignores cache type


standardgaussian:
Usage of cache.hasFrameData suggests that you can choose your cache type, but the actual function throws away that parameter and always checks Phaser.Cache.IMAGE instead, leading to errors when trying to add animations when using a bitmapdata key.

photonstorm:
`hasFrameData` doesn't have a cache parameter and it says in the docs specifically that it only checks the Image cache, so I don't believe this is an error here - however I do agree it's a useful feature, so I've added it in.

PR #1936, opened 2015-07-24
#1935: cache.hasFrameData ignores cache type

standardgaussian:
The cache is being heavily refactored, so I'm not sure if this PR is appropriate. Is the intent to only allow items stored in the Phaser.Cache.IMAGE portion of the cache to have frame data?

components.loadTexture suggests that getFrameData and setFrameData can take a parameter for which cache to check, eg. line 62. Is that being phased out?


photonstorm:
Thank you for the PR, but this has been rolled into the 2.4.2 release already.

PR #1957, opened 2015-07-30
Fix p2 BodyDebug pixelsPerLengthUnit bug.

vrecluse:
When mpx doesn’t return default value, the debug body size and position
goes wrong.


photonstorm:
:+1:

PR #1963, opened 2015-07-31
A Grunt `custom` task module.

rblopes:
To declutter the main `Gruntfile.js`, move your custom task code to a separate module.

Other few improvements included in this patch:
- Upgraded some Grunt task plugins related to the build process;
- Enabled Grunt JIT, for (slightily) faster builds.

2015-10-05: Rebased to include latest development commits.


photonstorm:
Sorry I sat on this one for so long - honestly, once the tasks are working I'm reluctant to go and change them all again without good reason. I like what is being done here, and will remember it for the future, but don't feel it's worth making this level of change at this stage now, sorry.

rblopes:
Wow, so much happened after this I almost forgot it still existed :)

From what I remember, this is a "follow up" to an attempt at improving the building system, but the way I edited the commits before was a complete mess. After trying to put things in place a couple times or so, I decided to decline the former PR in place of more focused, smaller edits, I'd deliver to incorporate the former changes.

I'm sure the edits here were just to move the task code inside another module, the task code itself and what it does wasn't changed at all. I'm not sure how much of this is up to date, though. And, even if the changes were moved into Phaser now, I doubt I have the energy to keep up doing this stuff at the moment.

Since preparing Lazer for its first release being your primary focus now, I perfectly understand bringing these kind of changes into Phaser doesn't make much sense. I'm also looking forward for Lazer too, getting very anxious already.

Thanks for replying back and for all the good time proportioned by Phaser. Not my last contribution here, though next ones won't try to "change the world" either.

My best regards.

PR #1966, opened 2015-08-01
Micro Definition Update + Readme.

clark-stevenson:
You caught almost all the changes already :+1:  Updated readme. I will get to these missing defs someday.


photonstorm:
:+1:

PR #1967, opened 2015-08-01
Fix misleading P2.Body.applyImpulseLocal comment.

vrecluse:
The comments of applyImpulseLocal said the arguments are in world space, but the code is in local space.
BTW, the force argument is represented in p2's coordinates, but the position in phaser's.It seems very confusing.


photonstorm:
Thanks, merged locally.

Issue #1976, opened 2015-08-04
imageData.data from a bitmapData does not contain pixel values


yahiko00:
This issue comes from a [post on the forum](http://www.html5gamedevs.com/topic/16215-cannot-access-imagedatadata-from-a-bitmapdata).

I would like to apply a filter on a picture with a `bitmapData` as a proxy to apply pixel transformations.

The problem comes from the array **`imageData.data`** of a `bitmapData` which is filled with zeros instead of being filled with correct values.

Here is a piece of code to illustrate my issue:

    var originalPic = game.add.sprite(0, 0, 'originalPic');
    var bmd = game.add.bitmapData(originalPic.width, originalPic.height);
    bmd.draw(originalPic);
    
    console.log(bmd.imageData.data); // ==> [0, 0, 0, 0, ...]
    
    var newPic = game.add.image(600, 0, bmd); // Display on screen is okay

Obviously, if I cannot access those data, I would not be able to make any computation on pixels. :/

I came up with a solution by using the **`context`** attribute of a `bitmapData`, instead of using directly the `imageData` attribute.
  
    var originalPic = game.add.sprite(0, 0, 'originalPic');
    var bmd = game.add.bitmapData(originalPic.width, originalPic.height);
    bmd.draw(originalPic);
    var imageData = bmd.context.getImageData(0, 0, originalPic.width, originalPic.height);
    
    console.log(imageData.data); // ==> now contains correct values
    // Here I can call my image processing function
    // ...
    
    var newPic = game.add.image(600, 0, bmd); // Display the processed image

So, `BitmapData.imageData` is quite buggy. I am going to report an issue about this on GitHub.

 Thank you for having read me.

photonstorm:
If you don't call `BitmapData.update` after manipulating it, it won't update the `imageData` property. This isn't done automatically because it's expensive, so you have to call it yourself.

yahiko00:
It works. Thank you for your answer.
The call to `BitmapData.update()` should be mentionned in the documentation so, under the `imageData` entry.
This would have saved me couple of hours understanding where was the problem. ^_^

PR #1983, opened 2015-08-06
Add extra information for the imageData property. Update TypeScript definition file.

yahiko00:
This pull request is related to this [issue #1976](https://github.com/photonstorm/phaser/issues/1976).

### DIFF
--- a/docs/Phaser.BitmapData.html
+++ b/docs/Phaser.BitmapData.html
@@ -1886,7 +1886,8 @@ Note that this is unavailable in some browsers (such as Epic Browser due to its 
 
     
     <div class="description">
-        <p>The context image data.</p>
+        <p>The context image data.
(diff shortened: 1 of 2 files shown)


photonstorm:
Hi - Thanks for the PR, but a couple of points: 1) All PRs against the dev branch, not master. And 2) The docs are automatically generated from the jsdoc blocks inside the source code, so don't edit html files or they get nuked when I next export the docs. Instead edit the jsdocs in the source code.

Also the comment is 100% correct :) I should have been more specific in my original reply to you: If you want to access the imageData *immediately* after drawing then you need to call update. However if you can wait for a single game loop to complete then it will automatically get updated. Drawing to a BitmapData flags the texture as being dirty, when a Sprite that uses a BitmapData as a texture is updated it sees the dirty flag and runs update on it. So it's a bit of a mixture what happens when, and when you need it.

yahiko00:
For the comment, I figured it out right after sending my PR.
I gonna add the comment in the JS doc, with this new information of yours. Thanks ;-)

PR #1993, opened 2015-08-09
Fix for fullscreen retina ( Desktop )

nickryall:
On my Retina display Macbook, fullscreen was showing double the scale. This fixes the issue.

### DIFF
--- a/src/core/ScaleManager.js
+++ b/src/core/ScaleManager.js
@@ -1358,8 +1358,8 @@ Phaser.ScaleManager.prototype = {
             }
             else if (scaleMode === Phaser.ScaleManager.NO_SCALE)
             {
-                this.width = this.game.width;
-                this.height = this.game.height;
+                this.width = this.game.width / this.game.device.pixelRatio;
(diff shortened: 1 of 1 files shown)


Cristy94:
I tried it on tablets with retina and everything was ok with the current version. I set the game width to window.innerWidth * window.devicePixelRatio and set the scalemanager to fit the window:
```
this.game.scale.scaleMode = Phaser.ScaleManager.SHOW_ALL;
this.game.scale.pageAlignHorizontally = true;
this.game.scale.pageAlignVertically = true;
```

Doesn't this work for you with the current version?

nickryall:
This one was specific to Desktop Chrome which has support for the fullscreen API  - on a Retina Macbook.  On IOS and Android everything looks okay ( although fullscreen API isn't supported for this at least for IOS, I'm not sure about Android, will check).

photonstorm:
I'm going to have to revert this merge because it completely breaks the game size on Retina screens when not in full-screen mode. The game size is being halved in all cases, which isn't right.

PR #1998, opened 2015-08-11
Adding missing event handlers in Typescript definitions

cloakedninjas:


photonstorm:
Thanks for this. Please don't edit the TS Comments defs directly though - as they are automatically generated by the build process, so all manual edits are overwritten.

PR #2001, opened 2015-08-13
(fix) BitmapText font change

AbrahamAlcaina:
Hi, 
2.4.2 Have an issue when changing the font, it doesn't refresh properly (appears the old one) .
To fix it, is needed to load the new one.
```javascript
this._data = game.cache.getBitmapFont(this._font);
```
just before calling updateText


photonstorm:
Thanks for this. I've merged (and fixed) your change locally (as you were using a global `game` var which needed replacing).

PR #2013, opened 2015-08-19
Remove unexistant parameter documentation

STuFF:


photonstorm:
Can't merge because it's against master, but fixed in dev anyway, thanks!

PR #2014, opened 2015-08-20
Update README.md

crumpled:
Remove reference to https://github.com/photonstorm/phaser/archive/master.zip and Koding. Link results in 404


photonstorm:
Hmm the link doesn't 404 for me at all - it just goes to the Koding home page, yes, but not a 404.

photonstorm:
I've updated the readme to reflect this.

PR #2018, opened 2015-08-22
Fixes #2017

### LINKED ISSUE #2017: Not fully implemented Phaser.Cache.addSpriteSheet
Hello, documentation of method `Phaser.Cache.addSpriteSheet` says `frameMax`, `margin`, `spacing` has a default value, but actually values are written as null. For this reason don't draw a sprite created with `addSpriteSheet` method.

See: https://github.com/photonstorm/phaser/blob/master/src/loader/Cache.js#L560

vladkens:


photonstorm:
I can't merge because this is against the wrong branch, but I've fixed this locally in dev. Thanks.

PR #2022, opened 2015-08-23
Fixed pixel rounding in PIXI.TilingSprite.prototype._renderCanvas.

ximop:
Changed "x|0" to "x |= 0" in PIXI.TilingSprite.prototype._renderCanvas.


photonstorm:
Thanks for the fix. I can't merge because you issued the PR against master, but have fixed it locally in dev.

PR #2025, opened 2015-08-25
Optimize the average function in math.js

rwrountree:
Instead of querying arguments.length repeatedly, save it out into another variable and use it in the body of the average function.

Phaser.Math.fuzzyCeil and Phaser.Math.fuzzyFloor should not return boolean values as stated in @return section of the comments


photonstorm:
:+1:

Issue #2032, opened 2015-08-26
Array.prototype.splice requires two parameters


shank-eric:
The spec for Array.prototype.splice technically requires two parameters (start and deleteCount) although it appears that most browsers support a call with a single parameter which removes everything after the start and returns it.  However, when using es5-shim with Safari, this style of call does not function the same and instead returns an empty array.  In Phaser, there are several calls structured like the following with only a single parameter being passed to `splice`:
```javascript
        var args = Array.prototype.splice.call(arguments, 1);
```

In my specific use-case this ends up resulting in an error when trying to click an element as can be seen here:  http://everfi-curriculums.s3.amazonaws.com/curriculums/nhl-future-goals/1_0_0/index.html#painttheice/painttheice-activity/page-7 .  The blue arrow for the instructions doesn't advance to the next set of instructions when clicked and an error is generated (`TypeError: undefined is not an object (evaluating 't.id')`).  This is because the `pointer` parameter passed to `_releasedHandler` is `undefined` which is caused by the call to `splice` in `callAll` returning an empty array. 

I believe the simplest solution would be to include a second parameter in these calls to `splice` to make it match to the spec.  My understanding is passing a value that extends past the end of the array will return all elements to the end of the array, so something like this should work:
```javascript
        var args = Array.prototype.splice.call(arguments, 1, arguments.length);
```

pnstickne:
All `splice.call`'s on *arguments* should probably be `slice.call`s - arguments and aliasing has some de-optimizations if it's mucked with.

Working on some jsperf test-cases; which should reveal if this change is outright 'OK' from any sort of performance view.

pnstickne:
Well, okay. Most (all) of the `splice.call(arguments, N)` code is incorrect. It replaced `slice.call(arguments, N)` and should have been `splice.call(arguments, 0, N)` to keep the similar semantics.

However *on `arguments`*, using `slice.call` is better across all desktop browsers (IE, FF, Chrome) by varying degrees - probably due to aliasing and de-optimizations. See http://jsperf.com/splice-vs-splice/2 (only the Slice-Arguments and Splice-Arguments results are relevant; sizes of the arrays and removals have been chosen based loosely on what might be expected).

pnstickne:
That change should affect all `splice.call`.

Hopefully it fixes the problem observed. However - at least in Chrome and FF - `splice` will "work" even when the length is omitted.

[...]

PR #2034, opened 2015-08-27
Change splice.call(arguments, ..) to use slice

pnstickne:
- Addresses concerns in #2032 and 'slight optimization'

- Bypasses issue of usage incorrectly omitting 2nd argument to `splice`

- More clear of intent; `slice` does not modifify `arguments`

- `slice` *on `arguments`*, with practical lengths, is faster across all desktop browsers by varying degrees
  - Probably due to parameter-aliasing and de-opts when modified.
  - Both `slice` and `splice` create a new Array object

### DIFF
--- a/src/core/PluginManager.js
+++ b/src/core/PluginManager.js
@@ -52,7 +52,7 @@ Phaser.PluginManager.prototype = {
     */
     add: function (plugin) {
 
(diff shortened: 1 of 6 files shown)


photonstorm:
:+1:

PR #2035, opened 2015-08-27
Fix TypeScript `addKeys` declaration

timotei:
Since we expect an object that's `any`
instead of an array `any[]`


photonstorm:
:+1:

Issue #1990, opened 2015-08-09
Text with anchor(0.5) not centered. 


nickryall:
It appears to be off to the left.  This can be seen in the example:  http://phaser.io/examples/v2/text/center-text-on-sprite

This is new to v2.4

pnstickne:
In the link the `sprite` appears to be to the left, but the text appears to be centered - *on* the sprite.

I've tested in recent Chrome and FF.

nickryall:
For me the text appears to be around 32px from the left edge of the sprite and 42px from the right. I think it should be 37px from either side.

photonstorm:
Welcome to the inaccuracy that is the HTML5 Canvas Context `measureText` function. If you'd like to fix this across all browsers for us then your text will suddenly and magically be perfectly centred!

nickryall:
Sure I can take a look :) The main reason for me logging this issue is that it only became apparent in my game since v2.4.

photonstorm:
Sorry I wasn't too clear - the discrepancy comes from the values `measureText` returns. Nothing changed in the way the Text width is calculated in Phaser 2.4, but that doesn't mean Chrome etc didn't tweak it slightly internally. Who knows.

nickryall:
Ah okay I see what you mean now. Ok, well I'll dig into it a little more anyway. Thanks

photonstorm:
I did think that it could use a pixel scan method - i.e. you create a canvas that's a bit too wide, draw the text to it, then scan every line to find out the furthest most pixel. Then it could resize the canvas and re-render the text so it would be perfect. This would account for the way different OSs and different browsers render fonts different too - but it'd be an insanely expensive operation too. Can't win them all :(

nickryall:
I looked into this further this morning and I found the following line to be what was causing the issue:

    var lineWidth = this.context.measureText(lines[i]).width + this.style.strokeThickness + this.padding.x;

By removing
    
     this.style.strokeThickness

from the line width calculation, the text appeared to be centered again.  So basically any text with stroke applied was appearing further to the left than it should when centered.

My game does have some custom scaling going on which could be causing this, so I will see if I can reproduce this on a basic example.

photonstorm:
While I agree with the strokeThickness bug and will merge that, it still won't solve the issue fully. It makes no difference to the example you linked to in your first post. Use the `backgroundColor` style property to easily view the problem:

[...]

PR #2039, opened 2015-08-27
Fix for #1990 - Remove duplicate addition of strokeThickness when calculating text width

nickryall:
Fix for #1990


photonstorm:
This was merged locally into dev.

PR #2043, opened 2015-08-28
Updated typescript definitions for Phaser 2.4.3

Garbanas:
Just updated the phaser.comments.d.ts file, since it has not been updated in a while.


qdrj:
@Garbanas please check definitions for `interface CachedImage`. There should be semicolons instead of commas.

Garbanas:
@qdrj Thanks for the hint.

photonstorm:
For some reason the grunt task that auto-generates the TS Comments defs file no longer runs. I spent a while looking at it but couldn't resolve it, so had to leave it be. Ideally we'd fix the grunt task rather than patching the file directly like this.

PR #2044, opened 2015-08-28
Prevent 'data:' URLs from being prefixed

rblopes:
Fixes an issue where 'data:' URLs may get prefixed by `#baseURL` and `#path` properties making these URLs invalid.

### DIFF
--- a/src/loader/Loader.js
+++ b/src/loader/Loader.js
@@ -2028,7 +2028,7 @@ Phaser.Loader.prototype = {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


pnstickne:
Don't forget about `blob:` prefixes, if those can make it in.

I would imagine a pattern might look like:

    if(url.match(/^(?:\w+:/)?\/\/))  // then already has a protocol or is protocol-relative, so not relative to base

rblopes:
@pnstickne Thanks for suggesting these.

photonstorm:
Yup, nice one.

PR #2047, opened 2015-08-29
Convenience function to remove a collision group from a P2 Body

Garbanas:
This removes a given collision group or a list of collision groups from a P2 Body without removing the other collision groups. Collision masks are updated and callback functions are either removed or kept, depending on user choice.

We're currently in production of a game where the player can enter a "ghost mode" and doesn't collide with obstacles anymore. This reduces the hassle of removing a collision group and adding it back with body.collide().


photonstorm:
:+1:

Issue #2031, opened 2015-08-26
Put the keyboard constants in the documentation


drhayes:
To get at them currently, you have to dig in the source. Not the most pressing issue, but inconvenient.

beagleknight:
I have added a PR with your suggestion. The only problem I have detected is now is hard to check the `Phaser.Keyboard` documentation. I could be useful if jsdoc have an option to include a separate section for them.

photonstorm:
Thanks for the PR. This is now resolved with the new KeyCode file.

PR #2051, opened 2015-09-01
Add jsdoc comments to Phaser.Keyboard key codes

beagleknight:
As suggested by @drhayes on issue #2031 I have added the jsdoc comments to include key codes on Phaser documentation.

I don't know too much about jsdoc but I couldn't find an option to include those key codes on another section. I think the Phaser.Keyboard docs became a bit messy :) after adding those comments.

### DIFF
--- a/src/input/Keyboard.js
+++ b/src/input/Keyboard.js
@@ -473,7 +473,7 @@ Phaser.Keyboard.prototype = {
     /**
     * Returns `true` if the Key was pressed down within the `duration` value given, or `false` if it either isn't down,
     * or was pressed down longer ago than then given duration.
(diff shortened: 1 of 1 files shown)


pnstickne:
Unfortunately the keys are 'part' of the actual Keyboard type itself and not separate.

From a documentation point it _may_ be better to create a pseudo-type and then link to it, explaining such. I thought there was already such a type setup, but I cannot find it, so I do not even know if this is currently feasible/recommended.

photonstorm:
Yeah these all exist in the `Phaser.KeyCode` namespace now, so closing off the old PR, sorry.

PR #2054, opened 2015-09-02
Added useCapture flags to removeEventListener where approriate

pmcmonagle:
Event listeners added with `true` passed in as the useCapture flag are only removed when `true` is also passed into the `removeEventListener` call. Adding this flag to the stop method where appropriate fixes a memory leak in IE where events on window are never removed.

See https://msdn.microsoft.com/en-us/library/ff975250(v=vs.85).aspx


pmcmonagle:
To clarify, I encountered this issue while using Phaser with AngularJS.

On Chrome we were able to clean up the game instance with no problems.
On IE11, a reference to the game instance was being held by the MSPointerUp event listener attached to window. The above change allows the MSPointerUp event to be released on game.destroy and allows the game instance to be properly garbage collected.

photonstorm:
Thanks but PRs against dev branch only.

PR #2055, opened 2015-09-02
Added useCapture flags to removeEventListener where approriate

pmcmonagle:
Event listeners added with true passed in as the useCapture flag are only removed when true is also passed into the removeEventListener call. Adding this flag to the stop method where appropriate fixes a memory leak in IE where events on window are never removed.

See https://msdn.microsoft.com/en-us/library/ff975250(v=vs.85).aspx

### DIFF
--- a/src/input/MSPointer.js
+++ b/src/input/MSPointer.js
@@ -425,20 +425,23 @@ Phaser.MSPointer.prototype = {
 
         var canvas = this.game.canvas;
 
-        canvas.removeEventListener('MSPointerDown', this._onMSPointerDown);
-        canvas.removeEventListener('MSPointerMove', this._onMSPointerMove);
-        canvas.removeEventListener('MSPointerUp', this._onMSPointerUp);
+        canvas.removeEventListener('MSPointerDown', this._onMSPointerDown, false);
+        canvas.removeEventListener('MSPointerMove', this._onMSPointerMove, false);
+        canvas.removeEventListener('MSPointerUp', this._onMSPointerUp, false);
+
+        //  IE11+ uses non-prefix events
+        canvas.removeEventListener('pointerdown', this._onMSPointerDown, false);
+        canvas.removeEventListener('pointermove', this._onMSPointerMove, false);
+        canvas.removeEventListener('pointerup', this._onMSPointerUp, false);
+
+        window.removeEventListener('MSPointerUp', this._onMSPointerUpGlobal, true);
         canvas.removeEventListener('MSPointerOver', this._onMSPointerOver, true);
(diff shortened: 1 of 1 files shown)


pmcmonagle:
To clarify, I encountered this issue while using Phaser with AngularJS.

On Chrome we were able to clean up the game instance with no problems.
On IE11, a reference to the game instance was being held by the MSPointerUp event listener attached to window. The above change allows the MSPointerUp event to be released on game.destroy and allows the game instance to be properly garbage collected.

pnstickne:
The events are not always registered with use capture (and usually are not).

It seems like the removal should always mirror the registration.

pnstickne:
Hmm, the double-registrations look odd. That should be revisited, perhaps.

Anyway, what about specifying `false` explicitly for the non-capture event removal to match the registrations?

pmcmonagle:
I'm not sure what you mean by double-registrations for pointerover. Every event is registered and deregistered twice: once with the MS prefix and once without. The ones with/without the prefix are simply ignored depending on the IE version.

Specifying `false` on the other events is a fine idea, but improving readability isn't really my concern. The PR as I've submitted it fixes a show-stopper memory leak which affects an application I have in production right now. My only concern is getting that specific fix into the next Phaser version so that I can stop using a custom Phaser build.

pnstickne:
Oh, just hard to follow matches. Taking the registrations and removing comments/if it is 

            canvas.addEventListener('MSPointerDown', this._onMSPointerDown, false);
            canvas.addEventListener('MSPointerMove', this._onMSPointerMove, false);
            canvas.addEventListener('MSPointerUp', this._onMSPointerUp, false);  

            canvas.addEventListener('pointerdown', this._onMSPointerDown, false);
            canvas.addEventListener('pointermove', this._onMSPointerMove, false);
            canvas.addEventListener('pointerup', this._onMSPointerUp, false);

             window.addEventListener('MSPointerUp', this._onMSPointerUpGlobal, true);
             canvas.addEventListener('MSPointerOver', this._onMSPointerOver, true);
             canvas.addEventListener('MSPointerOut', this._onMSPointerOut, true);

             window.addEventListener('pointerup', this._onMSPointerUpGlobal, true);
             canvas.addEventListener('pointerover', this._onMSPointerOver, true);
             canvas.addEventListener('pointerout', this._onMSPointerOut, true);

Then a Ctrl+R would make the required change change from `addEventListener` to `removeEventListener` would fix the bug and keep a clear paring.
.

pmcmonagle:
Oh yeah, I get what you mean now. Yeah, that doesn't sound like a bad idea. I'll throw it in.

pnstickne:
I think it looks good. Thanks for the bugfix and updates.

PR #2058, opened 2015-09-04
Adding Point.empty and Rectangle.empty static methods

iliak:
This allows you to create empty Rectangle and Point classes.


pnstickne:
FWIW, the same effect can be achieved with `new Point()` and `new Rectangle()`.

iliak:
I'm porting my game from XNA to Phaser and this would help other projects, but you are right.

PR #2065, opened 2015-09-07
fix #2062

### LINKED ISSUE #2062: button.events.onInputOver doesn't fire for a short while after clicking on canvas.
After you click anywhere on the game canvas for short while you can move the mouse over a button and onInputOver won't fire and the button won't change to the 'over' frame.

This is even the case with very simple code, with nothing special coded into the buttons. 
For example I used this demo http://phaser.io/examples/v2/buttons/action-on-click and made a short video with clicking audio cues to demonstrate the problem.
https://youtu.be/F_L6-UkeX0A

j27cai:
Added an option to ignore the pointer clicking delay in a button over event.

### DIFF
--- a/src/gameobjects/Button.js
+++ b/src/gameobjects/Button.js
@@ -469,20 +469,8 @@ Phaser.Button.prototype.setUpSound = function (sound, marker) {
 * @protected
 * @param {Phaser.Button} sprite - The Button that the event occurred on.
 * @param {Phaser.Pointer} pointer - The Pointer that activated the Button.
-* @param {boolean} fastClick - ignore pointer delay on input over a button if this is true
 */
-Phaser.Button.prototype.onInputOverHandler = function (sprite, pointer, fastClick) {
-
-    if (fastClick === undefined)
-    {
-        fastClick = false;
-    }
-
-    //  If the Pointer was only just released then we don't fire an over event
-    if (!fastClick && pointer.justReleased())
-    {
-        return;
-    }
+Phaser.Button.prototype.onInputOverHandler = function (sprite, pointer) {
 
     this.changeStateFrame(STATE_OVER);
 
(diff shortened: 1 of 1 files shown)


pnstickne:
It can be simplified as

    if(!fastClick && pointer.justReleased()) { ..

Anyway, I wonder what the primary point of the timeout check is/was..

src/gameobjects/Button.js
```diff
@@ -470,12 +470,20 @@ Phaser.Button.prototype.setUpSound = function (sound, marker) {
 * @param {Phaser.Button} sprite - The Button that the event occurred on.
 * @param {Phaser.Pointer} pointer - The Pointer that activated the Button.
```
stronglink: `fastClick` should be [documented](http://usejsdoc.org/tags-param.html) here. Something along the lines of ` * @param {boolean} [fastClick=false] - Whether an event should be fired when...`

pnstickne:
Then it makes me wonder what the use case(s) is(are) for 'slow clicks' - and maybe if such should be handled in the button itself (and thus handled *per* button, with the ability to recover from a re-hover) vs deferral to the input handling.

j27cai:
After looking at it a bit more, I can't seem to understand what use there is in delaying the clicks. Perhaps, even the deletion of the condition might be okay?

pnstickne:
I'm not sure why it was added either .. but it is there, so presumably there was a case for it. I suspect it might have been / might be to prevent 'resetting to hover state' if a click is immediate followed by a hover-move.

pnstickne:
I'm actually inline to say the check should be removed entirely - ie. revert https://github.com/photonstorm/phaser/commit/96fd0ade

I cannot find any mention of what bug it fixes, if any. This includes that commit and the following Phaser (2.3.1) release documentation.

pnstickne:
Anyway, this commit won't work right (and I should have spotted it sooner) because the `fastClick` parameter, coming from an Input event, will never be anything but undefined.

j27cai:
So, it's probably in best interest to delete the condition down right.

PR #2069, opened 2015-09-08
Fix for AudioBufferSourceNode listener memory leak.

Pappa:
In Chrome, the AudioBufferSourceNode onended listeners were never being garbage collected. This frees up the listener for collection.

More information here: http://www.html5gamedevs.com/topic/16903-problem-with-listeners-using-phasersoundonplayadd/


photonstorm:
Lovely.

PR #2072, opened 2015-09-09
Initial Noise Class implementation

codevinsky:
Simplex and Perlin noise


photonstorm:
I know this is months old now, but I'd love to see this in ES6 in Lazer, rather than padding out Phaser, if you ever get a whim to do such a thing :)

codevinsky:
You got it, boss.

PR #2073, opened 2015-09-09
Definition error in Keyboard.addKeys()

nexiuhm:
Now correctly takes an object as the argument instead of array.


nexiuhm:
Closing this since someone else already fixed it.

PR #2080, opened 2015-09-12
Adds better Button 'pressed state' handling and onClick signal

### LINKED ISSUE #2077: Re-hovering over a pressed button goes to 'Hover' state; expecting 'Down'
Per http://phaser.io/examples/v2/buttons/action-on-click - Phaser 2.4.x

- Click on the button and keep the mouse button pressed.
  - The button shows 'down' - GOOD
- Move the mouse off the button
  - The button image shows 'out' - OKAY, but confusing if `isOver` not checked in click-action/onUp callback. (But I think this is a separate issue.)
- Move the mouse back on the button.
  - The button image changes to 'over' - NOT EXPECTED. The button should go back to the 'down' state as the mouse/pointer is still down.

pnstickne:
- Impact: low/minimal, this is to address unexpected visual state.

- When a button is pressed (DOWN) and the pointer is moved OUT
  and then back over will trigger a DOWN frame/sound and *not*
  a OVER state/sound.

  - The onOver even is still triggered.

- Designed to work in multi-touch cases.

- Fixes #2077

### DIFF
--- a/src/gameobjects/Button.js
+++ b/src/gameobjects/Button.js
@@ -18,6 +18,8 @@
 *
 * Frames can be specified as either an integer (the frame ID) or a string (the frame name); the same values that can be used with a Sprite constructor.
 *
+* If a callback is supplied it is bound to the {#onClick}} signal.
+*
 * @class Phaser.Button
 * @constructor
 * @extends Phaser.Image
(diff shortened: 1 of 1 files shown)


pnstickne:
https://github.com/pnstickne/phaser/commit/5d8d4e5cc70dd96505a70d2cb294afe4a494242f ('Added Button.onClick signal') is a minor breaking change, although arguably for the better.

When the "action callback" is attached via the constructor is is associated with onClick - and *not* onUp. To handle the (unaffected) onUp event it must be bound to the signal explicitly.

This can cause slightly different - more correct for a button in general - results when existing code was relying on non-checked onUp behavior when the action was attached from the constructor.

PR #2081, opened 2015-09-13
Fixes incorrect "Pointer" type documentation and parameter names in InputHandler

pnstickne:
- Impact: none; documentation and local variable names only

- Incorrect documentation that claimed to take a Pointer when they really
  took a Pointer ID is fixed

- Also updates parameter names for consistenct/clarity


pnstickne:
Ref #2082

PR #2084, opened 2015-09-13
Add support for TiledJSON with base64 encoding

noidexe:
Tiled 0.13.0 added support for layer data compression when exporting as LUA or JSON. This means that any .tmx stored unsing base64 encoding will start exporting layer data as a base64 encoded string rather than a native array. 
Phaser would try to load the level anyway without emitting any errors, resulting in a corrupted level while playing. 
This PR adds detection and support for base64 encoded levels as long as they don't use compression.

Tiled commit adding base64 encoding: https://github.com/bjorn/tiled/commit/26a4675929e1622331a9b316291ee9283fc52a65

### DIFF
--- a/src/tilemap/TilemapParser.js
+++ b/src/tilemap/TilemapParser.js
@@ -232,6 +232,19 @@ Phaser.TilemapParser = {
             }
 
             var curl = json.layers[i];
+            
+            // Base64 decode data if necessary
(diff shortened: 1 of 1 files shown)


photonstorm:
Nice, thanks for this. I guess we'll need a `window.atob` polyfill for IE9 though.

PR #2089, opened 2015-09-14
Correctly interpret binaryString as a sequence of uint32-le values

noidexe:
The previous implementation was ignoring the 3 most significant bytes of each value and would result on improper parsing of any map with tile ids higher than 255

### DIFF
--- a/src/tilemap/TilemapParser.js
+++ b/src/tilemap/TilemapParser.js
@@ -239,8 +239,13 @@ Phaser.TilemapParser = {
                 var binaryString =  window.atob(curl.data);
                 var len = binaryString.length;
                 var bytes = new Array( len );
+                // Interpret binaryString as an array of bytes representing
+                // little-endian encoded uint32 values. 
                 for (var i = 0; i < len; i+=4) {
-                    bytes[i/4] = binaryString.charCodeAt(i);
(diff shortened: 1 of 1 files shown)


pnstickne:
Well, it's still not "correct" although it doesn't appear to be any worse - see https://mathiasbynens.be/notes/javascript-encoding The big takeaways is chatCodeAt works on code units (ie. surrogate pairs included) and not unicode code points..

noidexe:
@pnstickne What about using codePointAt() instead of charCodeAt() and do a zero-fill right shift on the result? It seems to work: 
``` 
var str = String.fromCharCode(255,255,255,255);
var num = (str.codePointAt(0) | str.codePointAt(1) << 8 | str.codePointAt(2) << 16 | str.codePointAt(3) << 24) >>> 0;
console.log(num); //4294967295
```

pnstickne:
@noidexe  Oh, hmm, never-mind previous comment.
It appears I was confused on the usage.

pnstickne:
@neurofuzzy My previous comment was incorrect.

For some reason I thought (wrongly) that it was being encoded back into a string; but that code just chucks off an array of values.

noidexe:
The ```>>> 0``` part seems to be necessary, though I don't know how many people will have tile ids over 2147483647. I'll add it anyway.

PR #2091, opened 2015-09-15
Fix for tile ids between 0x7FFFFFFF and 0xFFFFFFFF

noidexe:
2147483647 tile ids should be enough for anybody, but it is now working for any valid tile id.


photonstorm:
:)

PR #2093, opened 2015-09-16
(Resource: PhysicsEditor Exporter) Add option to prefix shape name and optimize JSON output

Garbanas:
Just improving upon the work of @georgiee with an optimized output.
Updated the JavaDoc of Body.addPhaserPolygon()


photonstorm:
Nice one, thanks!

PR #2123, opened 2015-10-06
Typescript definition updates.

clark-stevenson:
Fixed the plugin signature thanks to StackOverflow. Should cause no breaking changes.


photonstorm:
:+1:

PR #2125, opened 2015-10-06
Added updateLayout method definition

aweber1:
ScaleManager.updateLayout is present in phaser.js, but missing from the typescript definition. Not sure if that was intentional, but from what I read in the release notes ScaleManager.updateLayout is meant to replace ScaleManager.setScreenSize.


pnstickne:
It's a private method (not sure if that matters for the TS work).

photonstorm:
We don't include private methods in the TS defs.

PR #2128, opened 2015-10-08
Fixes for Cache::addBitmapFont(...)

nlotz:
- add missing API doc parameter "atlasType"
- apply default X/Y-Spacing when omitted as a parameter


photonstorm:
:+1:

PR #2156, opened 2015-10-23
Catchup on TypeScript Definitions bringing to Master 2.4.4

clark-stevenson:


photonstorm:
Thanks for this mate - I've now merged all of your changes into the current local file, so we should be in sync again.

PR #2186, opened 2015-10-31
Added `lifespan` argument to Particles.Arcade.Emitter.emitParticle.

danbolt:
This change enables a single particle to be emitted with a custom lifespan.

The change adds flexibility for users who may be using one emitter over a big sprite sheet.


danbolt:
I hadn't realized other commits from my fork have entered this pull request, so I'm going to close it for now. I only had planned on having bc2ea60 in there.

Thanks for taking the time to read this!

PR #2192, opened 2015-11-05
Enhanced the intro/outro usage.  The Pixidefaults inclusion have been…

spayton:
… decoupled from the Phaser Outro and put into a separate module, prior to this it would not be possible to exclude the UMD wrapper and keep the pixi defaults, which I'm not sure would have been much use to anyone, now you can optionaly keep the pixi defaults.

Also the Pixi intro module has been split into intro/main modules, so it's intro/outro modules can be excluded cleanly without being tied to the body of the Pixi code. Finally the Pixi into/outro exclusion is tied to Phaser's excludes, ie if Phasers intro/outro is excluded so are Pixi's - this could be separated by giving pixi it's own module list but I can't see why anyone would not want to exclude the wrappers from both, they can always generate the Phaser/Pixi modules in separate passes if they want differing wrapping anyway.


photonstorm:
:+1:

PR #2198, opened 2015-11-09
Force the usage of typescript 1.4.1

vulvulune:
Force the usage of typescript 1.4.1 because with higher versions the
generation of commented files failed. Refresh comments.d.ts files.


photonstorm:
Excellent, thanks.

PR #2209, opened 2015-11-17
Update BitmapData.setHSL to accept 0

FracturedShader:
Checking `if (h)` yields `false` when `h = 0`, but 0 is an acceptable value here. Had to decide between caching HSL verification and doing `typeof` 3x/pixel and opted for the temporary variables.


photonstorm:
Nice one, thanks.

PR #2210, opened 2015-11-17
Update BitmapData.shiftHSL, limitValue -> clamp

FracturedShader:
`math.limitValue` does not exist so it was changed to `math.clamp` which provides the desired functionality.


FracturedShader:
Request sent to master by mistake. This seems to happen to a lot of people.

PR #2211, opened 2015-11-18
Fix in typescript definition file: Pointer definition

juanmirod:
Pointer properties rightButton and LeftButton are DeviceButton objects instead of booleans (http://phaser.io/docs/2.4.4/Phaser.Pointer.html#rightButton)


danzel:
:+1:

According to the docs, these should be changed too:
```
backButton
eraserButton
forwardButton
middleButton
```

PR #2218, opened 2015-11-19
Allow direct import of Phaser.Image for RetroFont

milkey-mouse:
Before, when making a new RetroFont, you had to pass in a key that went to the Game.Cache. Now it works with either a string or a Phaser.Image.


milkey-mouse:
Crap, how do I delete a PR? I meant for the dev branch

PR #2219, opened 2015-11-19
Allow direct import of Phaser.Image for RetroFont

milkey-mouse:
Before, when making a new RetroFont, you had to pass in a key that went to the Game.Cache. Now it works with either a string or a Phaser.Image.

Now with 20% more the right branch!

### DIFF
--- a/src/gameobjects/RetroFont.js
+++ b/src/gameobjects/RetroFont.js
@@ -12,7 +12,7 @@
 * @extends Phaser.RenderTexture
 * @constructor
 * @param {Phaser.Game} game - Current game instance.
-* @param {string} key - The font set graphic set as stored in the Game.Cache.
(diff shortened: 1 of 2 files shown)


photonstorm:
I'm a bit dubious about this PR because (contrary to its name) a `Phaser.Image` can use all kinds of things as its texture, i.e. a RenderTexture, BitmapData, etc, which would break the RetroFont when it tries to render from it.

milkey-mouse:
Well, is there a better way to put images straight into a RetroFont? I needed this for a game I was making because there doesn't seem to be a better way to generate fonts at runtime without putting them into the cache.

photonstorm:
Can't you just put the image you need into the cache, then pass the key for it to the RetroFont?

photonstorm:
Sorry closing this off, just doesn't feel safe enough.

PR #2222, opened 2015-11-20
Update BitmapData.shiftHSL, limitValue -> clamp

FracturedShader:
`math.limitValue` does not exist so it was changed to `math.clamp` which provides the desired functionality.


photonstorm:
Nice one.

PR #2224, opened 2015-11-22
Adds Webpack documentation and custom build

deiga:


photonstorm:
I really like this idea, but I cannot merge it as it's against the wrong branch.

deiga:
Fix'd in #2331 2331

PR #2226, opened 2015-11-23
Change RetroFont.smoothed from string to boolean

milkey-mouse:
It's supposed to be a boolean as far as I can tell, and it works fine. If this is somehow wrong just close this PR.


photonstorm:
:+1:

PR #2235, opened 2015-11-30
Fixed AudioContext on single page apps.

stoneman1:
Added close to the audiocontext so it actually works in single page apps


photonstorm:
Closing duplicate PR.

Issue #2234, opened 2015-11-29
Phaser.Loader.  Errors in getAudioURL() and getVideoURL() using BLOB: or DATA:


jackfreak:
I'm trying to load sounds using data-uri scheme and I've found a small bug when using {uri: .., type: ..} pair.

In the preload() function of a state I add my audio like this:
```javascript
game.load.audio('Music', [{uri: 'data:audio/mp3;base64,SUQzAwAAAAAAFgAAAAAAAAAAAAAAAAAAAAAAAAAAAAD/+3BAAAAB/Rg.....', type: 'mp3'}] );
```

But when the audio gets loaded I get the error:
**Phaser.Loader - audio[Music]: error loading asset from URL false (No supported audio URL specified or device does not have audio playback support)**

I've tracked the error and is easily fixable, the problem is inside the getAudioURL() method of the Loader manager: When using url-type pairs, the url property gets overwritten before we can set the audioType property.

Currently the code looks like this: 
```javascript
            var url = urls[i];
            var audioType;

            if (url.uri) // {uri: .., type: ..} pair
            {
                url = url.uri;
                audioType = url.type;
            }
            else
```

And should be like this: 
```javascript
            var url = urls[i];
            var audioType;

            if (url.uri) // {uri: .., type: ..} pair
            {
                audioType = url.type;
                url = url.uri;
            }
            else
```

Also at the end of the for-loop, if we defined the audio with a {uri: .., type: ..} pair the code returns the url object that's inside the urls array instead of the url with the 'data:' 
```javascript
         if (this.game.device.canPlayAudio(audioType))
            {
                return urls[i];
            }
```

... so it breaks in the xhrLoad line of the loadFile() method when this.transfromUrl() gets executed because file.url is an Object instead of an url:
```javascript
this.xhrLoad(file, this.transformUrl(file.url, file), 'arraybuffer', this.fileComplete);
```

To fix this, getAudioURL() should be like:
```javascript
         if (this.game.device.canPlayAudio(audioType))
            {
                return url;
            }
```

We'll find these bugs in getVideoURL() too.

milkey-mouse:
I made a PR (#2236) to fix this, just making your changes in the Phaser.Loader source.

jackfreak:
Excellent, thanks!, I'm still new to all things related to git.

photonstorm:
Thanks both, fixed via PR merge.

lilijreey:
I user v2.5 but I same error 
[...]

PR #2236, opened 2015-11-30
Fix issue #2234: Errors in getAudioURL() and getVideoURL() when using blob: or data: URIs

milkey-mouse:
This PR contains the changes @jackfreak suggested for fixing getAudioURL() and getVideoURL() for blob: and data: URIs in issue #2234.


photonstorm:
:+1:

PR #2237, opened 2015-11-30
Fixed single page projects problem with too many AudioContexts

stoneman1:
There was no close() called on AudioContext so that lead to too many AudioContexts.


photonstorm:
I can't merge this because of all the other changes it encompasses, but have locally implemented a couple of them.

PR #2241, opened 2015-12-02
Fixed misleading error message

s4wny:
Shouldn't it be `tileset` instead of `key`?

We can't find any entry with the name `tileset` in the JSON file. The `key` hasn't anything to do with the JSON file, the `key` is our tilesset image we loaded with phaser.


photonstorm:
Yeah was probably a copy pasta error.

PR #2244, opened 2015-12-03
Rename inner loop variable from 'i' to 'j'

### LINKED ISSUE #2233: Multiple Tiled layers not loaded correctly due to loop variable reuse
Loop variable `i` is used for the outer loop:
https://github.com/photonstorm/phaser/blob/c9c85330ab60547b39ba4c9400c258ed7fb2a317/src/tilemap/TilemapParser.js#L227
and for the inner loop:
https://github.com/photonstorm/phaser/blob/c9c85330ab60547b39ba4c9400c258ed7fb2a317/src/tilemap/TilemapParser.js#L244
In the middle of a game jam right now, no time for a PR, but should be easy to fix :)

ttencate:
Fixes #2233

### DIFF
--- a/src/tilemap/TilemapParser.js
+++ b/src/tilemap/TilemapParser.js
@@ -242,11 +242,11 @@ Phaser.TilemapParser = {
                 var bytes = new Array( len );
                 // Interpret binaryString as an array of bytes representing
                 // little-endian encoded uint32 values. 
-                for (var i = 0; i < len; i+=4) {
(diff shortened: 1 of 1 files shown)


Offlein:
Not sure the workflow here, but I can attest that this works and fixes a critical bug in Phaser.

bsparks:
:+1: for merging this, I was just about to submit a PR myself until I saw this one

photonstorm:
:+1:

PR #2248, opened 2015-12-08
Fixed ScaleManager EXACT_FIT not working

06wj:
```
var game = new Phaser.Game({
    width:100,
    height:100,
    parent:'gameContainer',
    scaleMode:Phaser.ScaleManager.EXACT_FIT
});
```
this code not working bescause Phaser.ScaleManager.EXACT_FIT is zero.  

```
if (config['scaleMode'] !== undefined)//config['scaleMode'] default value is undefined

if (this._pendingScaleMode !== null) //this._pendingScaleMode default value is null
```
I fixed it by this code. 😄

### DIFF
--- a/src/core/ScaleManager.js
+++ b/src/core/ScaleManager.js
@@ -725,7 +725,7 @@ Phaser.ScaleManager.prototype = {
 
         this._booted = true;
 
-        if (this._pendingScaleMode)
+        if (this._pendingScaleMode !== null)
         {
             this.scaleMode = this._pendingScaleMode;
             this._pendingScaleMode = null;
@@ -742,7 +742,7 @@ Phaser.ScaleManager.prototype = {
     */
     parseConfig: function (config) {
 
-        if (config['scaleMode'])
+        if (config['scaleMode'] !== undefined)
         {
             if (this._booted)
             {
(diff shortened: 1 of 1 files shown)


src/core/ScaleManager.js
```diff
@@ -725,7 +725,7 @@ Phaser.ScaleManager.prototype = {
 
         this._booted = true;
 
-        if (this._pendingScaleMode)
+        if (this._pendingScaleMode !== undefined)
```
GoToLoop: Should consider instead `!= undefined` or `!= null`, so it also catches `null` too.

06wj:
```
this._pendingScaleMode !== null
```

I changed it to null because this._pendingScaleMode default value is null

GoToLoop:
@06wj, you didn't get it. Don't use `===` nor `!==` when checking for `undefined` or `null`!
Instead use `==` and `!=` b/c they catch both `undefined` & `null` in 1 checking only:  
`if (this._pendingScaleMode != undefined) {} // catches both undefined & null.`

06wj:
@GoToLoop   
this._pendingScaleMode  only set in this code, cannot be undefined.
```
if (config['scaleMode'] !== undefined)
        {
            if (this._booted)
            {
                this.scaleMode = config['scaleMode'];
            }
            else
            {
                this._pendingScaleMode = config['scaleMode'];
            }
        }
```

GoToLoop:
* It's not a question whether _this._pendingScaleMode_ cannot be `undefined` or not.
* Checking against `null` or `undefined` via `==` or `!=` is guaranteed to catch both cases at all times.
* 99.9% of times when checking against `undefined`, it was also meant to check against `null` and vice-versa!

Issue #2250, opened 2015-12-09
Loader.binary returns success=true even on a 404 error


mhstar89:
I'm using a loader instance to download a file:

            var x = new Phaser.Loader(game);
            x.onFileComplete.add(load_complete);
            x.binary('key', 'does_not_exist');
            x.start();

load_complete is called, but success is set to true (even though my browser - Chrome 47.0.2526.73 m logged the error in the console).

Looking into the phaser code, in Loader.js, xhr.onload is called, with xhr.readyState = 4 and xhr.status = 404. xhr.onerror is not called, and the error is never detected by Phaser.

This doesn't seem to be uncommon - looking at http://stackoverflow.com/questions/23986114/ajax-onerror-handler-not-working for example, they state that errors are considered responses and for this reason onerror isn't called. 

I suggest changing the function 

        xhr.onload = function () {

            try {

                return onload.call(_this, file, xhr);

            } catch (e) {

                //  If this was the last file in the queue and an error is thrown in the create method
                //  then it's caught here, so be sure we don't carry on processing it

                if (!_this.hasLoaded)
                {
                    _this.asyncComplete(file, e.message || 'Exception');
                }
                else
                {
                    if (window['console'])
                    {
                        console.error(e);
                    }
                }
            }
        };

to 

        xhr.onload = function () {
            try {
                if (xhr.readyState == 4 && xhr.status >= 400 && xhr.status <= 599) { // Handle HTTP status codes of 4xx and 5xx as errors, even if xhr.onerror was not called.
                    return onerror.call(_this, file, xhr);
                }
                else {
                    return onload.call(_this, file, xhr);
                }

            } catch (e) {
                //  If this was the last file in the queue and an error is thrown in the create method
                //  then it's caught here, so be sure we don't carry on processing it

                if (!_this.hasLoaded)
                {
                    _this.asyncComplete(file, e.message || 'Exception');
                }
                else
                {
                    if (window['console'])
                    {
                        console.error(e);
                    }
                }
            }
        };

which will cause the error to be handled appropriately (Phaser then logs the 404 status code, too).

milkey-mouse:
After testing and making sure they work, I made a PR with your changes.

photonstorm:
Fixed via PR merge.

PR #2251, opened 2015-12-10
Fix issue #2250: Loader.binary returns success on 404

milkey-mouse:
@mhstar89 put a proposed fix in his issue (#2250) which seemed to work, so here's a PR for it.


photonstorm:
Yes, good catch both.

Issue #2188, opened 2015-11-01
BitmapData.copy blendMode is wrong type on Typescript definition file


brinca:
The parameter type for the blendMode argument of the copy(), copyRect(), draw() and drawGroup() methods of BitmapData are set to type "number" on the Typescript definition file, while the expected (and working) values are of type "string".

This is probably because it used to expect values from the PIXI.blendModes enumeration (which is number based), and now is using HTML5's globalCompositeOperation (a string value), but the type remains "number" in the definition file (the documentation states the type as "string" though).

It is easily fixed by updating the definition file (phaser.d.ts or phaser.comments.d.ts), replacing all instances where "blendMode" is "number" to either "string" or "any".

milkey-mouse:
I just made a PR to fix this, thanks

photonstorm:
Fixed via PR merge.

PR #2252, opened 2015-12-11
Fix issue #2188: BitmapData.copy() blendMode type from number to string in TS defs

milkey-mouse:
This fixes a minor TS definitions bug (issue #2188) where the blendMode argument of BitmapData.copy(), .copyRect(), .draw(), and .drawGroup() all had the number type instead of string.


photonstorm:
:+1:

PR #2258, opened 2015-12-17
Updated Emmiter Definition makeParticles() and particleClass

### LINKED ISSUE #2257: Typescript Emitter makeParticles definition
Hi,

The currently version of the definition 2.4.4 in the repo of phaser.d.ts don't allow to call Phaser.Particles.Arcade.Emitter.makeParticles() without args but in the examples this works.

http://phaser.io/examples/v2/particles/particle-class

Also the particleClass property has a type of Phaser.Sprite when it should be any.

Thanks

iamfreee:
fixes #2257


photonstorm:
Thanks. I can't merge because wrong branch, but have added the changes locally.

PR #2263, opened 2015-12-18
Camera ignores scaling when checking for the world bounds

ForGorNorPor:
I added the scaling factor to the checkBounds() method of the camera. I assume that this is how the camera should behave, shouldn't it?


photonstorm:
Thanks. I can't merge your PR because it's on the wrong branch, but have made the changes locally.

PR #2267, opened 2015-12-22
Made Phaser.Text create a new object for this.style

asyncanup:
... instead of using the same reference for style options object.
This helps when user is using options hierarchies for instantiating game objects.


photonstorm:
I can't merge because it's the wrong branch, but it's a good idea so have merged locally.

PR #2277, opened 2016-01-03
Alow precalculating the results of text wrapping. 

slashman:
This may be useful for example if the developer wants to control pagination, in which case he may know in advance the total number of lines the text will wrap to, and then set it on batches.


slashman:
Fixing a potential issue.

photonstorm:
:+1:

PR #2281, opened 2016-01-04
Fixed bug in parseTiledJSON method. 

aweber1:
Fixed bug in parseTiledJSON method. When iterating the json.layers array, if the layer data being processed is base64-encoded, the loop for decoding the base64 data used the same "i" variable declared in the outer loop scope, thereby only allowing one layer to be processed from the json.layers array.


photonstorm:
Thanks, fixed via #2287

Issue #2213, opened 2015-11-18
PIXI pushMask: Cannot read property 'data' of undefined


geros:
Phaser  2.4.4
I apply a rectangle mask out of graphic and i get this error when i use WEBGL  :

 ```javascript
Uncaught TypeError: Cannot read property 'data' of undefined
```
At 
```javascript
/**
* Applies the Mask and adds it to the current filter stack.
* 
* @method pushMask
* @param maskData {Array}
* @param renderSession {Object}
*/
PIXI.WebGLMaskManager.prototype.pushMask = function(maskData, renderSession)
{
    var gl = renderSession.gl;

    if(maskData.dirty)
    {
        PIXI.WebGLGraphics.updateGraphics(maskData, gl);
    }

    if(!maskData._webGL[gl.id].data.length)return;

    renderSession.stencilManager.pushStencil(maskData, maskData._webGL[gl.id].data[0], renderSession);
};
```

stoneman1:
Have the same problem. Here is some data : 

![screen shot 2015-12-01 at 00 00 19](https://cloud.githubusercontent.com/assets/4244526/11487705/d3a5dedc-97c9-11e5-9730-f3ff79955030.png)
![screen shot 2015-12-01 at 00 04 59](https://cloud.githubusercontent.com/assets/4244526/11487712/e101cd2a-97c9-11e5-87c8-b29228438cb9.png)
![screen shot 2015-12-01 at 00 05 52](https://cloud.githubusercontent.com/assets/4244526/11487720/ed23d2b0-97c9-11e5-883e-97da38718d1e.png)
![screen shot 2015-12-01 at 00 02 26](https://cloud.githubusercontent.com/assets/4244526/11487747/2a0e001a-97ca-11e5-859f-de5131b5ecd3.png)

geros:
I just made a terrible and unexplainable hack that made my code working. 
Similar situation as the one i opened the issue for same error.
So 
I had to mask a group of sprites so i gave :
```
var mask = this.game.add.graphics(0, 0);
mask.beginFill(0x000);
mask.drawPolygon(points);
mask.endFill();

this.add(mask); 
this.diceGroup.mask = mask;
```
The above resulted to the same error as initially reported when render is WEBGL (it works on CANVAS). 
When i give:
```
var mask = this.game.add.graphics(0, 0);
mask.beginFill(0x000);
mask.drawPolygon(points);
mask.endFill();
this.add(mask);

[...]

Issue #2152, opened 2015-10-22
[2.4.4] cacheAsBitmap breaks graphics as mask


jmp909:
using a cached graphic as a mask used to work but not as of 2.4.4

``` javascript
var gMask = game.add.graphics(0, 0);
gMask.lineStyle(2, 0xFF0000, 1);
gMask.beginFill(0xFF0000,0.1)
gMask.drawCircle(0,0,mapSize/2,mapSize/2);    
gMask.endFill()

gMask.cacheAsBitmap=true // !!!! breaks mask

renderGroup.add(gMask)
renderSprite.mask=gMask;
```

error
`Uncaught TypeError: Cannot read property 'data' of undefined`
(phaser.js 17578)

seems to be this line
`if(!maskData._webGL[gl.id].data.length)return;`

geros:
Have you found any workaround on this? 
It happens to me when i use  renderer:Phaser.WEBGL while works fine  when i change it to  Phaser.CANVAS

jmp909:
quick fix for WebGL Graphics is to manually set the mask as dirty

```javascript
mask.cacheAsBitmap=true;
mask.dirty=true
```

but ideally would presumably for this to happen in the PIXI source in this function
`Object.defineProperty(PIXI.DisplayObject.prototype, 'mask', {`

https://github.com/photonstorm/phaser/blob/c9c85330ab60547b39ba4c9400c258ed7fb2a317/src/pixi/display/DisplayObject.js#L302

photonstorm:
Thank you for taking the time to open this issue. However, official support for Phaser 2 has now ended as we focus on preparing Phaser 3 for beta release. In a bid to keep the project tidy, and avoid confusion, we are closing all v2 related issues.

This does not mean your issue won't be looked-at. In November 2016 the [Phaser CE (Community Edition)](https://github.com/photonstorm/phaser-ce) project began. The aim of Phaser CE was to allow the community to continue v2 development. Since then it has gained significant traction, with many updates and releases.

If you are still interested in having your issue investigated we strongly recommend testing the latest Phaser CE release. And only if the problem persists, opening an issue on the Phaser CE repo.

PR #2282, opened 2016-01-05
Fixes an issue with changing the width/height of a mask. (WEBGL)

nexiuhm:
###### Jsfiddle for context:  https://jsfiddle.net/wdghvgkh/4/

If you change the width or height of a graphics mask after its already been assigned to another displayObject PIXI will crash with the following error:  `Uncaught TypeError: Cannot read property 'data' of undefined`. in `PIXI.WebGLMaskManager.prototype.pushMask()`. 

.. Many breakpoints later  - it's apperent that this is caused by the mask mistakenly(?) beeing set to ``not dirty`` by  `PIXI.Graphics.prototype.getBounds()`.

The PR should also solve this issue: https://github.com/photonstorm/phaser/issues/2213

### DIFF
--- a/src/pixi/display/DisplayObject.js
+++ b/src/pixi/display/DisplayObject.js
@@ -295,11 +295,19 @@ Object.defineProperty(PIXI.DisplayObject.prototype, 'mask', {
 
     set: function(value) {
 
-        if (this._mask) this._mask.isMask = false;
+        if (this._mask)
+        {
+            this._mask.isMask = false;
+        }
 
         this._mask = value;
 
-        if (this._mask) this._mask.isMask = true;
+        if (this._mask)
+        {
+            this._mask.isMask = true;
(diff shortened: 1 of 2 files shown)


nexiuhm:
Since its kind of related I also fixed the issue with setting cacheAsBitmap = true on a mask. Issue is posted here: https://github.com/photonstorm/phaser/issues/2152

nexiuhm:
I should add that this was kind of a surgical fix to the problem, I can imagine the rabbit hole is much deeper and might require some rewriting of Pixi to fix the underlying issue.

nexiuhm:
Closing this since I deleted the repo and can't make changes to it. I'll make a new PR if it turns out well :)

nexiuhm:
Closed the wrong PR , oh well :p

PR #2285, opened 2016-01-06
Fix cache reference bug in web GL context restoration.

allenevans:
Fixes error being reported in the console when the web GL context is restored in the game. 

phaser-cleargltextures-bug-console

PR fixes cache reference typo.

### DIFF
--- a/src/loader/Cache.js
+++ b/src/loader/Cache.js
@@ -1943,9 +1943,9 @@ Phaser.Cache.prototype = {
     */
     clearGLTextures: function () {
 
-        for (var key in this.cache.image)
+        for (var key in this._cache.image)
         {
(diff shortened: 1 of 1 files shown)


mazhlekov:
@i-dimitrov We are not alone. :dancers:

allenevans:
This PR only fixes the code defect. With this change in place, I now get these warnings in the console.

image

I've been able to reliably replicate this using Chrome 47 with the game running on an external monitor that when I unplug and plug back in throws the error.

photonstorm:
I'll merge this in, but I fully suspect that there are lots of other places the context loss needs to be captured (and even after doing that I'm not 100% sure how well Pixi dealt with the restore anyway, i.e. this may be a much deeper rabbit hole than it appears)

PR #2287, opened 2016-01-07
Support multiple layers in Tiled JSON

Saluev:
Awkward error with namespaces was not allowing to load more than one layer from Tiled JSON file. Fixed that.


Saluev:
Ooops. Three days ago someone already proposed that. #2281

PR #2298, opened 2016-01-15
Update FrameData.js

Weedshaker:
"Uncaught TypeError: Cannot read property 'index' of undefined"

The above error occurred, after updating https://github.com/englercj/phaser-tiled to be compatible with Phaser version 2.4.4. The above change fixed this issue!


photonstorm:
Not entirely sure of the reason for this change, but it's just a bit of extra hardening and won't break anything so I'll merge it in.

PR #2300, opened 2016-01-15
Typescript definitions - fixed return types on some functions of Arcade.Body

zimpy:


photonstorm:
Thanks. In future you don't need to update the 'comments' version of the defs, as it's auto-generated.

PR #2301, opened 2016-01-17
Stop only the named animation if `name` is passed

### LINKED ISSUE #2299: AnimationManager#stop(name) does not work as described
> If a name is given that specific animation is stopped […]

In fact if you pass the name of any existing animation, the function [stops the current animation (no matter which it is) and then loads the named animation](https://github.com/photonstorm/phaser/blob/db641ca82e608470bd7902de3447d048d9715ab5/src/animation/AnimationManager.js#L311).

samme:
Fixes #2299


photonstorm:
Yes, good catch.

PR #2315, opened 2016-02-02
Typescript defintions: RandomDataGenerator.sow parameter type should be any[]

staff0rd:
Currently set to `number[]`, but `any[]` is allowed.


staff0rd:
May conflict with #2315, I will fix if so.

photonstorm:
Yeah conflicts, have merged your other PR.

PR #2316, opened 2016-02-02
Typescript - RandomDataGenerator.sow should accept any[]

staff0rd:
Currently set to number[], but any[] is allowed.

### DIFF
--- a/typescript/phaser.comments.d.ts
+++ b/typescript/phaser.comments.d.ts
@@ -20434,7 +20434,7 @@ declare module Phaser {
         * 
         * @param seeds An array of values to use as the seed, or a generator state (from {#state}).
         */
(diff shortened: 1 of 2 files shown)


photonstorm:
This is already fixed in the dev branch (although only as of a few minutes ago!)

staff0rd:
@photonstorm how so?  Both [phaser.d.ts](https://github.com/photonstorm/phaser/blob/dev/typescript/phaser.d.ts) and [phaser.comments.d.ts](https://github.com/photonstorm/phaser/blob/dev/typescript/phaser.comments.d.ts) in dev still accept only `number[]` but [RandomData.js](https://github.com/photonstorm/phaser/blob/dev/src/math/RandomDataGenerator.js) accepts `any[]`.

Issue #2232, opened 2015-11-27
Group hash growth unbounded?


strawlion:
Hey All, I noticed that when moving a sprite from one group to another via the add call, it's not removed from the original group's hash. Is this intended?

IE (pseudo code)
```javascript
var sprite = someSprite;

groupOne.add(sprite);
// groupOne.hash === [sprite]
// groupTwo.hash === []

groupTwo.add(sprite);
// groupOne.hash === [sprite]
// groupTwo.hash === [sprite]
```

This is with V2.4.3

strawlion:
FWIW, I'd be happy to submit a fix along with tests here if it's a real issue

McIntozh:
Yes, I can confirm this for 2.4.4 too.
Group hash grows and hold references to already destroyed sprites, causing memory leaks.

But this only happens when the sprite has a physics body:

```
var group1;
var group2;
function create() {
    game.physics.startSystem(Phaser.Physics.P2JS);
    group1 = game.add.group();
    group2 = game.add.group();

    var sprite=game.add.sprite(0,0);
    game.physics.p2.enable(sprite);
    console.log("Should return false: "+(group1.hash.indexOf(sprite)>=0));
    
    group1.add(sprite);
    console.log("Should return true: "+(group1.hash.indexOf(sprite)>=0));
    
    group2.add(sprite);
    console.log("Should return false: "+(group1.hash.indexOf(sprite)>=0));
}
```

Workarounds:
1) Remove the child from the original group before adding it to another group
2) Call `if (this.parent && this.parent.remove)this.parent.remove(this);` before adding to a group

Suggested Bugfix:
The reason for this bug is, that `add` calls `this.addChild` (Line 31402 in phaser.js) wich removes the sprite from the children of its parent sprite. But because this is in PIXI code, `removeFromHash` is not called. Perhaps this should be called before `addChild`.

photonstorm:
This is now fixed in dev.

PR #2323, opened 2016-02-05
In groups, check for child.parent before removeFromHash()

spayton:
Particles are created with arcade physics bodies but have no parent when added to the particle group, so it isn't sufficient to just check child.body we must also check child.parent before calling child.parent.removeFromHash(), otheriwise it crashes.


photonstorm:
Why would a particle ever be in another Group in order to need removing from its previous hash?

spayton:
Calls to child.parent.removeFromHash() were added in response to issue #2232, I observed that this code could be called when child.parent was undefined causing a crash.

In my case the child was added to the group in Phaser.Particles.Arcade.Emitter.prototype.makeParticles().  At the bottom of this function the new particle being added doesn't have a parent but it does have a body, which are the requirements to subsequently crash.

So, I'm not saying there's particles in other groups being removed from hash, or that there's any problem in particular with the hash - I'm saying it's possible to hit the new code with child.parent == undefined.

spayton:
Just checked again in case it was just my modified phaser build.  On stock dev phaser the problem shows itself on the 1st particles example.

photonstorm:
This is merged locally into dev.

PR #2326, opened 2016-02-06
Fix #2320: properly size filter texture and viewport

### LINKED ISSUE #2320: Custom Resolution and Filter Bugs
Hello, I chatted a bit with @photonstorm over slack about these two specific issues and am following up with an issue to easily recreate them.

I have made a [demo page](http://wtf.amadeus.wtf/phaser-filter-bug-demo/#4) that can be used to see these two bugs in action. This demo page has some interactive dropdowns to change the in game resolution and toggle various settings that I will refer to later in this issue. I have also created a gist of my [state file](https://gist.github.com/amadeus/b9e0ed4ccf23dd88917b) that can be easily referred too.

Also, I believe both of these issues are related to the version of PIXI you manage internally for Phaser 2.x.

## Bug 1:

Filters are rendered differently depending on whether the filter is the last in the filters array or not.

If you check the [demo page](http://wtf.amadeus.wtf/phaser-filter-bug-demo/#4) and look closely, I have am applying a [noise filter to a background image](https://gist.github.com/amadeus/b9e0ed4ccf23dd88917b#file-menu-js-L89). This scene (using the link above) that is being rendered at a resolution of 4. The noise effect however looks a bit blurry (you may have to zoom in or screenshot and zoom in to see the blurriness). If you click the `Add Passthrough` button, I [add another filter](https://gist.github.com/amadeus/b9e0ed4ccf23dd88917b#file-menu-js-L58) to the background image and you will see that now the noise is nice and sharp and pixellated. Just for note, this second filter I apply actually does nothing, it just returns the colors of the scene exactly as they are ([passthrough filter](https://gist.github.com/amadeus/00e3da591a3bbf32d11e)).

It appears as if the all filters before the last take resolution into account properly and don't scale up. I've tested this with a variety of different filters all applied at the same time, and the only blurry one is ALWAYS the last one. I believe it has something to do with the [pushFilter method](https://github.com/photonstorm/phaser/blob/db641ca82e608470bd7902de3447d048d9715ab5/src/pixi/renderers/webgl/utils/WebGLFilterManager.js#L152) since a different path is taken for all but the [last filter](https://github.com/photonstorm/phaser/blob/db641ca82e608470bd7902de3447d048d9715ab5/src/pixi/renderers/webgl/utils/WebGLFilterManager.js#L192)

Here is a blown up image of the noise with no passthrough (as the page loads):
no passthrough

And here is a blown up image of the noise with the pass through filter appended to the filters array (after pressing the `Add Passthrough` button):
with passthrough

Bear in mind, this issue ONLY exists if the resolution is greater than 1. Of you set the resolution to 1 nothing changes based on whether the passthrough filter has been added or  not.

## Bug 2:

By [default](https://gist.github.com/amadeus/b9e0ed4ccf23dd88917b#file-menu-js-L75), the scene has 2 images added to a group. One of those images has a noise filter.

If you now add a filter to the entire stage, using the `Add Scene Filter` button ([code here](https://gist.github.com/amadeus/b9e0ed4ccf23dd88917b#file-menu-js-L63)), AND there is a custom resolution greater than 1, every child element gets scaled up relative to the resolution. You can test this by trying out different resolution combinations coupled with the `Add/Remove Scene Filter` button.

And for reference, here is what the scene looks like at a resolution of 2, with no stage filter (the correct view):
no stage filter

And here's the same scene, with a filter added to the stage at a resolution of 2:
with stage filter

And finally, here is what it SHOULD look like, with the stage filter, however this is only possible at a resolution of 1:
with stage filter, at 1x

I've also done testing with putting the filters on the group instead of the stage, and the same results occur. I think the bug occurs when I have filters on nested elements. If I add filters to sibling elements, nothing seems to get messed up.

englercj:
This patch is to fix the bugs brought up in #2320. It does so by properly sizing both the filter texture and the gl viewport.

### DIFF
--- a/src/pixi/renderers/webgl/utils/WebGLFilterManager.js
+++ b/src/pixi/renderers/webgl/utils/WebGLFilterManager.js
@@ -93,11 +93,11 @@ PIXI.WebGLFilterManager.prototype.pushFilter = function(filterBlock)
     var texture = this.texturePool.pop();
     if(!texture)
     {
-        texture = new PIXI.FilterTexture(this.gl, this.width, this.height);
+        texture = new PIXI.FilterTexture(this.gl, this.width * this.renderSession.resolution, this.height * this.renderSession.resolution);
     }
     else
     {
-        texture.resize(this.width, this.height);
+        texture.resize(this.width * this.renderSession.resolution, this.height * this.renderSession.resolution);
     }
 
     gl.bindTexture(gl.TEXTURE_2D,  texture.texture);
@@ -120,7 +120,7 @@ PIXI.WebGLFilterManager.prototype.pushFilter = function(filterBlock)
     gl.bindFramebuffer(gl.FRAMEBUFFER, texture.frameBuffer);
 
     // set view port
-    gl.viewport(0, 0, filterArea.width, filterArea.height);
+    gl.viewport(0, 0, filterArea.width * this.renderSession.resolution, filterArea.height * this.renderSession.resolution);
 
     projection.x = filterArea.width/2;
     projection.y = -filterArea.height/2;
--- a/src/pixi/renderers/webgl/utils/WebGLFilterManager.js
+++ b/src/pixi/renderers/webgl/utils/WebGLFilterManager.js
@@ -158,7 +158,7 @@ PIXI.WebGLFilterManager.prototype.popFilter = function()
 
     if(filterBlock.filterPasses.length > 1)
     {
-        gl.viewport(0, 0, filterArea.width, filterArea.height);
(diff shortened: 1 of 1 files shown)


amadeus:
This is awesome and definitely getting closer. One issue I have found however, is that it ends up breaking down if you add more than 2 filters in a very interesting way. I have update the demo code to reflect this - if you click the `Add Scene Filter` button, it creates a super tiny, filtered image super imposed on top of the unfiltered image. - http://wtf.amadeus.wtf/phaser-filter-bug-demo/#4

englercj:
Should be good now, I had missed the output texture location :P

PR #2328, opened 2016-02-07
Added RandomDataGenerator.sign (returns -1 or +1)

taylankasap:
I just started a new project and I needed this method 3 times already (random angularVelocity & velocity direction). Am I missing something?

I'm not sure about the method name though. One might expect it to return string '-' or '+'. Any recommendations?


photonstorm:
I think the name is fine, merging in.

PR #2331, opened 2016-02-09
Adds Webpack documentation and custom build

deiga:


photonstorm:
Merged, but in future please don't commit build files.

zerojuan:
The instructions in the docs wasn't working for me.

P2 wasn't loading correctly. I ended up loading it via script:

`{
        test: /(pixi.js|p2)/,
        loader: 'script'
      }`

PR #2349, opened 2016-02-24
Fixed issue with IE crashing on this.context.close() in SoundManager.…

stoneman1:
…js (There was no check if context exists)


photonstorm:
:+1:

PR #2353, opened 2016-02-25
Fix for world.centerX and world.centerY with negative bounds origin.

fillmoreb:
Phaser.World.centerX and Phaser.World.centerY only worked if the bounds had an origin of 0,0.  This change takes into account the actual origin.


photonstorm:
:+1:

PR #2357, opened 2016-02-26
Patch for destructive getLocalBounds in PIXI.DisplayObjectContainer

st0nerhat:
PIXI has an open upstream bug regarding getLocalBounds. See https://github.com/pixijs/pixi.js/issues/1698.

getLocalBounds destroys the worldTransforms on children until the next stage.updateTransform() call in Phaser.Game. This can make a number of things break including mouse input if width,height, or getLocalBounds methods are called inside of an update or preUpdate method.

Because Phaser uses a modified version of PIXI, I'm proposing to directly patch the issue here.


photonstorm:
Agreed, makes a lot of sense. Merged.

Issue #2325, opened 2016-02-06
No global volume setting


VitaZheltyakov:
If the browser does not use the Audio API (eg, Crosswalk or CocoonJS), there is not global volume setting.

Example:
If I have mySound with volume = 1, and want change global volume to 0.9, I may do:
game.sound.volume = 0.9
it set mySound.volume = 1*0.9 = 0.9

Then I want change global volume to 0.8:
game.sound.volume = 0.8
it set mySound.volume = 0.9*0.8 = 0.72 

Then I want return global volume to 0.9:
game.sound.volume = 0.9
it set mySound.volume = 0.72*0.9 = 0.648

If browser does use the Audio API global volume setting work сorrectly

photonstorm:
Thanks for submitting this issue. We have fixed this, and the fix has been pushed to the `dev` branch.

PR #2369, opened 2016-03-02
Fixes issue #2325

stoneman1:
Should fix that issue.

Seems like github wanted to put my MediaStream fix in this pull too..


photonstorm:
This is not backward compatible, both cases need to be covered (if getTracks exists, use that approach, otherwise use stop)

PR #2371, opened 2016-03-02
Fixed issue #2325 and fixed video for future

stoneman1:
Fixed issue #2325.
Checking if getTracks exists, then use MediaStreamTrack.stop() instead of MediaStream.stop().

### DIFF
--- a/src/sound/SoundManager.js
+++ b/src/sound/SoundManager.js
@@ -830,7 +830,7 @@ Object.defineProperty(Phaser.SoundManager.prototype, "volume", {
                 {
                     if (this._sounds[i].usingAudioTag)
                     {
-                        this._sounds[i].volume = this._sounds[i].volume * value;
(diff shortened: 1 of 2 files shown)


src/sound/SoundManager.js
```diff
@@ -830,7 +830,7 @@ Object.defineProperty(Phaser.SoundManager.prototype, "volume", {
                 {
                     if (this._sounds[i].usingAudioTag)
                     {
-                        this._sounds[i].volume = this._sounds[i].volume * value;
+                        this._sounds[i].volume = value;
```
photonstorm: While I agree what we have currently doesn't work, this won't fix it either. All it will do is normalise all of the sounds to have the exact same volume, which isn't the end result we want.

photonstorm:
Merged as I'm going to fix the sound issue locally now, but want the video fix in.

PR #2375, opened 2016-03-04
Fix tinting issue when loading textures

CptSelewin:
Previously when a sprite was tinted and a new texture was loaded then the tint did not apply to the texture and the old tinted texture was used. This change fixes the issue by resetting the cached tint upon texture loading.

Here is an example that failed:
```js
var mysprite = game.add.sprite(1, 2, 'texturea');
mysprite.tint = 0xff00ff;
```

Later:
```js
mysprite.loadTexture('textureb');
```
Now the tint was not applied to the new texture and the old tinted texture was used for rendering.


CptSelewin:
See PR #2383

PR #2377, opened 2016-03-05
Remove dangling 'mouseout' listener from Mouse.js

marcus-robinson:
Phaser currently throws run time errors after you destroy a game because the 'mouseout' event handler will continue to fire. This resolves that issue.


photonstorm:
This is already fixed in dev.

Issue #2374, opened 2016-03-03
Text cropped when using negative lineSpacing


gaelenh:
In 2.4.6, a change was introduced to the Text spacing code that reduced the height of the drawn text.  https://github.com/photonstorm/phaser/commit/0c52b9316a774aa88d08b3e824934f8d5da0563f.  This change causes text to be cropped at the bottom for larger lineSpacing values.  Given that this is a regression to how this was previously displayed, I am submitting it as a bug.

Example image: 
![screen shot 2016-03-03 at 6 08 00 pm](https://cloud.githubusercontent.com/assets/325937/13512852/43783efe-e16b-11e5-9d59-afa492a26488.png)

Code to reproduce: [cropped-text.zip](https://github.com/photonstorm/phaser/files/157837/cropped-text.zip)

gaelenh:
I created pull request https://github.com/photonstorm/phaser/pull/2379 to address this.

photonstorm:
Merged, thanks!

PR #2379, opened 2016-03-07
Negative lineSpacing shouldn't crop text

gaelenh:
2.4.6 introduced a change to the way line height is calculated. This caused negative lineSpacing to have cropped text. I added a conditional to handle negative line spacing while still honoring the new calculation for positive lineSpacing. I reported the issue as https://github.com/photonstorm/phaser/issues/2374.

Example of bug:
43783efe-e16b-11e5-9d59-afa492a26488

Example code: 
[cropped-text.zip](https://github.com/photonstorm/phaser/files/161894/cropped-text.zip)

### DIFF
--- a/src/gameobjects/Text.js
+++ b/src/gameobjects/Text.js
@@ -439,7 +439,8 @@ Phaser.Text.prototype.updateText = function () {
     //  Adjust for line spacing
     if (lineSpacing !== 0)
     {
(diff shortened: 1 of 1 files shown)


photonstorm:
I couldn't merge your PR because of conflicts, but have added your change locally, thanks.

PR #2394, opened 2016-03-22
Check for sprite destruction in onInputDown event.

zeterain:
This sprite might have been destroyed during the onInputDown event. Check to see if it was.

Also, set the pointer's dirty flag before the altered if-block, just in case the function returns.


photonstorm:
Sure, makes sense.

PR #2410, opened 2016-04-03
Add "maxLines" style attribute to Text object.

slashman:
Maximum number of lines to be shown for wrapped text or 0 for no limit. (default).

This prevents wrapped text from overflowing on a fixed layout.


slashman:
Found out an issue. (!)

photonstorm:
Nice.

PR #2412, opened 2016-04-03
Remove encoding metadata from Tiled json if data has already had its …

MannyC:
The parser updates the json passed into it to remove the base64 encoding.  I'm unsure if this is a good idea or not, but given that that's what it's already doing it should remove the encoding parameter so that a subsequent process doesn't try to decode the data again.  

If preferred I could rewrite it to use a local variable and leave the passed in layer data as it is, but of course this has more potential to conflict with existing code.  

The new line 252 is the only substantive change, everything else is from the whitespace removal due to the .editorconfig

### DIFF
--- a/src/tilemap/TilemapParser.js
+++ b/src/tilemap/TilemapParser.js
@@ -13,7 +13,7 @@
 Phaser.TilemapParser = {
 
     /**
-     * When scanning the Tiled map data the TilemapParser can either insert a null value (true) or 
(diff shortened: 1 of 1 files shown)


sauthrully:
Hello!

 

You have a new message, please read <http://legacy.top100projects.ca/front.php?qwy>

 

malaikat_benar@rocketmail.com

PR #2413, opened 2016-04-03
Detect if layer compression is used in a Tiled layer.  Display warnin…

MannyC:
…g and skip layer.

The TiledParser only supports uncompressed layer data.  Previously it would just silently fail if the layer was compressed.  This adds a warning message and skips the layer.  Note that skipping the layer is new functionality, previously it would create the layer but without any tile data.  This would essentially mean the layer is unusable.  However, I can remove the `continue` if preferred.

Other than new lines 253-256 everything is whitespace removal.

### DIFF
--- a/src/tilemap/TilemapParser.js
+++ b/src/tilemap/TilemapParser.js
@@ -13,7 +13,7 @@
 Phaser.TilemapParser = {
 
     /**
-     * When scanning the Tiled map data the TilemapParser can either insert a null value (true) or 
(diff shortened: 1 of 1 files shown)


sauthrully:
Hello,

Have you seen something as lovely as that stuff? I swear you haven't, just take a look here <http://shyshenketu.asiandice.com/lnyqakm>

malaikat_benar@rocketmail.com

PR #2415, opened 2016-04-04
Fixed and added warnings.

jakewilson:
This is very useful if you have multiple Tileset images and don't know which one is giving the warning.

The second commit will save an hour-long headache if someone accidentally applies `Phaser.Physics.Arcade` to a group instead of `Phaser.Physics.ARCADE` and can't understand why their `sprite.body` property is `null` (which just happened to me).


photonstorm:
Nice one.

PR #2420, opened 2016-04-07
Fixes export statement in .d.ts to reflect new syntax 1.8..

janus007:
So we can use them in Visual Studio and get intellisense and no namespace errors.


janus007:
Not even sure why we need d.ts's without comments? 
During transpiling we can configure it to remove the comments :) it would be easier to maintain.

clark-stevenson:
I have been shifted to dom based projects for a good long time and have fallen out of touch somewhat with this side of the net. 

The definition files were manually created and at the time, the rate of change of pace was pretty rapid. Sometimes it would take hours a week to maintain the commit history.  Had comments been involved, then it would have been impossible to compare 2 files easily without scrolling through 10 times as many lines. 

A cool dude came along and merged docs into the `phaser.d.ts` file to give us the comments. I only ever maintained the original `phaser.d.ts` and as far as I am aware, the `phaser-comments.d.ts` is just the original merged.

photonstorm:
The comments defs file is generated automatically by the grunt script, which parses the comments out of the JS files and merges them with the hand-moderated defs file. So no, the comments file isn't maintained by hand.

photonstorm:
Sorry but I'm closing this PR off. It's been too long and has become so stale, that merging now would be a nightmare. Also the inclusion of the comments files broke the merge ability as well, so if you do ever want to re-open the PR, please just include the defs files, not the auto-generated comments ones.

PR #2439, opened 2016-04-20
Allow debug.line to be used without error TS2346

kiswa:
When calling debug.line('Any string'); the TypeScript compiler give error TS2346 because there are no parameters in the method definition. This creates a dummy parameter which allows the same functionality without the compiler error.


photonstorm:
Hmmm this method shouldn't even be in the TS defs really, as it's not meant to be called directly. I probably ought to swap it from `protected` to `private`. I'll merge your PR in the meantime though.

kiswa:
@photonstorm Is there an official alternative for building up debug text output?

Or should we make our own text and not use the `game.debug` for that?

PR #2451, opened 2016-04-26
typescript defs: Fixed pixi reference at phaser.comments.d.ts

Dreaded-Gnu:
* TypeScript Defs

Fixed typescript definitions by rerunning `grunt tsdoc` for proper use of phaser with `tsd link`.

The reference to `pixi.d.ts` within file `phaser.comments.d.ts` and the reference within package.json to `pixi.comments.d.ts` lead to a lot of duplicate errors, when compiling a typscript project after executing `tsd link`.

After `grunt tsdoc` the reference is set back to `pixi.comments.d.ts` at `phaser.comments.d.ts`.

I hope now everything is okay with the pull request. :)

### DIFF
--- a/typescript/phaser.comments.d.ts
+++ b/typescript/phaser.comments.d.ts
@@ -1,7 +1,7 @@
-/// <reference path="pixi.d.ts" />
+/// <reference path="pixi.comments.d.ts" />
 /// <reference path="p2.d.ts" />
 
-// Type definitions for Phaser 2.4.7 - 22nd April 2016
+// Type definitions for Phaser 2.4.8 - 22nd April 2016
 // Project: https://github.com/photonstorm/phaser
 
 declare module "phaser" {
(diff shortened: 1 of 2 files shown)


photonstorm:
I'm certain that as soon as I run `grunt buildtsdoc` again (which we do as part of the release process), it will overwrite these changes and put us back into the previous situation before.

What we really need to fix this is the Grunt task updated to replace the ref link in the defs file it generates.

Dreaded-Gnu:
Which tasks will be executed, when creating a release (all steps)? I can dive a bit deeper into phaser then to fix it with this kind of Information. :)

photonstorm:
Doesn't really matter, all that matters is the buildtsdoc script handles the string replacement itself, or that a grunt replace task is run immediately after it. This is all grunt task related.

Dreaded-Gnu:
I've checked the replace task for grunt, and everything seems to be okay. When running `grunt build tsdoc` the typescript definitions are also containing correct (replaced) references.

Seems to be that the grunt task `replace:phasertsdefheader` hasn't been executed during build of release 2.4.7.
Currently I'm building phaser using nodejs 6.0. Are you using a different version?

photonstorm:
Going to close this one off because it'll be fixed automatically with the next rebuild (which I'll do today actually as part of RC1), but thank you for highlighting the issue all the same!

PR #2452, opened 2016-04-26
typescript defs: Allow game.physics.arcade.collide(group) for collisions within group

osev7:

* TypeScript Defs


photonstorm:
Thanks.

PR #2472, opened 2016-05-09
BitmapData.generatetexture fix for Safari

mattcolman:

- Nothing, it's a bug fix

Describe the changes below:

This fixes a bug with BitmapData.generateTexture seen in Safari on both desktop and mobile.
http://www.html5gamedevs.com/topic/20258-bitmapdatageneratetexture-doesnt-work-first-time-in-safari/

### DIFF
--- a/src/gameobjects/BitmapData.js
+++ b/src/gameobjects/BitmapData.js
@@ -525,17 +525,14 @@ Phaser.BitmapData.prototype = {
     * @param {string} key - The key which will be used to store the image in the Cache.
     * @return {PIXI.Texture} The newly generated texture.
     */
-    generateTexture: function (key) {
+    generateTexture: function(key) {
+        var copyCanvas = PIXI.CanvasPool.create(this, this.ctx.width, this.ctx.height);
+        var copyContext = newCanvas.getContext('2d', { alpha: true });
+        var imageData = this.ctx.getImageData(0, 0, width, height);
+        copyContext.putImageData(imageData, 0, 0);
 
-        var image = new Image();
-
-        image.src = this.canvas.toDataURL("image/png");
-
-        var obj = this.game.cache.addImage(key, '', image);
-
-        return new PIXI.Texture(obj.base);
-
-    },
+        return PIXI.Texture.fromCanvas(copyCanvas);
+    }
 
     /**
     * Resizes the BitmapData. This changes the size of the underlying canvas and refreshes the buffer.
--- a/src/gameobjects/BitmapData.js
+++ b/src/gameobjects/BitmapData.js
@@ -526,13 +526,16 @@ Phaser.BitmapData.prototype = {
     * @return {PIXI.Texture} The newly generated texture.
     */
     generateTexture: function(key) {
-        var copyCanvas = PIXI.CanvasPool.create(this, this.ctx.width, this.ctx.height);
-        var copyContext = newCanvas.getContext('2d', { alpha: true });
-        var imageData = this.ctx.getImageData(0, 0, width, height);
+        var copyCanvas = PIXI.CanvasPool.create(this, this.width, this.height);
+
+        var copyContext = copyCanvas.getContext('2d', { alpha: true });
+
+        var imageData = this.ctx.getImageData(0, 0, this.width, this.height);
+
         copyContext.putImageData(imageData, 0, 0);
 
         return PIXI.Texture.fromCanvas(copyCanvas);
-    }
+    },
 
     /**
     * Resizes the BitmapData. This changes the size of the underlying canvas and refreshes the buffer.


photonstorm:
I think this is far more likely to be a race condition, than an issue with creating an Image object on Safari. I.e. the function shouldn't return until the Image load event has fired.

BTW this is an API change (not just a bug fix), as the function no longer does what the docs say it does, so it's legacy breaking.

mattcolman:
Yep I agree that it is a race condition, but it only occurs in Safari. I experienced the issue when trying to generate a spritesheet from bmds and graphics objects. With bmd I would call ```
spritesheetBuilder.add(bmd.generateTexture())
bmd.destroy()
``` 
I guess the bmd is getting destroyed before the image load has completed.
The same code with a Graphics object works fine.

Yes sorry, I didn't account for storing the key in the AssetCache...

Maybe it's better to keep your code and return a Promise? Or add a callback for when the image load is complete?

**Comment by photonstorm** [comment]
Will have a play with this, we may be able to get away with just not returning until the onload fires, but will need testing.

**Comment by VitaZheltyakov** [comment]
I got the same problem in Firefox. The above code fixes it, but it does not save the image in the cache.

**Comment by VitaZheltyakov** [comment]
Such code solves this problem.
```
	Phaser.BitmapData.prototype.generateTexture = function (key) {

		var copyCanvas = PIXI.CanvasPool.create(this, this.width, this.height);
		var copyContext = copyCanvas.getContext('2d', { alpha: true });
		copyContext.putImageData(this.ctx.getImageData(0, 0, this.width, this.height), 0, 0);
		var texture = PIXI.Texture.fromCanvas(copyCanvas);

		PIXI.Texture.addTextureToCache(texture, key); //<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<

		return texture;
	};

Phaser.Component.LoadTexture.prototype.loadTexture = function (key, frame, stopAnimation) {

        if (key === Phaser.PENDING_ATLAS)
        {
            key = frame;
            frame = 0;
        }
        else
        {
            frame = frame || 0;
        }

        if ((stopAnimation || stopAnimation === undefined) && this.animations)
        {
            this.animations.stop();
        }

        this.key = key;
        this.customRender = false;
        var cache = this.game.cache;

        var setFrame = true;
        var smoothed = !this.texture.baseTexture.scaleMode;

        if (Phaser.RenderTexture && key instanceof Phaser.RenderTexture)
        {
            this.key = key.key;
            this.setTexture(key);
        }
        else if (Phaser.BitmapData && key instanceof Phaser.BitmapData)
        {
            this.customRender = true;

            this.setTexture(key.texture);

            if (cache.hasFrameData(key.key, Phaser.Cache.BITMAPDATA))
            {
                setFrame = !this.animations.loadFrameData(cache.getFrameData(key.key, Phaser.Cache.BITMAPDATA), frame);
            }
            else
            {
                setFrame = !this.animations.loadFrameData(key.frameData, 0);
            }
        }
        else if (Phaser.Video && key instanceof Phaser.Video)
        {
            this.customRender = true;

            //  This works from a reference, which probably isn't what we need here
            var valid = key.texture.valid;
            this.setTexture(key.texture);
            this.setFrame(key.texture.frame.clone());
            key.onChangeSource.add(this.resizeFrame, this);
            this.texture.valid = valid;
        }
        else if (Phaser.Tilemap && key instanceof Phaser.TilemapLayer)
        {
            // this.customRender = true;

            this.setTexture(PIXI.Texture.fromCanvas(key.canvas));
        }
        else if (key instanceof PIXI.Texture)
        {
            this.setTexture(key);
        }
        else if (typeof PIXI.TextureCache[key] !== "undefined")     //<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
        {
            this.setTexture(PIXI.TextureCache[key]);
        }
        else
        {
	         var img = cache.getImage(key, true);

            this.key = img.key;
            this.setTexture(new PIXI.Texture(img.base));

            if (key === '__default')
            {
                this.texture.baseTexture.skipRender = true;
            }
            else
            {
                this.texture.baseTexture.skipRender = false;
            }

            setFrame = !this.animations.loadFrameData(img.frameData, frame);
        }

        if (setFrame)
        {
            this._frame = Phaser.Rectangle.clone(this.texture.frame);
        }

        if (!smoothed)
        {
            this.texture.baseTexture.scaleMode = 1;
        }

    };
```

VitaZheltyakov:
I believe that the best way - is to use the code above (generateTexture), and add the ability to Phaser.cache add PIXI.Texture.

photonstorm:
Thank you for your PR! Today I released Phaser CE (Community Edition). This is my way of opening up Phaser 2 to the community, allowing you to control features and updates directly. Please see the README in the `v2-community` folder of the repo for more details about this change.

If you could please re-submit your PR against the new Phaser CE version then I will merge it immediately :) Note that PRs now go against the `master` branch, and don't forget to add your change to the README too!

mattcolman:
FYI I don't think I recommend this code anymore. I found conditions where is produces very bad results. I can't remember exactly, but it occurred with complex generated bitmapData instructions, I think from an adobe Illustrator export. I didn't look into further.

ForgeableSum:
@VitaZheltyakov's code works except PIXI.Texture.addTextureToCache(texture, key). How does one add a pixi texture to phaser's cache so that it can be used as an image texture on sprites?

PR #2493, opened 2016-05-21
Adding Alpha Variable to Flash

rgk:

* The public-facing API

Describe the changes below:

This would make the flash less intense if you want that effect. Allowing you to change the alpha value of the flash.


photonstorm:
I like this idea a lot, but I want to find a way to do it that doesn't change the method signature. It should be possible to simply extract the alpha channel from the color value given, then all we need to do is update the jsdocs (and add a little code).

Also please only commit the files you changed, not the whole branch.

photonstorm:
Thanks for submitting this PR. We have implemented our own version based on this, and credited you in the Change Log.

PR #2505, opened 2016-05-26
Added a reverse functionality to animations.

gotenxds:
Hi, I discovered pharser today and its awesome ! :)
I started messing around with it and stumbled on a few limitations in the animation front (can't reverse, flip frames, compose animations and more) so I added a thouse offline, this is my small addition to support reversing animation mid play.

So for example if you have an animation with 10 frames
and it played 7 frames.
now you can call 

```
animation.reverse();
```
and the animation will play frames 6 to 0.

Live example : http://phaserjs-site.orionhub.org:8000/index.html

update:

Added two methods, 
reverse - toggels the reversed state.
reverseOnce - toggles the reverse state for the current/next animation play.
This is really convenient if you want the animation to reverse and return to its original play state after it completes.

Hope you like it =]

### DIFF
--- a/src/animation/Animation.js
+++ b/src/animation/Animation.js
@@ -201,7 +201,7 @@ Phaser.Animation.prototype = {
         this._timeLastFrame = this.game.time.time;
         this._timeNextFrame = this.game.time.time + this.delay;
 
-        this._frameIndex = 0;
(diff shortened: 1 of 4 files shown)


gotenxds:
Looks like the build fails on src/geom/Line.js which I did not touch..

PR #2509, opened 2016-05-28
Arcade circles

VitaZheltyakov:

* Nothing, it's a bug fix

Describe the changes below:
 	Added support for circles in various functions

### DIFF
--- a/src/physics/arcade/Body.js
+++ b/src/physics/arcade/Body.js
@@ -347,7 +347,7 @@ Phaser.Physics.Arcade.Body = function (sprite) {
     /**
     * If true the Body will check itself against the Sprite.getBounds() dimensions and adjust its width and height accordingly.
     * If false it will compare its dimensions against the Sprite scale instead, and adjust its width height if the scale has changed.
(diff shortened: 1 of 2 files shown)


photonstorm:
Do these work against Tilemaps?

VitaZheltyakov:
No, it does not work.
I have completed a pilot test circles on a tiles. The result is terrible. [examples.zip](https://github.com/photonstorm/phaser/files/301486/examples.zip) See example5
Firstly, it is unclear why it is necessary to process a collision circles with tiles (?). Only the collision of corners will differ from collisions with rectangles.
Secondly, there are so many problems in the processing of collisions on the borders of the tiles. The circle can touch the border of the two tiles. This results in an unnatural behavior.

photonstorm:
Because unless it's guarded against it, I 100% guarantee you that it will generate numerous GitHub issues asking why their circle bodies don't work on a tilemap.

VitaZheltyakov:
You can add support for circles collisions with tiles, but with limitations.
Look at my decision in the archive.

I have not worked and do not work with tiles, so I will not move in that direction.

PR #2532, opened 2016-06-05
added spread function

Nuuf:
This PR changes

* The public-facing API

added a math function that spreads values
Phaser.Math.spread


Nuuf:
Delete this one, i'm still learning git and this was a mistake.

PR #2533, opened 2016-06-05
Phaser.Math.spread

Nuuf:
This PR changes
- The public-facing API

added Phaser.Math.spread.
A function that can be used for spreading values evenly across a given start value
for example, if the start value is 100, and you want to spread 10 values with a margin of 5, then you can use this function to spread 5 values to the left of 100 and 5 to the right.
first value will be 77.5 and last will be 122.5


Nuuf:
Could also be useful in Phaser.Group for automatic spreading, like:
`Phaser.Group.prototype.spread = function(startValue, margin, property) {
  for (var i = 0; i < this.children.length; ++i)
{
 var child = this.children[i];
 if (child.exists)
{
 child.property = startValue - (margin * (this.children.length - 1) * 0.5) + (i * margin);
}
}
};`

Or like Group.spreadX/Y only.

I thought it would fit best in Phaser.Math.

photonstorm:
Thanks, I quite like the look of this. Won't it fail however if you've a starting value of 0?

```
if (x && marginX) 
```

As the first comparison will return `false`.

Nuuf:
You're right, my thought was to fit both x and y in one function, choosing wether to spread x, y or both.
I think that if the statement checks for a null value instead, it should work if the start value is 0.

Nuuf:
It might be more convenient and less confusing with an option for Group.spreadOutX and Group.spreadOutY also.

Nuuf:
However I first thought to give the option for a property parameter, for this would also help in "fanning out" for rotation, even velocity perhaps.

PR #2561, opened 2016-06-14
Add Pointer.pointerMode

danzel:

* TypeScript Defs

Describe the changes below:

Adds missing field from Pointer.
Am I meant to update the comments file too or is it auto generated?


photonstorm:
The comments def is generated automatically, so no need to edit that, thanks.

Issue #2554, opened 2016-06-11
Tile map loader does not load all relevant data for object layers


gotenxds:
This Issue is about (delete as applicable)

* A bug in the API

Hi, Tilemap loader does not load some data when the object is an image, 
missing data that I found is width and height, you can can stretch the images in tiled but we cannot get this data since phaser only load it for polygons.

photonstorm:
Please post a sample Tiled export so we can see exactly the properties missing.

photonstorm:
Thanks for the PR. Merged.

PR #2564, opened 2016-06-15
Added some data to image object types.

gotenxds:
Hi, as described here: #2554 
Objects of type image now include size and flipped state data.


photonstorm:
Lovely, thanks.

PR #2576, opened 2016-06-19
Typings.json file to enable deployment of typescript definitions with typings

monagames:
This PR adds a typings.json file in order to allow easy integration of its type definitions with typings. This allows easily building a project using typescript + jspm + systemjs + typings


photonstorm:
Can you edit the `typescript\readme.md` file please, to explain how to use this. As I've no idea, and I need to know, as people will ask :) so it makes sense to just have it in the docs.

monagames:
Added a short explanation. My idea is to register the entry in typings as soon as it makes it to the main branch. This way, including phaser's typescript definitions will be as simple as: `typings install phaser -GD`

photonstorm:
Awesome, thanks.

PR #2577, opened 2016-06-19
Modifed Group.getClosestTo() and Group.getFurthestFrom() to add optional filter callback

LoneStranger:
This PR changes the public-facing API.

A pair of new functions for the Group class, getClosestToFilter() and getFurthestFromFilter(), allowing a filter function to be passed and only members of the group who satisfy the filter will be considered for distance to the target object.

### DIFF
--- a/src/core/Group.js
+++ b/src/core/Group.js
@@ -2213,8 +2213,12 @@ Phaser.Group.prototype.getBottom = function () {
 *
 * 'close' is determined by the distance from the objects `x` and `y` properties compared to the childs `x` and `y` properties.
 *
+* If callback returns true, the object is considered for distance. If callback is null, then any child will satisfy the criteria. 
+*
 * @method Phaser.Group#getClosestTo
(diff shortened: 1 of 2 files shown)


photonstorm:
Thanks for the PR :)

I think there are a couple of things you could tweak, and then I can happily merge this:

1) I don't think we need 2 new functions for this. Let's just make 'filter' an _optional_ argument to the existing getFurthest / getClosest - then it doesn't change the API for anyone already using it, but provides the exact same new functionality as you've added here.

2) Be sure to include the callback context as an argument, and then use `callback.call` to invoke the callback, rather than just hitting it directly like you do at the moment. That way it'll avoid scoping issues and keep it consistent with the way other similar functions work. Look at `Group.iterate` for an example of what I mean.

LoneStranger:
Ok, I took a look at Group.iterate and I modified the original functions instead, adding the optional callback and context.

PR #2585, opened 2016-06-21
Added file type mappings to accept headers.

### LINKED ISSUE #2485: Loader xhrLoad to include headers?
This Issue is about 

* A bug in the API

If you do this.game.load.json(key, url, overwrite). The server should know that you are actually requesting json. Otherwise the server might send you xml (which is default for many servers). 

That happened to me and setting the request header for accept to application/json fixed this. But right now there is no request headers set which is not a good thing.

This is very hacky, but did not have the time to figure out better way to set them so here is my solution for it right now :
```
        var xhr = new XMLHttpRequest();
        xhr.open("GET", url, true);
        xhr.responseType = type;
        if (type === 'json') {
            xhr.setRequestHeader("Accept", "application/json");
            xhr.responseType = 'text';
        }
```

So now I am asking if we should add those headers to the loader and then I might just do a PR with better solution, but not going to do it before asking if this is something people want.

stoneman1:

* Nothing, it's a "bug" fix

Describe the changes below:
Added accept headers to xhrLoad.

Fixes #2485

### DIFF
--- a/src/loader/Loader.js
+++ b/src/loader/Loader.js
@@ -102,8 +102,18 @@ Phaser.Loader = function (game) {
     */
     this.path = '';
 
+	/**
+     * @property {object} Headers - Maps file types to accept headers.
+     * @default
+     */
+
+    this.headers = {
+        json: "application/json",
+        xml: "application/xml"
+    };
+
     /**
-    * This event is dispatched when the loading process starts: before the first file has been requested,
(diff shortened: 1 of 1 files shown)


photonstorm:
Yeah ok, let's add this in. In future please adhere to the coding style set out in the rest of the library though (you'll see what I mean when I edit your code, and re-push it, in a moment)

PR #2593, opened 2016-06-25
Improved typescript definitions

monagames:
This PR changes typescript definitions.

Describe the changes below:
- Some any arguments in Color helpers replaced by more specific types.
- Similar changes for BitmapData.processPixel

Please, review if you are fine with the name given to the interface that represents the object with color properties: ColorComponents


photonstorm:
Looks good to me - does it all still compile ok?

monagames:
Yes, it does. Thanks

PR #2643, opened 2016-07-19
fix documentation for group.align -- row and column were swapped

deargle:
This PR changes

* Documentation
* Public-facing API (sort of)

Group.align had `row` and `column` parameters swapped in many (but not all) places. I think I caught all in the docstring. Did a simple search-replace in the function code.


photonstorm:
Yes you're right, I'll merge this in, but I'm going to remove all mention of `columns` and `rows` because I think it's more confusing than helpful. Will replace with a simple `width` and `height`.

PR #2652, opened 2016-07-19
Remove mozHidden because it is deprecated

leopoldobrines7:

* Nothing, it's a bug fix

Describe the changes below:
I use Firefox 47.0.1, when i was using phaser I saw a warning in console: " 'mozHidden' and 'mozVisibilityState' are deprecated.  Please use the unprefixed 'hidden' and 'visibilityState' instead. ". Then I removed using these variables.


photonstorm:
Hi - Please don't commit build files, as we can then never merge in your PR because it conflicts.

More importantly though, this will remove the warning from newer versions of Firefox, but it'll stop working entirely in older versions.

photonstorm:
Sorry forgot to add - a better check would be to see if `hidden` is available first, and if it is then never even check for the prefix version. That will stop the warning in FF and retain backward compatibility.

PR #2656, opened 2016-07-21
Check if hidden is available first

leopoldobrines7:
This PR changes
* Nothing, it's a bug fix

Check if 'document.hidden' is available first, and if it is then never even check for the prefix version. That will stop the warning (like "mozHidden and mozVisibilityState are deprecated") in newer versions of browsers and retain backward compatibility.


photonstorm:
Awesome, thanks.

PR #2664, opened 2016-07-26
Allow animation speed greater than 0

jayrobin:
This PR changes
* Documentation (jsdoc)
* Nothing, it's a bug fix

`Animation.speed=` sets delay to 1000 / value if value is greater than or equal to 1. However, setting delay directly allows for animation delays greater than 1000. Therefore, changing `Animation.speed=` to allow for speed greater than 0.


photonstorm:
Thanks, happy to merge this change in.

PR #2671, opened 2016-07-28
Fix for DisplayObject/DisplayObjectContainer - getting dimensions or bounds do NOT retrieve proper values

fmflame:
This PR changes:
* TypeScript Defs
* Nothing, it's a bug fix

Describe the changes below:

fixed getLocalBounds() to return the bounds of the DisplayObject without any transformations and adjusted width and height to work the way they did before to be compatible with how PIXI used to work - meaning they will return the getLocalBounds().width * scale.x or getLocalBounds().height* scale.y.

### DIFF
--- a/src/pixi/display/DisplayObjectContainer.js
+++ b/src/pixi/display/DisplayObjectContainer.js
@@ -435,14 +435,14 @@ PIXI.DisplayObjectContainer.prototype.getBounds = function (targetCoordinateSpac
 };
 
 /**
- * Retrieves the non-global local bounds of the displayObjectContainer as a rectangle. The calculation takes all visible children into consideration.
+ * Retrieves the non-global local bounds of the displayObjectContainer as a rectangle without any transformations. The calculation takes all visible children into consideration.
  *
  * @method getLocalBounds
  * @return {Rectangle} The rectangular bounding area
  */
 PIXI.DisplayObjectContainer.prototype.getLocalBounds = function () {
 
-    return this.getBounds(this.parent);
+    return this.getBounds(this);
 
 };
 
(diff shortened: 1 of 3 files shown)


src/pixi/display/DisplayObjectContainer.js
```diff
@@ -577,7 +577,7 @@ PIXI.DisplayObjectContainer.prototype._renderCanvas = function (renderSession) {
 Object.defineProperty(PIXI.DisplayObjectContainer.prototype, 'width', {
 
     get: function() {
-        return this.getLocalBounds().width;
+        return this.getLocalBounds().width + this.scale.x;
```
namark: A typo here.

fmflame:
I added a very nice commit related to the fix in this PR:

https://github.com/photonstorm/phaser/pull/2671/commits/64a02ddef4809c8ea485536a610df5e96cb808da

It fixes top, bottom, left, right, centerX and centerY to use the updated getBounds and compute their values depending on the parent local coordinate space. I think this is better than using the global getBounds and a lot more convenient because it makes total sense x and centerX for example to be from the same coordinate space and also with this change now these values would be working when the group is a top level display object and also when it is parented by something else. So this loosens the global only getBounds restriction. Also it doesn't really makes sense to align objects with totally different parents but if that is the case separate calculations could be made.

fmflame:
Just one more note to this. The commit above fixes alignIn, but I have a request to do a simple tweak for it - make it so that offsetX and offsetY always work the same way no matter if you select Phaser.LEFT_CENTER or Phaser.CENTER; For me personally offsets would follow the rules of the coordinate space, meaning that if it is a positive numer it would be added to the x or y or if negative substracted. At the moment for the different aligns the calculations break this rule and I don't think it is logical or convenient to be this way.

PR #2672, opened 2016-07-29
Weapon.Fire now tracks offset rotation

bobonthenet:

* Nothing, it's a bug fix

Describe the changes below:
I don't think Weapon.Fire is working properly when you are using an offset and tracking rotation. This fixes it so it works the way I think it should.

### DIFF
--- a/src/plugins/weapon/WeaponPlugin.js
+++ b/src/plugins/weapon/WeaponPlugin.js
@@ -730,14 +730,32 @@ Phaser.Weapon.prototype.fire = function (from, x, y) {
     }
     else if (this.trackedSprite)
     {
-        if (this.fireFrom.width > 1)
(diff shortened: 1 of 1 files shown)


photonstorm:
Thanks for the PR and your contribution. I like this change a lot, but I don't want to create a new Point object every time you fire a bullet. Instead please could you update it to use a private local point instead, and just recycle it?

bobonthenet:
So forgive my noobness and thank you for the opportunity to contribute to this project. It means a lot to me to be able to learn from this experience and be a part of this. I think I've done what you've asked and it seems to work in my application. Let me know if I should change anything else.

bobonthenet:
Thanks, his has been an exciting learning experience for me. No need to apologize. I hope to contribute more so I took a look at your commit.

PR #2696, opened 2016-08-17
changed return value of Plugin.Weapon's fire, fireAtXY, fireAtPointer…

JTronLabs:
…, fireAtSprite to Phaser.Bullet


photonstorm:
:+1:

PR #2697, opened 2016-08-18
Sound: Fixes an issue where loopFull would not return the soundinstan…

hilts-vaughan:
This PR changes:

* Nothing, it's a bug fix

Describe the changes below:
 
The documentation states that `loopFull` should return the sound instance, but it did not. This just fixes that.


photonstorm:
:+1:

PR #2729, opened 2016-08-29
Use centerX, centerY in Phaser.Line#fromSprite

samme:
This PR changes

* The public-facing API

This is an incompatible change but since Sprites have never had a `center` property it seems appropriate.


photonstorm:
Hmm the docs are quite explicit about this, but as we're merging with v3 we can afford the change. Merged!

samme:
Inspired by #2654 :)

PR #2742, opened 2016-09-06
Remove use of non-existent begin variable and reference beginIndex instead

### LINKED ISSUE #2741: Uncaught ReferenceError: begin is not defined when using removeChildren
This Issue is about

* A bug in the API

I am upgrading from Phaser 2.4.8 to 2.6.2, using Chrome 52.

When attempting to use the `removeChildren` method, an `Uncaught ReferenceError: begin is not defined error` is being thrown.

Example: http://phaser.io/sandbox/LGcoNJob

alex-espinoza:
This PR changes

* Nothing, it's a bug fix

Fixes #2741


suranyami:
👍

Issue #2758, opened 2016-09-14
Phaser.Math minor docs error & potential improvements


JTronLabs:
This Issue is about (delete as applicable)

* An error in the documentation
![selection_037](https://cloud.githubusercontent.com/assets/7429786/18498905/fd701390-79ff-11e6-9535-7ef126710d54.png)

http://phaser.io/docs/2.6.2/Phaser.Math.html#between

JTronLabs:
probably should add the keyword 'Random' in there too

JTronLabs:
I recommend `Returns a random integer between min and max values (inclusive)`

JTronLabs:
It looks like there's not a function to get a random float between 2 values???

photonstorm:
I've already merged your PR, however there are functions for this already: `RandomDataGenerator.interger` and `RandomDataGenerator.frac`, also `integerInRange` and `realInRange`.

JTronLabs:
I did not know this was a thing. Maybe we should add a link/reference to it in Phaser.Math?

PR #2760, opened 2016-09-15
Phaser.Math.between fixes & added Phaser.Math.random

JTronLabs:

* Documentation
* TypeScript Defs
* The public-facing API

Fixed docs for Phaser.Math.between(min,max), made it a bit more robust, added a random(min,max) function that acts the same as between(), but returns a floating point.

Ref https://github.com/photonstorm/phaser/issues/2758

### DIFF
--- a/src/math/Math.js
+++ b/src/math/Math.js
@@ -27,10 +27,10 @@ Phaser.Math = {
     * Returns a random float in the range `[min, max)`. If these parameters are not in order than they will be put in order.
     * Default is 0 for `min` and 1 for `max`.
     *
-    * @method Phaser.Math#between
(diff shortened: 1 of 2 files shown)


JTronLabs:
I made between() work with integers (before it was whole numbers - positive only). The updated code was copy+ pasted from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random 's `getRandomIntInclusive` function.

JTronLabs:
Any issues with this PR, or things I can improve on?

photonstorm:
Nope it's fine, have merged.

PR #2763, opened 2016-09-19
Add a parameter to load.spritesheet which tells it how many frames to…

arefiev:

* The public-facing API

Describe the changes below:

Add a parameter to load.sprite sheet which tells it how many frames to skip.  Good when you have huge PNGs with multiple separate sprite sheets in each.


photonstorm:
Thanks, this is nice. Looks like a rogue `console.assert` is left in though, but I'll remove that now.

Issue #2851, opened 2016-11-16
Emitters with custom names are not removed from Particles.emitters after destruction


jorio:
If you create a particle emitter through game.add and set its name property afterwards:
```javascript
let e = game.add.emitter(0, 0, 500);
e.name = "stars";
```
When this emitter gets destroyed (for instance when switching states), it is not removed from Particles.emitters.

Particles.add and Particles.remove use the emitter's name as a key in Particles.emitters. Since the emitter was given a default name on creation, and we changed it between invocations of add and remove, Particles.emitters ends up containing a stale emitter.

This is just a minor memory issue; the stale emitters are properly ignored by the update routine since they're not alive anymore.

Here's a [jsfiddle](https://jsfiddle.net/rqo4rd6f/) demonstrating the issue.

hilts-vaughan:
Fix in progress.

hilts-vaughan:
Push fixed for review.

PR #2853, opened 2016-11-17
Issue 2851

hilts-vaughan:

* Nothing, it's a bug fix

Describe the changes below:

Fixes: https://github.com/photonstorm/phaser/issues/2851

Instead of using name to track things, we should just track by internal ID.


hilts-vaughan:
Oops, another commit snuck in there.

PR #2882, opened 2016-11-25
Changelog updated, preparing for 2.7.1 update

stoneman1:

* Documentation (Changelog and Readme changelog)

Describe the changes below:
Updated changelog and readme

### DIFF
--- a/v2-community/CHANGELOG.md
+++ b/v2-community/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Change Log
 
+## Version 2.7.1
+
+### Updates
+
+* TypeScript definitions fixes and updates (thanks @Aleksey-Danchin)
+
+### Bug Fixes
+
+* Phaser.AnimationParser.spriteSheet() now works like it is supposed to work (thanks @stoneman1, @qarlosh)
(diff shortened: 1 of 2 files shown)


photonstorm:
Let me know when you want 2.7.1 published to npm.

stoneman1:
Probably best to publish is as soon as possible. There is that EarCut problem with the current npm version and it might cause some problems with people.

photonstorm:
Ok. In order to publish, the version number in the `package.json` and `src/Phaser.js` files both need updating. Then run `grunt release` so it rebuilds the docs / TS defs and creates the build files.

I don't mind doing the above of course, but I think it makes sense for the community to get used to what's involved.

stoneman1:
Will do that next time. And yeah of course it makes sense for the community to do that 👍

photonstorm:
Oh, you also need to decide if you want versions to have a release name any more too :) It's in the package.json, but it doesn't have to be a name, it could just be a date or something. Up to you all! There's a file in the `resources` folder that contains all of the names I was going to use for versions, and which ones have been used / are left. You can do whatever you want though, but just saying it's there if you want it.

PR #2895, opened 2016-12-01
This issue causes tsc to crap out under certain circumstances

mredbishop:
This PR changes

* TypeScript Defs sorta
* Nothing really, it's a fix for idiots like me using TypeScript

Describe the changes below:

Typings element needs to point to the root typings file.

### DIFF
--- a/v2/package.json
+++ b/v2/package.json
@@ -48,7 +48,7 @@
     "typescript": "1.4.1",
     "yuidocjs": "^0.3.50"
   },
-  "typings": "./typescript/typings.json",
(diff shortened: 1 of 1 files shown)


photonstorm:
Hi - The v2 (2.6.2) version is locked from any future changes. Please could you make this change to the v2-community release version, although I think it's already been done in there - but worth checking anyway!

mredbishop:
@photonstorm I've made the same change there now. 

Where is the npmjs.org package pulled from?

Issue #2910, opened 2016-12-10
Bitmaptext Placement Bug in HDPI displays when resolution > 1


jaredbrookswhite:
This Issue is about (delete as applicable)

* A bug in the API (always say which version you're using!)

## Bitmaptext Placement Bug in HDPI displays when resolution > 1

This is a bug that was introduced between 2.6.2 and the current build.
xOffset, yOffset, xAdvance and kerning do not account for double DPI when parsed from either XML or JSON file formats.

(apologies these examples do not run in safari)
[2.6.2](https://jsfiddle.net/jwhite0042/vu4wumz0/1/)
[current npm 2.7.2](https://jsfiddle.net/jwhite0042/52pwt66n/4/)

This error can be fixed by [dividing all the values by the windows pixel ratio in scr/loader/LoaderParser.](https://github.com/jaredbrookswhite/phaser/blob/master/v2-community/src/loader/LoaderParser.js)* as seen in [this example](https://jsfiddle.net/jwhite0042/413n52fr/2/). However, this does not work when one opts to not address hdpi displays when instantiating the game object, otherwise you end up with [this mess](https://jsfiddle.net/jwhite0042/413n52fr/4/). I can finish this fix and put in a PR, but I didn't want to go too far away from how the community would like to address the issue. Maybe passing the resolution variable from game instantiation down to the parser?

*Edit - Original fix branch from my fork as been removed

jaredbrookswhite:
Addressed in PR #2915.

photonstorm:
I merged your PR, and fixed it a little where it wasn't passing in the `resolution` argument in the method call.

jaredbrookswhite:
awesome, thanks.

PR #2915, opened 2016-12-12
Added Support for HDPI Displays to atlas based BitmapText

jaredbrookswhite:
A simple fix to help account for #2910


jaredbrookswhite:
thanks for merging, sorry about the build in the commit. I saw where you requested people don't do that after I put in the pr

photonstorm:
Yeah if this had been the original Phaser, I couldn't have merged it, but with CE it's ok. It's better in future to not include them (makes the git history cleaner), but it isn't the end of the world either as it's all versioned in npm.

PR #2918, opened 2016-12-13
Enable delta scroll rendering by default for tilemaps

mdutra:

* Nothing, it's a bug fix

Describe the changes below:

In the documentation enableScrollDelta is true by default but I think it was forgotten as false, so I changed to true. This stopped the lag I had when scrolling tilemaps.

### DIFF
--- a/v2-community/src/tilemap/TilemapLayer.js
+++ b/v2-community/src/tilemap/TilemapLayer.js
@@ -96,7 +96,7 @@ Phaser.TilemapLayer = function (game, tilemap, index, width, height) {
     * @default
     */
     this.renderSettings = {
-        enableScrollDelta: false,
+        enableScrollDelta: true,
         overdrawRatio: 0.20,
         copyCanvas: null
     };


cloakedninjas:
Would the fix for this not be to update the documentation instead? Changing defaults to suit your needs has wider implications.

mdutra:
You're right, maybe it's the documentation that needs change. But here's the comment about this property:
```javascript
/**
*     Settings that control standard (non-diagnostic) rendering.
*
*     @property {boolean} [enableScrollDelta=true] - Delta scroll rendering only draws tiles/edges as they come into view.
*     This can greatly improve scrolling rendering performance, especially when there are many small tiles.
*     It should only be disabled in rare cases.
*     ...
*/
```
So that's why I thought it was a mistake with the value.

mdutra:
By the way, about a year ago, these people were having performance problems similar to mine:
http://www.html5gamedevs.com/topic/20202-fps-drop-with-camera-movement-tilemaps/
http://www.html5gamedevs.com/topic/19507-upgrading-game-from-222-to-244-kills-performance/
Unfortunately, it's too late to see if changing `enableScrollDelta` would solve their problem.

photonstorm:
There is probably a good reason this was changed from `true` to `false` (and the docs weren't updated to reflect it), but I fail to remember what, so I'll merge this in. If it causes significant problems for people then it can be reverted.

PR #2957, opened 2017-01-16
Invalid variable in BaseCache

vulvulune:
This PR fixes a bug:

- In the constructor, the variable in entries, but files in the methods => replace files by entries.


photonstorm:
Yes, isn't complete yet, but merging anyway :)

PR #2959, opened 2017-01-16
Update Basetransform: using modules

vulvulune:
Use modules in BaseTransform


photonstorm:
Yeah I'm not sure we're going to keep BaseTransform, but I'll merge this all the same - won't hurt.

PR #2967, opened 2017-01-18
Remove duplicate request fullscreen method names

vulvulune:


GGAlanSmithee:
Seems to me the strings differ by the capitalization of the "s" in full**S**creen, is this change safe?

photonstorm:
Both forms are needed to pick-up legacy full screen request strings. The w3c standard version is with lower-case 's'. Even so I think this probably ought to move to a polyfill, rather than the device class.

vulvulune:
I didn't see the capitalization. Sorry...

PR #2969, opened 2017-01-18
Fix: Particle Render. global variable

vulvulune:
declare _indexB_ to avoid implicit global variable

For more details:
http://stackoverflow.com/questions/1758576/multiple-left-hand-assignment-with-javascript


photonstorm:
Just to say it's probably best to ignore anything in the `renderer` folder for now, as it's all going to be changing _lots_.

PR #2970, opened 2017-01-18
Minor fix: Use property instead private variable

vulvulune:


photonstorm:
Using `dirty` forces it to get the value from a getter. Using `_dirty` gets the value directly.

vulvulune:
I know the difference between the 2 syntaxes are the both are legal and works in this case.

But what is the rule in Phaser with "private" variable (starting with _)?

I come from real object language (Java, C#) where the rule is to always access the getter except inside the class, and I think is a good rule.

For exemple, in the GameObject class:
`pivotX: {
        enumerable: true,
        get: function ()
        {
            return this.transform._pivotX;
        },
       set: function (value)
        {
            this.transform._pivotX = value;
            this.transform.dirty = true;
            this.transform.updateCache();
        }
    }`

In the setter, the variable `_pivotX` is set and after some other things are done on the transform object, why not use `this.transform.pivotX = value` and remove the 2 last lines.

With modern JS engine, the performance difference between a getter and access the private variable is almost null.

photonstorm:
`_property` is a warning to developers more than anything, to not touch it from their games. Internally though, I'm not as strict about that, especially in really hot areas of code, where every ms counts.

PR #2971, opened 2017-01-18
Bug fix: replace scale.x by scaleX

vulvulune:
The property scale is a number. so try to access scale.x is invalid.


photonstorm:
Perhaps it would be helpful if I marked the top of each file to show which hasn't been converted yet? (as this one is only partially done)

vulvulune:
Yes, it could be a good idea. I find these bugs adding the types for the TypeScript definitions. I don't do a code review.

Kizilaslan:
Mehmet Kızılaslan

PR #2990, opened 2017-02-19
Fix typo in Color

johnfn:
I noticed a small typo.

### DIFF
--- a/v3/src/graphics/color/Color.js
+++ b/v3/src/graphics/color/Color.js
@@ -25,7 +25,7 @@ var Color = function (red, green, blue, alpha)
     this.setTo(red, green, blue, alpha);
 };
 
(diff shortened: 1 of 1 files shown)


photonstorm:
@stefos96 the attitude is unnecessary - it hasn't been merged because we're debating the need for the class in its current form

antoine-pous:
In the meantime as it's a minor change it could be validated. Being forced to modify Phaser for so little while it might be available on the master branch is not really fun.

Nothing prevents to continue the debate on its implementation despite validating the PR :wink:

photonstorm:
The fact it's the master branch is irrelevant. v3 still compiles and doesn't throw any warnings at runtime even with the typo. While I agree it should be changed, it's hardly life threatening.

PR #3013, opened 2017-05-01
[Phaser 3] Upgrade of development dependencies

rblopes:
Hi, @photonstorm.

I'm issuing this pull request including some important updates that may interest you. Please, make sure of testing these changes before applying them. I can update this PR if requested.

This PR is mostly an upgrade of the tooling used to develop and build Phaser 3, mainly an upgrade of Webpack itself. It seems you're still using a beta release of Webpack 2, instead of a stable one. Included in this PR is an upgrade to Webpack 2.4.1, and a little fix in the config file to work with it. I also added a runnable script, so everyone can build Phaser 3 just using `npm run build` (or `yarn run build`).

I also added ESLint to the project and its runnable script. I know, an ESLint configuration is already there, but `eslint` (the package) is not. Like Webpack, running ESLint against the project source code should be much easier now, and you can also `--fix` (although, you'll need to `{npm|yarn} run lint -- --fix`, note the double dashes before the flag.) any linting warnings.

Below is a summary of the current changes.

---

- Removal of the following redundant or unused packages:

  - `json-loader`: Webpack 2 can handle JSON files by default;
  - `copy-webpack-plugin`: unused;
  - `webpack-dev-server`: unused.

- Upgrade of the following packages to their latest versions:

  - `fs-extra`;
  - `webpack`: upgraded to the latest stable version;
  - `webpack-shell-plugin`.

- Addition of the following packages:

  - `uuid`, in replacement of the deprecated `node-uuid`, with no changes to the public API.

  - `eslint`, with ESLint linting utility.

- Addition of two runnable scripts to help development tasks:

  - `build`: An alias for running `webpack`.
  - `eslint`: Alias for running ESLint using the provided configuration in the project.

- Fix the package entry point, so package consumers will be capable to use library in a future release.

- Update `.npmignore`, so npm will include only the relevant package files.

- Some clean-up and a few code fixes.


photonstorm:
Thanks for doing this :) I've merged it, pulled and run some test builds and it all seems fine here, so thank you :)

PR #3015, opened 2017-05-17
Point compares "false" to null or undefined

GameDevFox:

* Nothing, it's a bug fix

Describe the changes below:

A slight change so that following code will return false instead of throwing a ReferenceError
```
const somePoint = new Phaser.Point();
const x = null;
return somePoint.equals(x)
```

### DIFF
--- a/v2/src/geom/Point.js
+++ b/v2/src/geom/Point.js
@@ -274,7 +274,7 @@ Phaser.Point.prototype = {
     */
     equals: function (a) {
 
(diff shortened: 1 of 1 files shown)


photonstorm:
Thank you for taking the time to work on this and open your pull request. However, Phaser 2 is now maintained in the [Phaser CE GitHub repo](https://github.com/photonstorm/phaser-ce). If the issue still exists in Phaser CE then please submit your PR there and I'm sure the community will be only too happy to merge it.

PR #3021, opened 2017-06-15
Change mask bmd blend mode

khachatur91:

* Nothing, it's a bug fix

Describe the changes below:
Change mask bmd blend mode to make it invisible, if the source has transparent pixels.
In my case it was weird when I use an alphamask on my custom shaped source image which has transparent parts in it, and the mask bmd is being drawn over those parts. I assume that this update would cause more logical behavior for alphamask. I'm not aware of blending modes, but the blendSourceIn one worked for me.

### DIFF
--- a/v2/src/gameobjects/BitmapData.js
+++ b/v2/src/gameobjects/BitmapData.js
@@ -1749,11 +1749,11 @@ Phaser.BitmapData.prototype = {
 
         if (maskRect === undefined || maskRect === null)
         {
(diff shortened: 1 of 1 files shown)


photonstorm:
Thank you for taking the time to work on this and open your pull request. However, Phaser 2 is now maintained in the [Phaser CE GitHub repo](https://github.com/photonstorm/phaser-ce). If the issue still exists in Phaser CE then please submit your PR there and I'm sure the community will be only too happy to merge it.

PR #3022, opened 2017-06-25
Assorted minor fixes

alexrford:
This PR changes:
* Nothing, it's a bug fix

This PR fixes a few minor bugs and potential sources of bugs:
- The first commit fixes some `require` calls where the `'` and `)` characters were transposed.
- The second moves some instances of `var _this = this` binds s.t. they appear before `_this` is first used.
- The third replaces some `typeof children !== null` (always true) with `children !== null`
- The final commit adds some `var` declarations for variables that appear to be intended as locally scoped variables, but were scoped globally due to the lack of declarations.

These results were found on https://lgtm.com/projects/g/photonstorm/phaser/alerts/ - there are further results there, most of which would probably be best handled by people more familiar with the codebase. If you are interested in this kind of analysis, then [Pull Request Integration](https://lgtm.com/projects/g/photonstorm/phaser/ci/) is available. (Full disclosure - I work on lgtm.com).

PR #3025, opened 2017-06-26
Lgtm fixes

alexrford:
* Nothing, it's a bug fix

This is a new version of #3022 with changes restricted to `v3/src`. There are a few additional changes here in response to a few other things I found flagged up. 

22237d8..7d09434 are small maintainability related changes. d522461 fixes what appears to be a minor bug - `<` has higher precedence than `|`, so the right hand operands of these `&&` expressed `({colors,alphas}.length < (vertices.length / 2)) | 0` rather than truncating `vertices.length / 2`, causing an error to be skipped when `vertices.length === {colors,alphas}.length * 2 + 1`.

Results taken from https://lgtm.com/projects/g/photonstorm/phaser/alerts/.

### DIFF
--- a/v3/src/gameobjects/mesh/Mesh.js
+++ b/v3/src/gameobjects/mesh/Mesh.js
@@ -40,12 +40,12 @@ var Mesh = new Class({
             throw new Error('Phaser: Vertex count must match UV count');
         }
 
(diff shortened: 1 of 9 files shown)


alexrford:
Thanks for the review - I've removed the two mentioned commits.

photonstorm:
Thanks, happily merged now.

PR #3040, opened 2017-08-10
Add missing comma to Frame Rate Config

jaminscript:
This PR changes:

* Documentation

Describe the changes below:

As I was experimenting with Phaser 3 I copied and pasted the example FPS config only to find it was missing a comma.


photonstorm:
Actually, `max` needs removing entirely now! But I'll merge until I update it :)

PR #3048, opened 2017-09-12
Uncomment Loader.loadArray

igorpetushkov:
Should Loader.loadArray be uncommented?  This produced an errors while running 'render to texture' example where file array is used.


photonstorm:
Actually, I should remove it entirely. All loader file types (image, etc) can now take arrays as standard, so there is no longer a need for a specific `loadArray` method. It's the example that is outdated.

PR #3050, opened 2017-09-13
Fix Sprite animation with BuildGameObject

igorpetushkov:
The flow **GameObjectCreator -> SpriteCreator -> BuildGameObject** doesn't add sprite object into **sys.updateList** list.

But **GameObjectFactory** does.

To reproduce:
- this.add.sprite(400, 400, 'gems').play('ruby'); **works**
- this.make.sprite(config) **doesn't work**


photonstorm:
👍

PR #3053, opened 2017-09-14
fix Range produce repeat +1 items

igorpetushkov:
To reproduce:
```
var group = this.add.group();
group.createMultiple({ key: 'cube', frame: 'frame1', repeat: 107 });
```
Range returns array with length 108, but 107 has to be. As result, Group has 108 entries with repeat 107 in config.

### DIFF
--- a/v3/src/utils/array/Range.js
+++ b/v3/src/utils/array/Range.js
@@ -86,7 +86,7 @@ var Range = function (a, b, options)
         }
     }
 
-    for (var i = 0; i <= repeat; i++)
+    for (var i = 0; i < repeat; i++)
     {
         var chunk = BuildChunk(a, b, qty);
 


photonstorm:
This is the correct behavior for `Range.repeat`. The idea is that it _repeats_ the output data the number of times given:

```
Range ([a,b,c], [1,2,3], repeat: 1) =
a1, a2, a3, b1, b2, b3, c1, c2, c3, a1, a2, a3, b1, b2, b3, c1, c2, c3
```

In the above, `repeat 1` means 'repeat the whole range once'. As it's a repeat you would get 2 lots of objects back again, the first set and then the repeated set. Much like if you had a Tween with 'repeat 1' it would play the Tween through twice, the first run followed by the repeated run.

I'm happy this function works as intended. The question then becomes should `repeat` in the context of `createMultiple` be changed to something else so you don't have to mentally account for the 'off by one'.

igorpetushkov:
Yeah, I see, got it. Missed the fact that it needs to take the first set/run. 
Thanks for explanation :) 
Closing PR.

Kitanga:
@photonstorm This adds a lot of confusion in the Group.createMultiple() function. Quite curious as to why you didn't use something like size?

PR #3055, opened 2017-09-16
Should sprite use it's custom anims config or only base animation config if it's existed?

igorpetushkov:
1. We create animation with attributes delay/repeat/repeatDelay/yoyo.
2. We create sprite (via _make_) with custom animation (anims) config and with key of the animation which we have created in 1 step.
3. When sprite animation is loading it's overwritten by base animation config.

Question: should sprite use it's custom anims config or only base animation config if it's existed?

Reproduce:
```
this.anims.create({ key: 'blast', 
   frames: this.anims.generateFrameNames('lazer', 
      { prefix: 'lazer_', start: 0, end: 22, zeroPad: 2 }), repeat: -1, repeatDelay: 5 });

this.make.sprite({
        key: 'lazer',
        frame: 'lazer_01',
        x: 50,
        y: 50,
        anims: {
            key: 'blast',
            repeat: -1,
            repeatDelay: 7,
            delayedPlay: function ()
            {
                return 1;
            }
        }
    });
```
Will result sprite animation with repeatDelay = 5 but looks like it should be 7.

### DIFF
--- a/v3/src/animation/frame/inc/Load.js
+++ b/v3/src/animation/frame/inc/Load.js
@@ -14,10 +14,10 @@ var Load = function (component, startFrame)
         component.duration = this.duration;
         component.msPerFrame = this.msPerFrame;
         component.skipMissedFrames = this.skipMissedFrames;
-        component._delay = this.delay;
-        component._repeat = this.repeat;
-        component._repeatDelay = this.repeatDelay;
-        component._yoyo = this.yoyo;
+        component._delay = component._delay || this.delay;
+        component._repeat = component._repeat || this.repeat;
+        component._repeatDelay = component._repeatDelay || this.repeatDelay;
+        component._yoyo = component._yoyo || this.yoyo;
         component._callbackArgs[1] = this;
         component._updateParams = component._callbackArgs.concat(this.onUpdateParams);
     }


photonstorm:
Good point. It fails because there is already an animation defined with that key, and they need to be unique. Two choices: We allow newly defined animations with the same key to overwrite old ones (which could create quite a mess), or we throw a run-time warning if they try to add an animation with an existing key. I think I probably prefer the warning myself.

igorpetushkov:
Mm, what do you think about this case?

We will create a lot of animations with different configs and frames via `this.anims.create({...})`.
Then we will just use them within sprites. Will not create new one every time for new sprite but just take already created and have the choice to override config attributes if needed. We can also throw something like `override: true` in config of the sprite anims.

photonstorm:
We could do. It feels to me like `override: true` is just a safety blanket allowing devs to be messy in their code :) Really, they should be planning more carefully which animations are loaded from configs, and which are created directly in their own code. Which is why a warning tends to enforce this.

However, I guess there are situations when it may be needed. Perhaps loading an external set of animation data generated by an app of some kind, that you wish to override some built-in defaults. So perhaps `override` makes sense, and if that isn't set then a warning is thrown.

igorpetushkov:
Yeah, it will be useful for an app to have `override` flag. For example, for the future phaser editor (my pet project) where you will want to load all animations and then use some of them with custom attributes, save them as base, etc... :)

Will modify my PR.

igorpetushkov:
Mm, another one question :)

Should this code create a new animation? According to the code, I see that animation `new_blast` is trying to be loaded. 
```
this.make.sprite({
        key: 'lazer',
        frame: 'lazer_01',
        x: 50,
        y: 50,
        anims: {
            key: 'new_blast',
            repeat: -1,
            repeatDelay: 7,
            delayedPlay: function ()
            {
                return 1;
            }
        }
    });
```

If not, then should we create `new_blast` animation and then load it for the sprite?

igorpetushkov:
Looks like I need to slow down a bit with `override` idea and keep things as they are. Not critical issue and it can wait for a better solution in the future :)

photonstorm:
The point of the `anims` property in the sprite config is to tell it which animation to use to play, it wasn't meant to create the animation as well, although I can see why that would be useful! :) I do think it might be easier to restrict the functionality at this time though, so if the animation exists, it just adds it to the sprite. If it doesn't, nothing bad happens.

igorpetushkov:
Agree with restriction of that functionality for current state :)
Will close the PR and keep things as they are.

PR #3056, opened 2017-09-16
Move loading __DEFAULT and __MISSING images to boot function.

igorpetushkov:
**TextureManager** loaded ___DEFAULT_ and ___MISSING_ images inside constructor function.
When we use Phaser.WEBGL **TextureSource** is trying to _createTexture_ inside _init_ function. [TextureSource, line 39]

But **game.renderer** is null because it's not yet created.

With Phaser.AUTO and Phaser.CANVAS there are no any errors.

screenshot

createtexture

### DIFF
--- a/v3/src/boot/Game.js
+++ b/v3/src/boot/Game.js
@@ -72,6 +72,8 @@ var Game = new Class({
 
         AddToDOM(this.canvas, this.config.parent);
 
+        this.textures.boot();
+
         this.anims.boot(this.textures);
 
         this.scene.boot();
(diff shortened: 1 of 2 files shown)


photonstorm:
I'll merge this in because it's sensible :) but I think it needs to be more robust - at the moment it doesn't wait for the DEFAULT or MISSING onload event to fire, it just carries on going through setting up the systems. In some cases the delay is enough that it doesn't matter (for example in Chrome I never get an error) but sometimes if your `create` function needs one of those textures it can error because they've not finished 'loading' yet.

igorpetushkov:
agree :)

PR #3057, opened 2017-09-17
[.Offscreen-For-WebGL-XXXXXXXX] render error

igorpetushkov:
```
this.add.text(350, 250, '', { font: '16px Courier', fill: '#00ff00' });
```

Text with empty value produces WebGL render error.

no_texture

(**TextWebGLRenderer** call **SpriteBatch.addSpriteTexture** with _src.canvasTexture = null_)


photonstorm:
I'm sure I used to have that check in there! Or maybe the check is in the Text constructor (setting the '' to a single space). Either way, am merging this in :)

PR #3058, opened 2017-09-17
Fix 'scene is not defined' in GroupCreator

igorpetushkov:
```
var group = this.make.group({ key: 'block', frameQuantity: 12 });
```
scene_is_not_defined


photonstorm:
Thanks, missed that one from the recent update to the Factories.

PR #3062, opened 2017-09-26
Allow for keys to be numbers

Twilrom:

* Nothing, it's a bug fix

Describe the changes below:

Allow for keys to be numbers as well as strings. It will map to a string, but 1 is just so much shorter than '1'

### DIFF
--- a/v3/src/animation/frame/inc/GetFrames.js
+++ b/v3/src/animation/frame/inc/GetFrames.js
@@ -17,8 +17,8 @@ var GetFrames = function (textureManager, frames)
     var i;
     var textureKey;
 
-    //  if frames is a string, we'll get all the frames from the texture manager as if it's a sprite sheet
-    if (typeof frames === 'string')
+    //  if frames isn't an object, we'll get all the frames from the texture manager as if it's a sprite sheet
+    if (typeof frames !== 'object')
     {
         textureKey = frames;
 
--- a/v3/src/gameobjects/BuildGameObjectAnimation.js
+++ b/v3/src/gameobjects/BuildGameObjectAnimation.js
@@ -9,12 +9,7 @@ var BuildGameObjectAnimation = function (sprite, config)
         return sprite;
     }
 
-    if (typeof animConfig === 'string')
-    {
-        //  { anims: 'key' }
-        sprite.anims.play(animConfig);
-    }
-    else if (typeof animConfig === 'object')
+    if (typeof animConfig === 'object')
     {
         //  { anims: {
         //              key: string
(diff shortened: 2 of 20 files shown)


icbat:
If accepted, it seems like you could also use a bool as your key, or a function, right? Anything that's not explicitly an object?

Twilrom:
Sure. They'd all get converted to a string though.

photonstorm:
It would have to map to an Object property. So using a boolean would get converted into `true` or `false` as the property key. A function equally would work, but the text of the function itself is stored (unless invoked, in which case the return value is), so you'd have to retrieve it using heck of a key:

```
var bob = { x: 0 };
bob[true] = 98
bob[function(){}] = 'hello';
bob[function wibble(){}] = 'hello2';

Object.keys(bob);
(4) ["x", "true", "function (){}", "function wibble(){}"]
```

I mean it's beyond horrendous to look at, but technically it works.

photonstorm:
I'm struggling over this PR. Personally, I don't like the idea of using anything other than strings for keys. There's no reason you'd ever use a purely numeric key, as it doesn't identify anything. It feels like it just encourages lazy behavior.

On the flip-side, there's no reason not to allow it either. It makes no difference technically or performance wise. The only area it adds complexity is to the documentation, although I'd be inclined to not even mention it as a valid datatype to avoid that (and to discourage from actually doing it.)

Going to wait for @bitnenfer review before we decide.

Twilrom:
I'm really just trying to get away from using keys at all. I could maybe do
it in a cleaner way though...

On Sep 26, 2017 17:24, "Richard Davey" <notifications@github.com> wrote:

I'm struggling over this PR. Personally, I don't like the idea of using
anything other than strings for keys. There's no reason you'd ever use a
purely numeric key, as it doesn't identify anything. It feels like it just
encourages lazy behavior.

On the flip-side, there's no reason not to allow it either. It makes no
difference technically or performance wise. The only area it adds
complexity is to the documentation, although I'd be inclined to not even
mention it as a valid datatype to avoid that (and to discourage from
actually doing it.)

Going to wait for @bitnenfer <https://github.com/bitnenfer> review before
we decide.

—
<https://github.com/photonstorm/phaser/pull/3062#issuecomment-332234606>,
or mute the thread

GGAlanSmithee:
This would also allow for `null` and `undefined` to be set as properties, which could invite some weird bugs.. In general, I guess it's better to whitelist than blacklist, maybe `typeof frames === 'string' || typeof frames === 'number'` ?

photonstorm:
Sorry, closing this one off. Keys should be strings.

PR #3063, opened 2017-09-27
fix KeyCombo [TypeError: Cannot read property 'on' of undefined]

igorpetushkov:
Don't sure if it's a correct fix for this bug.

keycombo


igorpetushkov:
ah, wait, have a new idea..

igorpetushkov:
ah, wait, have a new idea... (done, fixed PR)

photonstorm:
Thanks :)

igorpetushkov:
:)

Issue #3069, opened 2017-10-10
Phaser breaks for browsers that don't support URL API v3.0.0-beta.6


pavle-goloskokovic:
This Issue is about:

* A bug in the API v3.0.0-beta.6

I've created a repo for purposes of testing Phaser 3 here: 
[https://github.com/pavle-goloskokovic/phaser3-test](https://github.com/pavle-goloskokovic/phaser3-test)

If you try to run code from this repo in any browser that does not support [URL API](https://developer.mozilla.org/en-US/docs/Web/API/URL) it breaks on this line:
[https://github.com/photonstorm/phaser/blob/master/v3/src/loader/filetypes/ImageFile.js#L75](https://github.com/photonstorm/phaser/blob/master/v3/src/loader/filetypes/ImageFile.js#L75)
with `URL` being `undefined` and this error message:
> Unhandled Error: Cannot convert 'URL' to object

Here is a table of browser support for URL API: 
[https://developer.mozilla.org/en-US/docs/Web/API/URL#Browser_compatibility](https://developer.mozilla.org/en-US/docs/Web/API/URL#Browser_compatibility)

photonstorm:
Have added a window.URL polyfill that should sort this out.

PR #3071, opened 2017-10-12
Adding fallback for loading files on browsers where URL API is not supported

pavle-goloskokovic:
This PR changes:
* Nothing, it's a bug fix

Issue [https://github.com/photonstorm/phaser/issues/3069](https://github.com/photonstorm/phaser/issues/3069) was not fixed by adding `URL` polyfills as they too depend on presence of `URL API`:

[https://github.com/photonstorm/phaser/commit/3a096bc7ec63324df67aff1e63ba056b2e8c2a95#diff-4d4f29fa826d75f111b57ec7ff9c743aR290](https://github.com/photonstorm/phaser/commit/3a096bc7ec63324df67aff1e63ba056b2e8c2a95#diff-4d4f29fa826d75f111b57ec7ff9c743aR290)
[https://github.com/photonstorm/phaser/commit/3a096bc7ec63324df67aff1e63ba056b2e8c2a95#diff-4d4f29fa826d75f111b57ec7ff9c743aR294](https://github.com/photonstorm/phaser/commit/3a096bc7ec63324df67aff1e63ba056b2e8c2a95#diff-4d4f29fa826d75f111b57ec7ff9c743aR294)

I added a fallback where if `URL API` is not supported (usually on old browsers) it falls back to creating  `Base64` encoded url using `FileReader` and using it instead.
This logic is placed in two newly added static methods of `File` class and used in it's sub-classes instead of corresponding `URL` methods.

- `URL API` is still used where it is supported so this change will not affect existing behavior, it will just properly load files where `URL API` is NOT supported.
- `URL` polyfill is removed as it is not required and it did not fix anything.
- An issue where `URL` methods were called unnecessarily for `SVGFile` class is now fixed too.

Here is a repo with example where all affected file types are loaded using `phaser` build that includes this update:
[https://github.com/pavle-goloskokovic/phaser3-test](https://github.com/pavle-goloskokovic/phaser3-test)


photonstorm:
Thanks. Let's give this a go and see what happens in the next beta.

Issue #3075, opened 2017-10-24
moveDown a scene throws an error


mindcity:
using Phaser-3.0.0-beta.7, if the initial scene config is like below, then `moveDown(sceneC) ` operation will be failed.
```
let config = {
  scene: [
    { key: 'sceneA', active: true},
    { key: 'sceneB', active: false},
    { key: 'sceneC', active: true}
  ]
}

//in sceneC
this.scene.moveDown();
```
throws an error
```
Uncaught TypeError: Cannot set property 'index' of undefined
    at GlobalSceneManager.SwapPosition [as swapPosition] (SwapPosition.js:27)
    at GlobalSceneManager.MoveDown [as moveDown] (MoveDown.js:22)
    at SceneManager.update (SceneManager.js:76)
    at Systems.step (Systems.js:121)
    at Game.step (Game.js:220)
    at TimeStep.step (TimeStep.js:457)
    at step (RequestAnimationFrame.js:26)
```

please see the example in JSBin: http://jsbin.com/xunikahewi/edit?html,js,output

bobonthenet:
I created a new PR to try and fix this. I'm not sure if I understood everything correctly but hopefully, I did and this is fixed with my PR.

photonstorm:
Fixed in most recent commit.

PR #3076, opened 2017-10-24
Fixed SwapPosition bug

bobonthenet:

* Nothing, it's a bug fix

Describe the changes below:

This is a possible fix for #3075. The index of the scene in the scenes array can't be used to identify the scene in the active array. They won't necessarily have the same index. So instead I loop through the active scenes to find the right one and then set the index property. I also added logic to skip the swap if index1 matches index2. This won't always be caught by the initial if statement comparing scene1 and scene2.


photonstorm:
Closing due to SceneManager rewrite.

PR #3086, opened 2017-11-03
Some bug fixes

Twilrom:

* Nothing, it's a bug fix

Describe the changes below:

Some bug fixes, and removal of unnecessary closePath. closePath moves the pen back to the start in order to "close the loop". It does not need to be called for every path.

### DIFF
--- a/v3/src/curves/cubicbezier/CubicBezierCurve.js
+++ b/v3/src/curves/cubicbezier/CubicBezierCurve.js
@@ -74,7 +74,6 @@ var CubicBezierCurve = new Class({
         }
 
         graphics.strokePath();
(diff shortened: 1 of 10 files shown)


Twilrom:
Fixed

photonstorm:
Thanks. I'm dubious about the `gameObject.input` check removal, because I'm sure that was added specifically to catch cases where a game object is input _disabled_ during a down event, and stopping it then falling over in the up event. I'll merge anyway and see how it pans out, can always roll back easily enough.

Twilrom:
Ah, of course. I tend to make a lot of faulty assumptions about other people's code. I figured it was leftover code from restructuring.

PR #3089, opened 2017-11-14
Bugfixes

Twilrom:

* Nothing, it's a bug fix

Describe the changes below:

Fixed various bugs. Also worked around using "delete". A bit worried I missed something and there was a reason for using delete. Possibly using "in" or "hasOwnProperty" on the objects from outside the class, but I don't see why you would do that.
I don't really understand why "in" and "hasOwnProperty" are used so much. They're slow and dumb.

### DIFF
--- a/v3/src/camera/2d/Camera.js
+++ b/v3/src/camera/2d/Camera.js
@@ -9,7 +9,7 @@ var Camera = new Class({
 
     function Camera (x, y, width, height)
     {
-        this.scene;
+        this.scene = null;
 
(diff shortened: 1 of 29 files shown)


photonstorm:
Thanks for this PR, but I cannot merge such a massive change wholesale. There are elements of these changes we would happily merge, and others that we won't (all the properties being assigned to `null` for example).

As lots of smaller PRs they can be reviewed on their own. Some merged instantly and others discussed.

I trust you understand.

PR #3091, opened 2017-11-15
Bugfixes

Twilrom:

* Nothing, it's a bug fix

Describe the changes below:

Just fixing some minor bugs and typos

### DIFF
--- a/v3/src/events/EventDispatcher.js
+++ b/v3/src/events/EventDispatcher.js
@@ -212,7 +212,7 @@ var EventDispatcher = new Class({
     {
         for (var binding in this.bindings)
         {
-            binding.destroy();
+            this.bindings[binding].destroy();
         }
 
         this.bindings = {};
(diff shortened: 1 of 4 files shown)


v3/src/events/EventDispatcher.js
```diff
@@ -212,7 +212,7 @@ var EventDispatcher = new Class({
     {
         for (var binding in this.bindings)
         {
-            binding.destroy();
+            this.bindings[binding].destroy();
```
GGAlanSmithee: This doesn't seem right.

[`for ... in`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in) will iterate `this.bindings` properties.

`this.bindings[binding]` would be the proper syntax if we were iterating the keys;
`for (var key in Object.keys(this.bindings))`

Twilrom: for...in iterates over the property names/keys.

GGAlanSmithee: my bad, you are right

PR #3096, opened 2017-11-20
Removing some uses of "delete"

Twilrom:

* Nothing, it's a bug fix

Describe the changes below:

Removed some uses of delete, since it breaks optimization.

### DIFF
--- a/v3/src/events/EventBinding.js
+++ b/v3/src/events/EventBinding.js
@@ -262,7 +262,7 @@ var EventBinding = new Class({
 
     destroy: function ()
     {
-        delete this.dispatcher.bindings[this.type];
+        this.dispatcher.bindings[this.type] = undefined;
 
         this.active.length = 0;
 
--- a/v3/src/events/EventDispatcher.js
+++ b/v3/src/events/EventDispatcher.js
@@ -15,15 +15,12 @@ var EventDispatcher = new Class({
 
     getBinding: function (type)
     {
-        if (this.bindings.hasOwnProperty(type))
-        {
-            return this.bindings[type];
-        }
+        return this.bindings[type];
     },
 
     createBinding: function (type)
     {
-        if (!this.getBinding(type))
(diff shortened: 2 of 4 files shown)


photonstorm:
Why are there so many uses of `typeof` for an undefined check?

` if (typeof this._beforeCallbacks[key] !== 'undefined')`

I don't see the need to invoke `typeof` here, it could simply be:

` if (this._beforeCallbacks[key] !== undefined)`

Also, I'm dubious about the wholesale replacement of `delete` for `undefined`. They're two totally different things. In the Data Store we use delete because we want to actually get rid of that key from the object. Not just blank it, but literally remove it. I agree that mutation is bad for hot objects, but that doesn't really count the Data Store or any `destroy` call (but may include the shaderCache, would have to look further at that one).

Twilrom:
typeof is ever so slightly faster than comparing against undefined, and I'm
a sucker for pointless speed gains. I could definitely change that though.
Comparing against undefined makes for smaller code after all.
I didn't see anywhere where setting to undefined would have a different
effect than deleting, as long as you compare against undefined instead of
using "in". Is there any particular reason for it that I've missed?

On Nov 22, 2017 13:12, "Richard Davey" <notifications@github.com> wrote:

Why are there so many uses of typeof for an undefined check?

if (typeof this._beforeCallbacks[key] !== 'undefined')

I don't see the need to invoke typeof here, it could simply be:

if (this._beforeCallbacks[key] !== undefined)

Also, I'm dubious about the wholesale replacement of delete for undefined.
They're two totally different things. In the Data Store we use delete
because we want to actually get rid of that key from the object. Not just
blank it, but literally remove it. I agree that mutation is bad for hot
objects, but that doesn't really count the Data Store or any destroy call
(but may include the shaderCache, would have to look further at that one).

—
<https://github.com/photonstorm/phaser/pull/3096#issuecomment-346332862>,
or mute the thread

PR #3101, opened 2017-11-23
Update p2 typings

Nek-:
This PR changes:

* TypeScript Defs

Describe the changes below:

The suggested typings comes from the official p2 typings. (from DefinitelyTyped)
I copy/paste the whole typings because they are not compatible with the phaser version.


Nek-:
closing as done on wrong repository

PR #3103, opened 2017-11-24
Fix the Global Scene Manager by not remembering position for inactive scenes

Twilrom:
* Nothing, it's a bug fix

Describe the changes below:

The global scene manager has a lot of problems centered around trying to keep the active scenes in the same order as they are in the scenes array. I decided to fix this, but I wasn't sure which solution you would prefer, so I made two. Both solutions end up with the active array being a simple array of scenes, like the scenes array.
This PR solves that problem by simply ignoring the order of inactive scenes. A very simple solution, but if a scene is stopped and then re-created it won't be in the same position.


photonstorm:
Closing due to SceneManager rewrite.

PR #3104, opened 2017-11-24
Fix the Global Scene Manager by making sure scenes are moved in both arrays

Twilrom:
* Nothing, it's a bug fix

Describe the changes below:

The global scene manager has a lot of problems centered around trying to keep the active scenes in the same order as they are in the scenes array. I decided to fix this, but I wasn't sure which solution you would prefer, so I made two. Both solutions end up with the active array being a simple array of scenes, like the scenes array.
This PR solves that problem by making sure to always move the scenes in both arrays, and inserting new active scenes according to their position in the scenes array.


Twilrom:
@photonstorm Any interest for these changes? There are currently three pull requests trying to fix #3075, and I'm not sure if you've even noticed them? Do you have other plans for the scene manager?

photonstorm:
Sorry, had always intended to revisit the SceneManager and did so today. Fixes this issue and makes things a lot easier to deal with as a result.

PR #3122, opened 2017-12-12
Upgrade development dependencies.

rblopes:
This PR contains updates to the package lock files and upgrades development dependencies to their latest versions:

- eslint: ^4.3.0 → ^4.13.1
- fs-extra: ^4.0.0 → ^5.0.0
- uglifyjs-webpack-plugin: ^0.4.6 → ^1.1.2
- webpack: ^3.4.1 → ^3.10.0

### DIFF
--- a/v3/package-lock.json
+++ b/v3/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "phaser",
-  "version": "3.0.0-beta.9",
+  "version": "3.0.0-beta.13",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


photonstorm:
Related, I wonder if we ought to gitignore the lock files.

pavle-goloskokovic:
@photonstorm 
`package-lock.json` should be commited: 
https://docs.npmjs.com/files/package-lock.json
```
This file is intended to be committed into source repositories, 
and serves various purposes:

- Describe a single representation of a dependency tree such 
that teammates, deployments, and continuous integration 
are guaranteed to install exactly the same dependencies.

- Provide a facility for users to "time-travel" to previous states 
of node_modules without having to commit the directory itself.

- To facilitate greater visibility of tree changes through readable 
source control diffs.

- And optimize the installation process by allowing npm to skip
 repeated metadata resolutions for previously-installed packages.
```

PR #3125, opened 2017-12-16
Fix deprecated WebAudio value change on FX

orblazer:
This fix replace `.value = volume` for set volume on FX by `.setTargetAtTime()`, because `.value = volume` is deprecated on chrome (https://www.chromestatus.com/features/5287995770929152)

This PR changes :

* Fix Deprecated WebAudio value change on FX

### DIFF
--- a/v3/src/sound/dynamic/FX.js
+++ b/v3/src/sound/dynamic/FX.js
@@ -60,15 +60,15 @@ var FX = new Class({
 
         //  Set the values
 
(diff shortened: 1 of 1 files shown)


samme:
Specifically, it's Chrome's automatic de-zippering of gain changes that's deprecated, not use of the value setter.

samme:
Note <https://github.com/WebAudio/web-audio-api/issues/76#issuecomment-352211472> though.

PR #3157, opened 2018-01-12
Fix: .sub and .len to .subtract and .length

yupaul:
Renamed no longer existing Vector2/3 .sub and .len to .subtract and .length:

.sub and .len have been removed from Math.Vector2 and Math.Vector3, but several files were still using them


photonstorm:
👍

PR #3158, opened 2018-01-12
set data to KeyCombo for KeyComboMatchEvent

### LINKED ISSUE #3155: Phaser 3: KeyComboMatchEvent - not able to determine originating KeyCombo
Hello!

When using the `KEY_COMBO_MATCH_EVENT`, we're unable to determine which `KeyCombo` triggered the event.

Looking at [`KeyComboMatchEvent.js`](https://github.com/photonstorm/phaser/blob/master/src/input/keyboard/combo/KeyComboMatchEvent.js#L14), it assigns the `KeyCombo` to `target` and the `KeyDownEvent` of the last character to the `data` attribute. However, it looks like the `target` attribute is overwritten instead with the `EventDispatcher` instance.

Here is some example code to recreate the scenario (its just the [KeyCombo example](http://labs.phaser.io/edit.html?src=src\input\keyboard\create%20key%20combo.js) with some console logs added):

```js
var config = {
    type: Phaser.CANVAS,
    parent: 'phaser-example',
    scene: {
        create: create
    }
};

function create() {
    //  Here we'll create a simple key combo
    //  When you type in ABCD the event will be triggered on the entry of the final character (D)
    //  An incorrect key press will reset the combo back to the start again

    var combo = this.input.keyboard.createCombo('ABCD');

    this.input.keyboard.events.on('KEY_COMBO_MATCH_EVENT', function (event) {

        console.log('Key Combo matched!');
        console.log('event.data:', event.data);
        console.log('event.target:', event.target);

    });

}
```

Here is the console output from Chrome when the key combo is matched:
image

- The `KeyDownEvent` looks to be just the last letter of the combo.
- I wasn't able to find a reference to the original `KeyCombo` in the `EventDispatcher` instance.

Does it make sense to instead set `data` to the `KeyCombo` itself? Am I missing something obvious?

Happy to send a PR if needed!

Flet:
This PR changes:
* Nothing, it's a bug fix

Describe the changes below:

When using the `KEY_COMBO_MATCH_EVENT`, we're unable to determine which KeyCombo triggered the event.

This PR updates `KeyComboMatchEvent.js` to return the `KeyCombo` reference in the `data` attribute of the event. It also removes the line setting the `target` as it was getting overwritten anyways.

Closes #3155

### DIFF
--- a/src/input/keyboard/combo/KeyComboMatchEvent.js
+++ b/src/input/keyboard/combo/KeyComboMatchEvent.js
@@ -11,9 +11,7 @@ var KeyComboMatchEvent = new Class({
     {
         Event.call(this, 'KEY_COMBO_MATCH_EVENT');
 
-        this.target = keyCombo;
-
-        this.data = keyboardEvent;
+        this.data = keyCombo;
     }
 
 });


photonstorm:
Nice one, thanks.

PR #3164, opened 2018-01-16
Pass event to ProcessKeyCombo

Flet:
This PR changes
* Nothing, it's a bug fix

Describe the changes below:

Thanks for updating the event emitter logic and great writeup in the latest Phaser World! :) It sounds like it will be a nice upgrade.

When updating my game to use the new beta, I noticed a small bug in `ProcessKeyCombo`:
screen shot 2018-01-16 at 11 20 25 am

It looks like `ProcessKeyCombo` was using `event.data` (a holdover from the old event logic I presume).

This PR updates `KeyCombo.js` to pass `event` instead of `event.data` to `ProcessKeyCombo`

Hope this helps! Thanks again!

### DIFF
--- a/src/input/keyboard/combo/KeyCombo.js
+++ b/src/input/keyboard/combo/KeyCombo.js
@@ -90,7 +90,7 @@ var KeyCombo = new Class({
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

PR #3183, opened 2018-01-23
TimerEvent startAt field not scaled

ladrower:
A note regarding the feature `startAt` of `TimerEvent ` 

When an event gets added it should be either scaled by clock `timeScale` as well or not scaled at all - depending on the feature purpose.

```javascript
// Clock preUpdate loop
event = this._pendingInsertion[i];
event.elapsed = event.startAt * event.timeScale
```

In my understanding it should not be influenced by `timeScale` at all.
As a developer I would use the feature of `startAt` to specify the exact offset for my event.

In case I have looped `Timer` for one second and want to start the first loop in the middle, I would set the `startAt` to half a second. And scaling is applied during the timer run time as a factor of speed it elapses.

* Nothing, it's a bug fix

Describe the changes below:

TimerEvent startAt field not scaled

### DIFF
--- a/src/time/Clock.js
+++ b/src/time/Clock.js
@@ -101,8 +101,6 @@ var Clock = new Class({
         {
             event = this._pendingInsertion[i];
 
-            event.elapsed = event.startAt * event.timeScale;
-
             this._active.push(event);
         }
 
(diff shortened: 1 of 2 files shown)


photonstorm:
Agreed, merging this in.

PR #3209, opened 2018-02-09
matterjs PointerConstraint.js world.events.on undefined

vulcanoidlogic:

* Nothing, it's a bug fix

Describe the changes below:

First-ever open source PR - please tell me what I should have done better.

Running the example for physics/matterjs/100 world bodies.js throws an "on is undefined for this.world.events.on" exception.  This change removes the events property from both this.world.events.on and this.world.events.off in PointerConstraint.js.  It looks like World.js extends EventEmitter which seems to expect ".on" instead of "events.on".


photonstorm:
Nice one, thanks.

PR #3214, opened 2018-02-13
Update WebGL Graphics.arc to work more like on Canvas

Twilrom:
I've updated the Graphics WebGL renderer to render arcs closer to how they're rendered by canvas.


Twilrom:
I believe I found the problem. beginPath wasn't resetting the lastPath variable. It should work now.
I'm setting a PI2 variable in the render function now, but I noticed it's defined in math\const.js, so perhaps we should include that and use that instead. I also found some closePath calls that could be removed.

PR #3216, opened 2018-02-13
Fix stray let keywords

agilul:
As Phaser 3 is supposed to be in ES5 those should be `var`s.
Noticed when using `webpack -p` and it complained about them 😋


samme:
#3220, #3221

Issue #3228, opened 2018-02-14
Feature: Plugins via import


nkholski:
It would be nice to add plugins in the package.json and import them into the scenes. Like this:

`import BasePlugin from 'base-plugin;`

```
preload(){
    this.load.plugin('BasePlugin', BasePlugin); // Object instead of url
}
```

Actually the above works if this is added to PluginFile in PluginFile.js but will throw a 404 in the console:
```
        if(typeof url !== "function"){
            this.key = key;
            window[key] = url;
            window[key].register(PluginManager);
        }
```

photonstorm:
Agreed, have marked for 3.0.2. If you want to submit a PR I'd merge it too, but we'll do this regardless :)

photonstorm:
Thanks for opening this issue, and for submitting a PR to fix it. We have merged your PR into the `master` branch and attributed the work to you in the [Change Log](https://github.com/photonstorm/phaser/blob/master/CHANGELOG.md). If you need to tweak the code for whatever reason please submit a new PR.

PR #3234, opened 2018-02-14
load.plugin now accepts class as an argument besides an url-string

nkholski:
Support for importing and adding plugin classes without an url to be loaded as discussed in #3228.

This approach is tested and works fine, however, I don't know if you find it too ugly. Basically, it free rides on the ordinary loader but skips making a file request. With this the url-parameter passed to load.plugin might not be an url but the plugin itself.

### DIFF
--- a/src/loader/File.js
+++ b/src/loader/File.js
@@ -65,7 +65,7 @@ var File = new Class({
         {
             this.url = GetFastValue(fileConfig, 'path', '') + this.key + '.' + GetFastValue(fileConfig, 'extension', '');
         }
-        else if(typeof(this.url) !== "function")
+        else if (typeof(this.url) !== 'function')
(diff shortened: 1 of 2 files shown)


photonstorm:
A couple tiny formatting issues but otherwise I'm happy with this! Will merge for 3.0.2.

nkholski:
Cool. I'm unable to see the formatting issues. If you give me a hint I could fix them, or should I just leave it for you?

rblopes:
@nkholski usually, the parenthesis of a `if` statement should have a space before it and blocks begin on a new line.

```js
if (/* condition */)
{
  /* ... */
}
```

photonstorm:
Can now get this in 3.2.0 - thank you :)

PR #3240, opened 2018-02-16
Update WebGL Graphics.arc to work more like on Canvas

Twilrom:
Resubmitting this as I broke the other one. Fixed the problems with circle rendering, and removed some unneeded calls.

### DIFF
--- a/src/gameobjects/graphics/Graphics.js
+++ b/src/gameobjects/graphics/Graphics.js
@@ -350,7 +350,6 @@ var Graphics = new Class({
     {
         this.beginPath();
         this.arc(x, y, radius, 0, MATH_CONST.PI2);
-        this.closePath();
         this.fillPath();
 
         return this;
@@ -372,7 +371,6 @@ var Graphics = new Class({
     {
         this.beginPath();
         this.arc(x, y, radius, 0, MATH_CONST.PI2);
-        this.closePath();
         this.strokePath();
 
         return this;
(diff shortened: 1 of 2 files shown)


Twilrom:
While poking around in the Graphics renderer I've noticed that the paths work very differently to how paths in the regular canvas works. This could lead to very inconsistent rendering, but allows for some fancy effects. Perhaps these "fancy lines" should be broken out into a separate class, so the WebGL renderer can be made to mimic the canvas as closely as possible instead?

I've also been wondering about some eslint rules, like for example "eol-last" and "space-before-function-paren". All the code seems to do the exact opposite of what they're set to enforce. Do you not use eslint?

photonstorm:
Having just finished running eslint on the whole codebase, `eol-last` was respected everywhere bar one file. The most common issue was redeclared vars, but they're all fixed now. The entire codebase passes our config (3rd party libs excluded)

Twilrom:
But eslint is configured to disallow ending newlines, and spaces before function parenthesis, and all files end in newline and have spaces before function parenthesis? I get an error for eol-last for every file, and space-before-function-paren for every function. Perhaps something changed in how eslint interprets the rules?

photonstorm:
Our config is set to error if a newline is missing:

`"eol-last": [ "error" ]`

Which it does (providing you run eslint using our config, although pretty sure the default for [eol-last](https://eslint.org/docs/rules/eol-last) it to warn anyway.) Change `error` to `off` or `warn` and it'll allow them through.

Twilrom:
Oh, I'm terribly sorry. It turns out eslint was reading unrelated rules from other directories, and that's what was confusing it. You might want to add "root: true" to your eslint file to prevent that, so no one else has to look stupid.

photonstorm:
Good idea, added.

kefniark:
@photonstorm not related to this PR directly but there are lot of comments latelly about linting.
It could be nice to setup some CI (like travis) to automatically run `npm run lint` on each PR.

Twilrom:
Any problems preventing this from being merged?

PR #3254, opened 2018-02-18
Bugfixes in Structs.Set

yupaul:
Fixed errors in Structs.Set - entries instead of values and .set instead of .add


smartsingam:
Yupaul can u mail me.. I have doubts on phaser development ...my mail Id- singaraj671@gmail.com

photonstorm:
Thanks, have merged and added to the change log.

PR #3258, opened 2018-02-19
Fix camera jitter

### LINKED ISSUE #3245: Camera Jitter
After updating to 3.0.0 release or above, camera follow now seems to have a random 1-pixel jitter that affects everything on the screen, like some kind of off-by-one or rounding error. This was never an issue in beta 20 or earlier.

This is my game config:
```javascript
{
      parent,
      type: Phaser.WEBGL,
      width: 400,
      height: 225,
      pixelArt: true,
      physics: {
        // @todo: Switch to 'matter' so we can use sloped tiles
        default: 'arcade',
        arcade: {
          // debug: true,
          gravity: {
            y: 450
          }
        }
      },
      scene: new RootScene(projectPath)
    }
```

In my create() function I have this call:
```javascript
this.cameras.main.setRoundPixels(true);
```

It may be worth noting that the jitter happens on both the X and the Y axis even though the camera boundary height is the same as the screen resolution height.

If I call setRoundPixels(false) instead then the issue goes away, BUT I get slight corruption in the edges of tiles being rendered plus a warbling/underwater effect due to pixel columns being slightly different sizes.

bradharms:
This fixes #3245.

This removes certain if() blocks added in 7b1ad0b in order to fix camera
jitter that occurs when both `roundPixels` is enabled and sprite
following is enabled.

**NOTE:** It's not clear to me what the if() blocks removed had been doing,
but the associated commit message suggested they were trying to make
`roundPixels` actually do something, as if that flag weren't having any effect.
However prior to that commit `roundPixels` seemed to be already working
normally. If the removed blocks actually were resolving a different
problem, then camera jitter will have to be resolved differently than
just removing those blocks.

### DIFF
--- a/src/renderer/webgl/pipelines/FlatTintPipeline.js
+++ b/src/renderer/webgl/pipelines/FlatTintPipeline.js
@@ -168,7 +168,7 @@ var FlatTintPipeline = new Class({
     {
         WebGLPipeline.prototype.resize.call(this, width, height, resolution);
         this.projOrtho(0, this.width, this.height, 0, -1000.0, 1000.0);
-        
+
         return this;
     },
 
@@ -206,7 +206,7 @@ var FlatTintPipeline = new Class({
         {
             this.flush();
         }
-        
+
         var renderer = this.renderer;
         var resolution = renderer.config.resolution;
         var vertexViewF32 = this.vertexViewF32;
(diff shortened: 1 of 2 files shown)


photonstorm:
Thank you for this PR. We have fixed this issue independently via some other work we did in the renderer, so I'm afraid we cannot merge this, but hopefully, you'll see the issue is resolved all the same.

bradharms:
Thank you, I had hoped you would. Do you mind my asking how it was fixed or which commits I would look at to understand how this works?

photonstorm:
It's split over a few, but this is the main one https://github.com/photonstorm/phaser/commit/50c79c14af5ba3124058e59ef23831a977341f86 (which you'll see is similar to your PR) and look at the other commits around the same time from bitnenfer

PR #3259, opened 2018-02-19
Use webpack-shell-plugin `onBuildExit` event

rblopes:
With this change, `webpack watch` will update the build in the `phaser3-examples` repository too, making it easier to test quick changes in the examples repository.

([Reference](https://github.com/1337programming/webpack-shell-plugin#api).)


photonstorm:
Nice one.

Issue #3277, opened 2018-02-23
Collision with groups


nkholski:
Arcade physics already support passing an array for collision checks, but not groups. My current solution to check a sprite against a group with sprites looks like this:
```
this.scene.enemyGroup.children.entries.forEach(enemy => {
      this.scene.physics.world.overlap(this, enemy, ()=>{enemy.kill()});
});
```
If I could pass groups it would look like this:
`this.scene.physics.world.overlap(this, this.scene.enemyGroup,(me,enemy)=>{enemy.kill()});`

If I'm successful with the git-stuff I will make a PR within a few minutes.

photonstorm:
Strange, the Making your first game tutorial I wrote uses sprite vs. Group and group vs. group and it works for overlap and collide. What’s the difference between how I do it and your approach?

nkholski:
You add 
`platforms = this.physics.add.staticGroup();`
And I just add a group
`enemyGroup = this.add.group();`
I never seen staticGroup before. Maybe that's the correct way to do it? On the other hand adding support for collision with normal groups just add two lines of code.

photonstorm:
A Physics Group is basically just a Group with a couple helper methods. The important part is this: https://github.com/photonstorm/phaser/blob/master/src/physics/arcade/World.js#L1395-L1409 which should work on any Group, because they all set the 'isParent' property.

nkholski:
Ok. I digged a little bit more. Everything works fine with isParent but something goes wrong becase the group has no physics:
https://github.com/photonstorm/phaser/blob/master/src/physics/arcade/World.js#L1493
I updated my PR to target groups without physics only. IDK if that's the right approach, but with this PR a normal group will be handled just like an object containing an array of collidable objects so the collision handler could just focus on the array.

photonstorm:
Thanks for opening this issue, and for submitting a PR to fix it. We have merged your PR into the `master` branch and attributed the work to you in the [Change Log](https://github.com/photonstorm/phaser/blob/master/CHANGELOG.md). If you need to tweak the code for whatever reason please submit a new PR.

PR #3278, opened 2018-02-23
Groups arcade collision

nkholski:
As described in issue #3277


photonstorm:
Thanks for tidying this up. Merging for 3.2.

Issue #3239, opened 2018-02-16
`Key#duration` not behaving as documented


rblopes:
Using the keyboard input manager API, it is not possible to evaluate how long a given key is being pressed down.

<details>
<summary>Sample code (<a href="https://jsfiddle.net/rblopes/hmtxj5hb/">live</a>) (Tested with Phaser 3.1.0.)</summary>

```js
const game = new Phaser.Game({
  type: Phaser.CANVAS,
  width: 400,
  height: 300,
  backgroundColor: '#222288',
  scene: {
    create() {
    	this.keys = this.input.keyboard.addKeys({
      	leftKey: Phaser.Input.Keyboard.KeyCodes.LEFT,
        rightKey: Phaser.Input.Keyboard.KeyCodes.RIGHT
      });

      this.t1 = this.add.text(16, 16);
      this.t2 = this.add.text(16, 64);
    },
    
    update() {
      const {leftKey, rightKey} = this.keys;
      this.t1.setText([
        `leftKey.isDown = ${leftKey.isDown}`,
        `leftKey.duration = ${leftKey.duration}`
      ].join('\n'));
      this.t2.setText([
        `rightKey.isDown = ${rightKey.isDown}`,
        `rightKey.duration = ${rightKey.duration}`
      ].join('\n'));
    }
  }
});
```
</details>

<details>
<summary>Screenshots</summary>

![image](https://user-images.githubusercontent.com/61425/36316254-f81c3da4-1320-11e8-9c52-aeb8d72f15c0.png)
![image](https://user-images.githubusercontent.com/61425/36316316-27c278a2-1321-11e8-9cf5-af1da663e9d6.png)
</details>

#### Expected behaviour
The `duration` property should update while a key is held down, preserve the last value when that key is held up, as documented.

#### Actual result
While a key is held down, `duration` is always zero. When that key is released, `duration` yields a random value.

agilul:
Seems like it's firing `ProcessKeyDown` continously after ~500ms, updating `key.timeDown` in the meanwhile.
I guess it's the default Operating System behaviour with input management, like when you keep a key pressed to type it many times.

Could check in `ProcessKeyDown` if the key was already pressed and prevent overwriting `key.timeDown`.

photonstorm:
[...]

PR #3279, opened 2018-02-24
Fix KeyboardManager keydown events keeps firing

squaresun:
Keydown event is fired repeatedly when holding down keys.
**This PR fixes this bug by changing ```Phaser.Input.Keyboard.KeyboardManager#update```:**
1. Global ```keydown``` and ```keydown_KEYCODE``` event is fired only when ```keys[code].isDown === false || key[code] === undefined```.
> About the condition ```key[code] === undefined ```, since I observed there is no other way to get the previous ```isDown``` state of keys that are not added in KeyboardManager, so global events will still be fired repeatedly during key holding down which isn't added in KeyboardManager.
2. ProcessKeyDown is fired only when ```keys[code].isDown === false``` (#3239 )

I don't confirmed if there is another better solution🤔

Thanks @agilul for the advice of this issue.

### DIFF
--- a/src/input/keyboard/KeyboardManager.js
+++ b/src/input/keyboard/KeyboardManager.js
@@ -362,23 +362,26 @@ var KeyboardManager = new Class({
             var event = queue[i];
             var code = event.keyCode;
 
-            //  Will emit a keyboard or keyup event
-            this.emit(event.type, event);
-
             if (event.type === 'keydown')
             {
-                if (KeyMap[code])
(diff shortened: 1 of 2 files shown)


rexrainbow:
But `key.repeats` won't be increasing if keydown events won't keep firing.

squaresun:
Updated.

Function ```Phaser.Input.Keyboard.Keys.ProcessKeyDown``` will still be fired for increasing key.repeats

Thanks for reminding me😀 @rexrainbow

PR #3287, opened 2018-02-27
BugFix for ScenePlugin.launch, data now passed to queued scenes

gaudeon:
This PR changes

* Nothing, it's a bug fix

Fixed the scene plugin launch function so that data is passed to scenes queued for processing when they are not yet running.


photonstorm:
Lovely, thanks.

PR #3292, opened 2018-02-27
Support MultiAtlas with TexturePackers "Phaser 3" atlas format 

CodeAndWeb:
TexturePacker's new JSON atlas format "Phaser 3" exports one combined atlas file for all image files when multi-pack (MultiAtlas) is used.
This change enables Phaser to load the textures with the combined atlas file:

`this.load.multiatlas('sheet', ['sheet-0.png', 'sheet-1.png'], 'sheet.json');`

### DIFF
--- a/src/loader/LoaderPlugin.js
+++ b/src/loader/LoaderPlugin.js
@@ -656,7 +656,7 @@ var LoaderPlugin = new Class({
                 //  Yup, add them to the Texture Manager
 
                 //  Is the data JSON Hash or JSON Array?
-                if (Array.isArray(data[0].frames))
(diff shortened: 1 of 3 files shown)


photonstorm:
Cheers - this will do nicely until we support it properly (i.e. loading just the json on its own, which then populates the loader with extra files)

CodeAndWeb:
Thanks for merging!

Proper support for it welcome ;-)
It's quite useful if you don't have to care about the number of sprite sheets to load in multipack.
Adding normal maps from the atlas is also still a bit "rough" - see https://www.codeandweb.com/blog/2018/03/01/how-to-create-light-effects-in-phaser3

PR #3301, opened 2018-03-01
Fix RandomDataGenerator's randomness

### LINKED ISSUE #3289: RandomDataGenerator.frac is not so random
* A bug in the API v3.1.2

The difference between first two numbers returned by RandomDataGenerator.frac
is always almost-equal even if the seed is different.

    Phaser.Math.RND.sow([Date.now()]);
    console.log(Phaser.Math.RND.frac() - Phaser.Math.RND.frac()); // always about -0.056

EDIT: The original implementation of Alea is different from that of Phaser in the following points.

1) In [hashing](https://github.com/coverslide/node-alea/blob/master/alea.js#L89) function, Alea caches "n" but Phaser doesn't.
2) In [sowing](https://github.com/coverslide/node-alea/blob/master/alea.js#L37) function, Alea initializes (s0, s1, s2) using the same string, but Phaser doesn't.

It seems (1) causes such regularity. (2) is not essential.

migiyubi:
Fix #3289

### DIFF
--- a/src/math/random-data-generator/RandomDataGenerator.js
+++ b/src/math/random-data-generator/RandomDataGenerator.js
@@ -68,6 +68,17 @@ var RandomDataGenerator = new Class({
         this.s2 = 0;
 
         /**
+         * Internal var.
+         *
+         * @name Phaser.Math.RandomDataGenerator#n
+         * @type {number}
+         * @default 0
+         * @private
+         * @since 3.0.0
+         */
+        this.n = 0;
+
+        /**
          * [description]
          *
          * @name Phaser.Math.RandomDataGenerator#sign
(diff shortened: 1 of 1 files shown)


src/math/random-data-generator/RandomDataGenerator.js
```diff
@@ -67,6 +67,17 @@ var RandomDataGenerator = new Class({
...
         this.s2 = 0;
 
+        /**
+         * Internal var.
+         *
+         * @name Phaser.Math.RandomDataGenerator#n
+         * @type {number}
+         * @default 0
+         * @private
+         * @since 3.0.0
```
photonstorm: 3.2.0 please :)

photonstorm:
Nice one, thanks.

Issue #3309, opened 2018-03-03
HTML5 audio doesn't work in safari (on latest MacOS)


alexandernst:
For example http://labs.phaser.io/view.html?src=src\audio\HTML5%20Audio\AudioSprite.js

photonstorm:
With the update merged in I'd love to know what errors you now get @alexandernst

alexandernst:
```
NotAllowedError (DOM Exception 35): The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.  — HTML5AudioSound.js:346
playCatchPromise — HTML5AudioSound.js:340
pickAndPlayAudioTag — HTML5AudioSound.js:256
play — HTML5AudioSound.js:113
playAudioSprite — BaseSoundManager.js:258
(función anónima) — AudioSprite.js:67
emit — index.js:182
processDownEvents — InputPlugin.js:506
update — InputPlugin.js:1510
emit — index.js:203
step — Systems.js:278
update — SceneManager.js:478
step — Game.js:363
step
step — TimeStep.js:519
step
step — RequestAnimationFrame.js:102
```

Note that I was never asked about allowing (or denying) anything.

sftsk:
It seems like the same is happening on iOS when switching scenes and trying to play audio. The first audio you play works, but the next one will fail (silently) with this error.

sftsk:
@photonstorm any progress looking into this?

sftsk:
Alright, just removed

```
audio: {
    disableWebAudio: true
  },
```
from the game config and now audio works just fine on iOS/ Safari on macOS.
Not sure why I thought HTML5 audio would be the way to go but this might happen to others as well. Perhaps tell in the examples it's advised to use Web Audio.

Cheers!

photonstorm:
I’ll keep this open because old school Audio should still work on Safari, but you absolutely don’t want to use it unless forced to.

pavle-goloskokovic:
@photonstorm I have a fix for this but can't test it since all HTML5 Audio examples are broken with following error:
```
Uncaught RangeError: Maximum call stack size exceeded
    at HTML5AudioSound.set [as mute] (HTML5AudioSound.js:627)
    at HTML5AudioSound.setMute (HTML5AudioSound.js:655)
    at HTML5AudioSound.set [as mute] (HTML5AudioSound.js:636)
    at HTML5AudioSound.setMute (HTML5AudioSound.js:655)
[...]

PR #3313, opened 2018-03-03
Warn if HTML5AudioSound#audio fails to play

samme:
Otherwise it's hard to tell what went wrong (#3309).

This PR changes

* The public-facing API


photonstorm:
I'm in two minds about adding in warns like this, but I guess we an always revert it if people complain :)

PR #3327, opened 2018-03-05
Fix for collideGroupVsSelf is undefined

patrickgalbraith:
**This PR changes (delete as applicable)**

* Nothing, it's a bug fix

This happens when you call collide with a single group as the first param.

Seems like the `collideGroupVsSelf` function was not implemented. I've changed it to just call the `collideGroupVsGroup` with the first object twice which worked when I tested it on my project.


photonstorm:
Ahhh, the joys of porting legacy code! Thanks for the catch.

Issue #3231, opened 2018-02-14
Phaser.CANVAS doesn't rotate tilesprites


amkazan:
* A bug in the API (always say which version you're using!)
Phaser 3.0.0

http://jsbin.com/mevarapizu/edit?html,js,output

In Canvas mode tilesprites are not rotated. They are if we change CANVAS to WEBGL

photonstorm:
Well, that's embarrassing :) We're on it, will be fixed in 3.0.1.

photonstorm:
Extra note - it doesn't scale them either.

amkazan:
cool! Found it accidentally while trying to understand fast fan spinning when 100 sprites are added (half rotated)

(p.s. By the way: Thanks for the great framework!!! 👍 )

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

PR #3337, opened 2018-03-06
Added rotation, flipping and scaling to TileSpriteCanvasRenderer

TCatshoek:

- Nothing, it's a bug fix

Describe the changes below:

Fix for [#3231](https://github.com/photonstorm/phaser/issues/3231):

The tilesprite CANVAS renderer did not support rotation and scaling. Furthermore we discovered that it did not support flipping. This PR fixes the rotation and the flipping. Scaling of the tilesprite also works now, but the texture still does not scale with it if the texture is larger than the area being drawn. We are planning to create a new issue for this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Change Log
 
+## Version 3.2.1 - In Development
+
+### New Features
+
+### Bug Fixes
+
+### Updates
+
+
+
 ## Version 3.2.0 - Kaori - 5th March 2018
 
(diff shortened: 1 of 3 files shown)


TCatshoek:
[Jsfiddle](https://jsfiddle.net/goL3t8q6/13/) demonstrating rotation and texture scrolling (adapted from phaser3 examples).

TCatshoek:
Sorry, made a mistake here. I'm opening a new pull request.

PR #3344, opened 2018-03-07
Fix Phaser.Utils.String.Format

samme:

* Nothing, it's a bug fix


iamchristopher:
Curious why you didn't change to `return string.replace(/* .. */)` instead of assigning back to `string`?

samme:
Just lazy.

iamchristopher:
I can respect laziness! But thanks for considering my nitpicky comment.

Issue #3353, opened 2018-03-08
Camera zoom + startFollow combine in unexpected ways


jdotrjs:
Shoutout to discord member PF_Cactus for this one.

They found a bug in camera zoom + `startFollow` where camera movements seem to be exaggerated relative to the sprite being followed.

Repro gif:

![zoomstartfollow](https://user-images.githubusercontent.com/37101247/37180786-2d39ed1a-22df-11e8-9ec8-57c6c3cd30de.gif)

Code (just a modified version of [this labs example](http://labs.phaser.io/edit.html?src=src/camera/follow%20sprite.js):

<details>

```javascript
var config = {
    type: Phaser.WEBGL,
    parent: 'phaser-example',
    scene: {
        preload: preload,
        create: create,
        update: update
    },
    width: 800,
    height: 600
};

var clown;
var image;
var iter = 3.14;
var game = new Phaser.Game(config);

function preload() {

    this.load.image('CherilPerils', 'assets/tests/camera/CherilPerils.png');
    this.load.image('clown', 'assets/sprites/clown.png');

}

function create() {

    image = this.add.image(0, 0, 'CherilPerils');
    clown = this.add.image(500, 620, 'clown');
    image.originX = 0;
    image.originY = 0;
    this.cameras.main.setSize(400, 300);
    this.cameras.add(400, 0, 400, 300);
    this.cameras.add(0, 300, 400, 300);
    this.cameras.add(400, 300, 400, 300);
    this.cameras.main.startFollow(clown).setZoom(2);
}

function update()
{
    clown.x = 250 + Math.cos(iter) * 200;
    clown.y = 310 + Math.sin(iter) * 200;
    iter += 0.02;
}
```
</details>

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

PR #3354, opened 2018-03-08
fixed camera offset when combining zoom+follow

brandonvdongen:
This PR changes

* Nothing, it's a bug fix

Describe the changes below:

fixes bug mentioned in: #3353 

Camera scroll did not account for zoom level when following a sprite causing it to gain an offset
this change fixed it based on my tests so far.

```
this.scrollX = (originX - width * 0.5) / zoom;
this.scrollY = (originY - height * 0.5) / zoom;
```

### DIFF
--- a/src/cameras/2d/Camera.js
+++ b/src/cameras/2d/Camera.js
@@ -852,8 +852,8 @@ var Camera = new Class({
             originX = follow.x;
             originY = follow.y;
 
-            this.scrollX = originX - width * 0.5;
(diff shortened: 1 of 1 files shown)


samid737:
Seems to work even with A zero case , setZoom(0). Note that the zero case will have A infinite scrollX:

https://jsfiddle.net/7eotsk1x/

photonstorm:
👍

Issue #3338, opened 2018-03-06
Texture scaling differs between WebGL and canvas TileSpriteRenderer


TCatshoek:
* A bug in the API (v3.2.0)

I noticed a difference between the scale textures are drawn between the webgl and canvas tilesprite renderers.

[Example here](https://jsfiddle.net/5ozxfh1v/16/)

This should draw the given image with both renderers, and I would expect them to draw the exact same thing here, but they draw the texture at different scales.

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

TCatshoek:
[Jsfiddle](https://jsfiddle.net/5ozxfh1v/19/) using a fresh build of the master branch. Seems like it's working perfectly now!

PR #3358, opened 2018-03-09
Added rotation, scaling and flipping to TileSpriteCanvasRenderer

TCatshoek:
This PR changes

* Nothing, it's a bug fix

Describe the changes below:

Fix for #3231:

The tilesprite CANVAS renderer did not support rotation and scaling. Furthermore we discovered that it did not support flipping. This PR fixes the rotation and the flipping. Scaling of the tilesprite also works now, but we noticed there are still some texture scaling issues compared to webgl (see #3338).

[Jsfiddle](https://jsfiddle.net/goL3t8q6/20/) demonstrating rotation, scaling and flipping on the canvas renderer (adapted from phaser3 examples).


photonstorm:
Lovely, thank you.

PR #3365, opened 2018-03-10
Add Phaser.GameObjects.TileSprite#setTilePosition

samme:
Chainable setter method.

This PR changes

* The public-facing API


photonstorm:
Yeah looks useful. Let's save this for 3.3.0 as it can't go in 3.2.1.

Issue #3371, opened 2018-03-12
Input area is off when resolution !== 1


sftsk:
This Issue is about (delete as applicable)

* A bug in the API 3.2.0 (always say which version you're using!)

I am setting up a basic game with `resolution` set to the device pixel ratio. Then I am adding a clickable sprite which is scaled down to half the size. The click area and position does not match the displayed sprite, it seems to be ignoring the resolution.

```
var config = {
  type: Phaser.AUTO,
  parent: 'app',
  width: window.innerWidth,
  height: window.innerHeight,
  resolution: window.devicePixelRatio,
  autoResize: true,
  scene: [ menuScene ],
  audio: {
    disableWebAudio: true
  },
};

var game = new Phaser.Game(config);

class menuScene extends Phaser.Scene {

  constructor () {
    super({
      key: 'menuScene'
    });
  }

  preload () {
    this.load.atlas('menuSprites', 'assets/images/menu/menu.png', 'assets/images/menu/menu.json');
    this.load.image('menuBackground', 'assets/images/menu/intro_blank.jpg');
  }

  create () {
    var _this = this;
    textures.menu = {}; // reference to the textures
    hideTextBox();
    this.scene.launch('UIScene');
    this.events.on('resize', this.resize); // trigger scene resize

    textures.menu.bg = this.add.sprite(view.width / 2, view.height / 2, 'menuBackground').setScale(0.5);
    textures.menu.danny = this.add.sprite(Phaser.Math.Between(0, 300), Phaser.Math.Between(0, 300), 'menuSprites', 'Danny.png').setScale(0.5).setInteractive().setName('dannyScene');


    this.input.on('gameobjectdown', function (pointer, gameObject) {
      _this.scene.stop('UIScene');
      _this.scene.start(gameObject.name);
    });
  }

}
```

Might be that I am missing something like `.syncGameObject()` but somehow I would expect this to work out of the box.

Thanks for any help!

sftsk:
Did some more investigating. All the coordinates are off by whatever the multiple of the resolution is.
If set to `0.5` the clickable spots are half of the actual coordinates, if set to `1.5` they are at 150% of where they should be.

Could it have to do with the canvas width/ height getting set twice so somehow the click goes to the smaller/ larger canvas size?

`<canvas width="555" height="416" style="width: 1111px; height: 833px;"></canvas>`

sftsk:
I have added a pull request. It fixes the issue :)
[...]

PR #3376, opened 2018-03-12
fix offset in hitTest when resolution !== 1

sftsk:

* Nothing, it's a bug fix
Fixes issue #3371

Describe the changes below:
Hit area was off when setting resolution to anything but 1.
I have multiplied the hit point with the resolution to make sure it adds up.

### DIFF
--- a/src/input/InputManager.js
+++ b/src/input/InputManager.js
@@ -390,8 +390,8 @@ var InputManager = new Class({
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


photonstorm:
I'm happy with this fix, just not how it's implemented. The resolution never changes, so we can cache it in a local var before the for loop, to avoid the config object look-up for every object.

PR #3377, opened 2018-03-12
set default gamepad button threshold to 1

jmcriat:
this.threshold = 0  was making all gamepad buttons to be always pressed down (this.pressed = true). 
this.threshold = 1 will correct the issue and in the case of analogue buttons it will only emit the event when the 'pressure' is at max.

* Nothing, it's a bug fix


photonstorm:
Thanks. The jsdoc `default` value needs updating too, but I'll do that.

Issue #3378, opened 2018-03-12
Exteneded GameObject (Sprite) not fire update method, but preUpdate 


hadikcz:
It's little bit wierd but all extened Sprites in V2 fires update method in class, but right now update not working. I discovered in UpdateList that is preUpdate is fired. But I think this solution is not clean as should be and can avoid some backend process.

Look at `TestSprite.update` and `TestSprite.preUpdate`

https://jsfiddle.net/athntebo/19/

samme:
It looks like you just need to add a preUpdate method (even empty).

<https://github.com/photonstorm/phaser/blob/22901a1b05abf5c37ff1e6fd0c4c7974d90d6017/src/gameobjects/GameObjectFactory.js#L95-L101>

kalebwalton:
@photonstorm Why is this complex/difficult? I just ran into this myself as the Sprite docs even state that 'update' is to be overridden.

kalebwalton:
I'm surely too naive about this... but I'll ask anyway. Why can't we just add 'update' call after this line:

https://github.com/photonstorm/phaser/blob/1be486fab4177a25186117583e65c55ab5052434/src/gameobjects/UpdateList.js#L196

samme:
See <https://github.com/photonstorm/phaser/pull/3379#issuecomment-373718957>.

kalebwalton:
Doing     this.scene.on('update', this.update) for now. Thank you for the reference @samme !

photonstorm:
It’s flagged as complex because it requires a significant refactor to the internals, the creation of a new plugin and changes to how Groups work. To do this right touches lots of elements.

samme:
I made <https://github.com/samme/phaser-plugin-update>.

photonstorm:
Going to close this issue. These are plugins available to address this, and it's just the way v3 works. Will re-consider for v4.

Krankomat:
> Doing this.scene.on('update', this.update) for now. Thank you for the reference @samme !

@kalebwalton This did not work for me. However, this works (Phaser v3.21.0): 
```js
this.scene.events.on('update', (time, delta) => { this.update(time, delta)} );
```
I wrote a class that extends `Phaser.GameObjects.Container`, put this into the `constructor` and voila, it works.

juanitogan:
[...]

PR #3379, opened 2018-03-12
The update method for game objects should be called after the preUpdate.

wtravO:

* Nothing, it's a bug fix

Closes Issue #3378 

Describe the changes below:
For extended game objects, the update method is never fired unless the preUpdate is overridden with an update call.  This will ensure it is properly called directly after the preUpdate call in the GameObject.UpdateList.


wtravO:
@photonstorm Thoughts on this?

photonstorm:
This will conflict with `Group.preUpdate`, which calls the `update` of any children automatically. So some objects could potentially have their `update` called twice per step. I can see what this is trying to do, but I think we need to find a slightly different way to do it.

wtravO:
I see.  Is it necessary that the group call the update methods for its children? As long as all of the preUpdates are ran before any of the updates, it should have the same effect.  The `Group.preUpdate` would simply do the prep and `GameObject.update` would take that prep and apply it automatically.

photonstorm:
As it stands, the problem will be the execution order.  `UpdateList` runs early on, as it's one of the core scene plugins. Clock, Input, Loader and Tweens all update after it. Most the time you'd want `GameObject.update` called _after_ all of the plugins have run their updates. And in some cases, it would be useful to have the GameObjects update before that happened.

More importantly though, I don't want to add every single Game Object to the UpdateList. As it will have to iterate them all, every frame, and in most cases the majority of them won't actually do anything.

I'm happy to remove the `update` call from Group, but we still need a slightly different way to handle it, because I'm not willing to call `update` on every single Game Object created. The Scene emits an `update` event which could be hooked in to. Or we could add a method (or an argument to `setActive`) that inserts it into an UpdateList (and we rename UpdateList to PreUpdateList, as that is actually what it is)

wtravO:
I like the idea of adding an argument to `setActive`.  Seems to be the most unobtrusive way of handling this.  I'll work on that now and revise my PR.  Then we can go from there.

wtravO:
My thoughts are that the `Phaser.GameObjects.UpdateList` is in need of a refactor - considering it does not do any actual updating.  For clarity sake, I'll be referring to the `Phaser.GameObjects.UpdateList` as the `PreUpdateList` henceforth.  

Currently, when `GameObject.setActive(true);` is called, it simply adds a `GameObject` to the `PreUpdateList`.  Perhaps we could add an additional parameter to `GameObject.setActive` that would also inject the `GameObject` into the scenes `UpdateList`.  Just set this parameter to false by default to avoid confusion and over-injection.

wtravO:
This commit was closed due to upcoming alternative approaches to this issue.

PR #3381, opened 2018-03-13
Fix debug draw of scaled arcade body

pixelpicosean:
bug fix: debug draw of arcade circle body ignores scale of its parent game object


samid737:
What if the intention is to scale the circle according to height instead of width? i.e:

http://jsfiddle.net/237k97bz/3/

pixelpicosean:
@samid737 do you mean let the user to manually change radius based on scale?

PR #3406, opened 2018-03-16
Technical debt fix 1

Tomas2h:
We ran SonarCloud on the phaser repo and it scored really good, but there were still about 100 issues found (bugs, code smells and vulnerabilities). With this pr we fix only 4 findings, but we will try to come up with some more following weeks.

* Nothing, it's a bug fix 

Describe the changes below:
In the BlendMode class there was a useless statement. In RotateMatrix we replaced these self-assignments replaced by the classes made for these.

### DIFF
--- a/src/gameobjects/components/BlendMode.js
+++ b/src/gameobjects/components/BlendMode.js
@@ -9,7 +9,7 @@ var BlendModes = require('../../renderer/BlendModes');
 /**
  * Provides methods used for setting the blend mode of a Game Object.
  * Should be applied as a mixin and not used directly.
- * 
+ *
  * @name Phaser.GameObjects.Components.BlendMode
  * @since 3.0.0
  */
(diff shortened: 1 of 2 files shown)


src/gameobjects/components/BlendMode.js
```diff
@@ -56,8 +56,6 @@ var BlendMode = {
                 value = BlendModes[value];
             }
 
-            value | 0;
```
photonstorm: This shouldn't be removed, it should be fixed: `value |= 0;`

photonstorm:
I'm definitely interested in any bugs that may have been found during automated testing, but there aren't any _bug_ fixes in this PR. The one change that could have been a fix  was removed. And the other changes seem to just trade a native `Array.reverse` call for a branch out to another function.

As it stands I'm not too sure what to make of it. I've gone ahead and fixed the blendMode value issue, and on the plus side, this PR made me look at the Matrix functions, which highlighted the fact that `RotateLeft` was actually rotating to the right, and vice versa. Plus the functions weren't even in the build file because they weren't part of the Utils namespace. Definitely useful, but perhaps not in the way you were hoping.

If you've other things you have uncovered I'd be happy to see them, as long as they do actually fix something please.

PR #3409, opened 2018-03-17
Remove scene key in SceneManager

macbury:
When we remove scene we did set its key value to undefined. Because of that, when i would like to create a new scene with the same key(after removing the old scene), the `SceneManager#createSceneFromFunction` would throw exception `Cannot add a Scene with duplicate key`.

This fix deletes key from hash.


photonstorm:
👍

Issue #3389, opened 2018-03-14
Result of Text.getWrappedText() is different from Text.updateText()


rexrainbow:
* A bug in the API (v3.2.1)

I tried to get wrapped lines in text object. And found that:

Result (**Text.runWordWrap(text)**) of [Text.getWrappedText(text)](https://github.com/photonstorm/phaser/blob/master/src/gameobjects/text/static/Text.js#L489) is different from[Text.updateText()](https://github.com/photonstorm/phaser/blob/master/src/gameobjects/text/static/Text.js#L926)

It might because that [the font is not sync to canvas](https://github.com/photonstorm/phaser/blob/master/src/gameobjects/text/static/Text.js#L920) in Text.getWrappedText()
In my testing, font size is
- Text.getWrappedText: "12px sans-serif"
- Text.updateText: "16px Courier"

Should I call `style.syncFont(canvas, context)` in my code?

rexrainbow:
Here is another problem of Text.getWrappedText() -- [sample code](https://codepen.io/rexrainbow/pen/xWZeNW?editors=0010)

I try to display text page by page.
So I get wrapped text, display them in text object, and expect the line number should be equal to set text directly.

In this sample code, string in line 3 
```
community-lead continuation of the Phaser 2 codebase 
``` 
will be wrapped to 2 lines
```
community-lead continuation of the Phaser 2 
codebase 
```

It seems that extra "\n" will affect the wrapping result.

photonstorm:
Thanks for opening this issue, and for submitting a PR to fix it. We have merged your PR into the `master` branch and attributed the work to you in the [Change Log](https://github.com/photonstorm/phaser/blob/master/CHANGELOG.md). If you need to tweak the code for whatever reason please submit a new PR.

PR #3417, opened 2018-03-18
Bug fix: font might not sync to context

rexrainbow:

* Nothing, it's a bug fix

Describe the changes below:

Sync font before runWordWrap.

#3389 
In Text.updateText() function, **style.syncFont()** will be invoked before Text.runWordWrap(). But Text.getWrappedText() won't call style.syncFont(), therefore the wrapping result between Text.updateText() and Text.getWrappedText() might not be equal.

### DIFF
--- a/src/gameobjects/text/static/Text.js
+++ b/src/gameobjects/text/static/Text.js
@@ -486,6 +486,8 @@ var Text = new Class({
     {
         if (text === undefined) { text = this.text; }
 
+        this.style.syncFont(this.canvas, this.context);
+
         var wrappedLines = this.runWordWrap(text);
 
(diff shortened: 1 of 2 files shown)


rexrainbow:
Sorry I had not idea how to spit PRs, my mistake.

The 2nd PR is about custom loader.
I had integrated [google webfont loader into loader system by extending File Class. But I could not reference [CONST.FILE_POPULATED](https://github.com/rexrainbow/phaser3-rex-notes/blob/master/plugins/webfontloader/webFontLoaderPlugin.js#L23). Therefore I tried to expose `FILE_CONST` instead using the value `19` directly.

rexrainbow:
Thanks. I missed that submit before.

PR #3419, opened 2018-03-19
Update JSDoc on many files

orblazer:
This PR changes :

* Documentation

Describe the changes below:

I have fix many types in JSDoc. You abrove param description like [this](https://github.com/photonstorm/phaser/commit/af3c03a3cb1b91ef8c737dc0df5d3bf0a3e38d4a#diff-0ec8f711d16cbd1a5f9efefcb32401a9R39) ?

### DIFF
--- a/src/cameras/2d/CameraManager.js
+++ b/src/cameras/2d/CameraManager.js
@@ -200,7 +200,7 @@ var CameraManager = new Class({
             this.cameraPool.slice(poolIndex, 1);
             return camera;
         }
-        
+
         return null;
     },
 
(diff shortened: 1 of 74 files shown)


photonstorm:
This is superb, thank you. I think I'm going to have to expand out things like this `{{origin:Phaser.Math.Vector3,direction:Phaser.Math.Vector3}}` into actual TypeDefs blocks, but otherwise this is a superb amount of work and gets us even closer.

orblazer:
Ok, if you want after fix all types i can work on typeDefs generator, if you have tips or other i am interesting :)

PR #3422, opened 2018-03-19
Fix types and begin object TypeDef

orblazer:

* Documentation

Describe the changes below:

I have fix some types (like `array`) and begin replace `object` by reference to `@typedef`.


photonstorm:
Lovely, merged.

I've been getting rid of return types that are in-lined objects and swapping to TypeDefs or actual instances.

Also, please be aware of the case for these: `{Object.<string, any>}` (should be `{object.<string, any>}`.

orblazer:
Ops, the `Object` is wrong :) if you dont have fix this before my next PR i fix that.

photonstorm:
I need to head off for a bit, will carry on later tonight.

photonstorm:
Just to add, I've been using this to check the return types from functions: http://phaser.io/newdocs/3.0.0/types this will bring up a list of all distinct return types, then you can filter down to see what returns them. Handy for figuring out typos, etc :)

PR #3424, opened 2018-03-19
Added callback methods to Camera2D effects: Shake, Fade and Flash

pixelscripter:
This PR changes:
* The public-facing API

Added callback methods to Camera2D effects: Shake, Fade and Flash, so you can perform logic when they are finished.

So instead of having to use delayed calls:
`
this.camera.fade(3000);
this.time.delayedCall(3000, () => {
    this.scene.start('GameScene');
});
`

You can supply a callback method to the effect method:
`
this.camera.fade(3000, 0, 0, 0, false, () => {
    this.scene.start('GameScene');
});
`

I also added a Camera Effect Callbacks example to phaser3-examples. [See this PR](https://github.com/photonstorm/phaser3-examples/pull/38).


photonstorm:
Could be really useful, so have merged this in. I think I'm going to expand it slightly to pass over the camera instance to the callback too.

PR #3430, opened 2018-03-20
Fix types in Tilemaps

orblazer:

* Documentation

Describe the changes below:

I have fix many types in `Tilemaps` and i have change `{...|null}` by `{?...}` for conforming to nullable parameter on JSDoc.


photonstorm:
Are you sure you can use a question mark for a ‘@return’ ? I thought it was just for nullable properties.

orblazer:
Yes, the all js doc types is sames and is referenced here : [@use JSDoc > @type](http://usejsdoc.org/tags-type.html)

PR #3433, opened 2018-03-20
Fix "any" types

orblazer:

* Documentation

Describe the changes below:

I have fix `any` for conforming to JSDoc (replace `any` by `*`).


orblazer:
I have an question, for the custom types we merge all `@typedef` in `index.js` of namespace like here : https://github.com/photonstorm/phaser/blob/master/src/curves/index.js

photonstorm:
Merge them into the `index.js` or whatever file is suitable, we don't want to put them all in their own file, they need to remain with the code they are setting the def for.

PR #3439, opened 2018-03-21
Fix "object" types

orblazer:

* Documentation

Describe the changes below:

I have fix many `object` types.

I have 2 question :
- For the object (Point, Vector2, Rectangle) we create an objects like `PointLike`, `Vector2Like`, `RectangleLike` or just replace `object param by `Point`, `Vector2` and `Rectangle` ?
- We can generify Structs like default `Array` and `Object` (like that :  `List.<(type)>`) ?


photonstorm:
1) If all the method does is read the x/y values, then we could use `PointLike`. But if it passes the object elsewhere, or needs a vector op like subtract or add, then we need to just keep the type fixed.

2) Can you give me an example of where you'd use `List.<(type)>` please.

orblazer:
Ok, for exemple :

This list property ([LoaderPlugin.js#L169](https://github.com/photonstorm/phaser/blob/0ef92e1172ede8664fa78f3a896cac2f1255994c/src/loader/LoaderPlugin.js#L169)) can be change with `Phaser.Structs.Set.<Phaser.Loader.File>` and now when you get element in this list is an `File` object.

We can apply the generic in functions with out for increase autocomplete. Exemple (in TS):
```ts
export function Angle<G extends : Phaser.GameObjects.GameObject[]> (items: G, value: number, step?: number, index?: number, direction?: number): G
```
Usage :
```ts
const items: Phaser.GameObjects.Text = [...]

console.log(Angle(items, 10)) // The returned array is an same type of the "items" and not "Phaser.GameObjects.GameObject[]"
```

photonstorm:
Ok fine, when the Set (or whatever) contains a single data-type, let's do that.

I just pushed up a bunch of changes to the Tween files btw.

PR #3442, opened 2018-03-21
Change structs for generic types

orblazer:

* Documentation

Describe the changes below:

I have change the structs type for make generics types like `Array` and `Object` types.

The types `integer` and `float` is wrong or is correct ? If is correct we need create an alias for `number` type.

### DIFF
--- a/src/animations/AnimationManager.js
+++ b/src/animations/AnimationManager.js
@@ -80,7 +80,7 @@ var AnimationManager = new Class({
          * [description]
          *
          * @name Phaser.Animations.AnimationManager#anims
-         * @type {Phaser.Structs.Map}
+         * @type {Phaser.Structs.Map.<string, Phaser.Animations.Animation>}
          * @protected
          * @since 3.0.0
          */
--- a/src/cache/BaseCache.js
+++ b/src/cache/BaseCache.js
@@ -33,7 +33,7 @@ var BaseCache = new Class({
          * You can query the Map directly or use the BaseCache methods.
          *
          * @name Phaser.Cache.BaseCache#entries
-         * @type {Phaser.Structs.Map}
+         * @type {Phaser.Structs.Map.<string, *>}
          * @since 3.0.0
          */
         this.entries = new CustomMap();
--- a/src/cameras/sprite3d/Camera.js
+++ b/src/cameras/sprite3d/Camera.js
@@ -227,7 +227,7 @@ var Camera = new Class({
          * [description]
          *
          * @name Phaser.Cameras.Sprite3D#children
-         * @type {Phaser.Structs.Set}
+         * @type {Phaser.Structs.Set.<Phaser.GameObjects.Sprite3D>}
          * @since 3.0.0
          */
         this.children = new Set();
--- a/src/gameobjects/blitter/Blitter.js
+++ b/src/gameobjects/blitter/Blitter.js
@@ -90,7 +90,7 @@ var Blitter = new Class({
          * [description]
          *
          * @name Phaser.GameObjects.Blitter#children
-         * @type {Phaser.Structs.List}
+         * @type {Phaser.Structs.List.<Phaser.GameObjects.Blitter.Bob>}
          * @since 3.0.0
          */
         this.children = new List();
@@ -179,7 +179,7 @@ var Blitter = new Class({
      * @since 3.0.0
      *
      * @param {integer} quantity - The quantity of Bob objects to create.
-     * @param {string} [frame] - The Frame the Bobs will use. It must be part of the Blitter Texture.
+     * @param {[type]} [frame] - The Frame the Bobs will use. It must be part of the Blitter Texture. // TODO 21/03/2018 Fix this type
      * @param {boolean} [visible=true] - [description]
      *
      * @return {Phaser.GameObjects.Blitter.Bob[]} An array of Bob objects that were created.
--- a/src/gameobjects/group/Group.js
+++ b/src/gameobjects/group/Group.js
@@ -114,7 +114,7 @@ var Group = new Class({
          * [description]
          *
          * @name Phaser.GameObjects.Group#children
-         * @type {Phaser.Structs.Set}
+         * @type {Phaser.Structs.Set.<Phaser.GameObjects.GameObject>}
          * @since 3.0.0
          */
         this.children = new Set(children);
--- a/src/gameobjects/particles/ParticleEmitterManager.js
+++ b/src/gameobjects/particles/ParticleEmitterManager.js
@@ -114,7 +114,7 @@ var ParticleEmitterManager = new Class({
          * A list of Emitters being managed by this Emitter Manager.
          *
          * @name Phaser.GameObjects.Particles.ParticleEmitterManager#emitters
-         * @type {Phaser.Structs.List}
+         * @type {Phaser.Structs.List.<Phaser.GameObjects.Particles.ParticleEmitter>}
          * @since 3.0.0
          */
         this.emitters = new List(this);
@@ -123,7 +123,7 @@ var ParticleEmitterManager = new Class({
          * A list of Gravity Wells being managed by this Emitter Manager.
          *
          * @name Phaser.GameObjects.Particles.ParticleEmitterManager#wells
-         * @type {Phaser.Structs.List}
+         * @type {Phaser.Structs.List.<Phaser.GameObjects.Particles.GravityWell>}
          * @since 3.0.0
          */
         this.wells = new List(this);
--- a/src/loader/LoaderPlugin.js
+++ b/src/loader/LoaderPlugin.js
@@ -166,7 +166,7 @@ var LoaderPlugin = new Class({
          * [description]
          *
          * @name Phaser.Loader.LoaderPlugin#list
-         * @type {Phaser.Structs.Set}
+         * @type {Phaser.Structs.Set.<Phaser.Loader.File>}
          * @since 3.0.0
          */
         this.list = new CustomSet();
@@ -175,7 +175,7 @@ var LoaderPlugin = new Class({
          * [description]
          *
          * @name Phaser.Loader.LoaderPlugin#inflight
-         * @type {Phaser.Structs.Set}
+         * @type {Phaser.Structs.Set.<Phaser.Loader.File>}
          * @since 3.0.0
          */
         this.inflight = new CustomSet();
@@ -184,7 +184,7 @@ var LoaderPlugin = new Class({
          * [description]
          *
          * @name Phaser.Loader.LoaderPlugin#failed
-         * @type {Phaser.Structs.Set}
+         * @type {Phaser.Structs.Set.<Phaser.Loader.File>}
          * @since 3.0.0
          */
         this.failed = new CustomSet();
@@ -193,7 +193,7 @@ var LoaderPlugin = new Class({
          * [description]
          *
          * @name Phaser.Loader.LoaderPlugin#queue
-         * @type {Phaser.Structs.Set}
+         * @type {Phaser.Structs.Set.<Phaser.Loader.File>}
          * @since 3.0.0
          */
         this.queue = new CustomSet();
@@ -202,7 +202,7 @@ var LoaderPlugin = new Class({
          * [description]
          *
          * @name Phaser.Loader.LoaderPlugin#storage
-         * @type {Phaser.Structs.Set}
+         * @type {Phaser.Structs.Set.<(Phaser.Loader.File|object)>} // TODO 21/03/2018 Fix type
          * @since 3.0.0
          */
         this.storage = new CustomSet();
--- a/src/loader/LoaderPlugin.js
+++ b/src/loader/LoaderPlugin.js
@@ -37,6 +37,14 @@ var XHRSettings = require('./XHRSettings');
  */
 
 /**
+ * @typedef {object} LinkFileObject
+ *
+ * @property {string} type - [description]
+ * @property {Phaser.Loader.File} fileA - [description]
+ * @property {Phaser.Loader.File} fileB - [description]
+ */
+
+/**
  * @classdesc
  * [description]
  *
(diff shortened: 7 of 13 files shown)


photonstorm:
Where are the templates for T and K defined?

`integer` and `float` are fine and define an explicit fixed type, where `number` could be either.

photonstorm:
Just to add, why would we need an alias for `number`? These are all valid numeric types.

orblazer:
The template is delcrared here [Map.js#28](https://github.com/photonstorm/phaser/pull/3442/files#diff-8d4005e984349525daadeacee34db9feR28) and is it same on all structs files.

Yes in programing the type `integer` and `float` is rigth but in JS we have only `number` type.

photonstorm:
Are you sure `@template` is actually parsed by jsdoc and emits a valid type in the output? It's not listed as a known tag on their site.

The purpose of `integer` and `float` are to explain intent to someone reading the docs, regardless if it's a valid data type in JS. If TS doesn't support them then they'll need converting to `number` types as part of the conversion process.

orblazer:
Hum I don't find references for `integer` or `float` in jsdocs. 

The `@template` is an tag from Google Closure but jsdocs supports that.

Src: https://github.com/jsdoc3/jsdoc/issues/605

photonstorm:
It doesn't matter that int/f aren't listed in jsdocs, neither is the `@webglOnly` tag I use a lot, what matters is that the docs template knows about them and uses them, which helps the devs reading them. For the sake of conversion to TS they'll need converting to `number` types.

I need to run some tests to see what the jsdoc output looks like when it hits a template definition, because when browsing the source code it is now completely meaningless, but as long as jsdoc understands the type is actually an object and not just the letter T then it will be fine.

orblazer:
OK I can write an test tomorrow (I'm French) and for int/float we can make an alias for the ide recognize that types.

photonstorm:
Ok, as it stands currently jsdoc is having trouble parsing some of the new types. Here is a full dump of all the errors (this is from a branch created from this PR):

```
ERROR: Unable to parse a tag's type expression for source file D:\wamp\www\phaser\src\display\color\Color.js in line 80 with tag title "type" and text "{[number,number,number,number]}": Invalid type expression "[number,number,number,number]": Expected "!", "$", "'", "(", "*", ".", "...", "0", "?", "@", "Function", "\"", "\\", "_", "break", "case", "catch", "class", "const", "continue", "debugger", "default", "delete", "do", "else", "enum", "export", "extends", "false", "finally", "for", "function", "if", "implements", "import", "in", "instanceof", "interface", "let", "new", "null", "package", "private", "protected", "public", "return", "static", "super", "switch", "this", "throw", "true", "try", "typeof", "undefined", "var", "void", "while", "with", "yield", "{", Unicode letter number, Unicode lowercase letter, Unicode modifier letter, Unicode other letter, Unicode titlecase letter, Unicode uppercase letter or
[1-9] but "[" found.
ERROR: Unable to parse a tag's type expression for source file D:\wamp\www\phaser\src\display\color\Color.js in line 87 with tag title "type" and text "{[number,number,number,number]}": Invalid type expression "[number,number,number,number]": Expected "!", "$", "'", "(", "*", ".", "...", "0", "?", "@", "Function", "\"", "\\", "_", "break", "case", "catch", "class", "const", "continue", "debugger", "default", "delete", "do", "else", "enum", "export", "extends", "false", "finally", "for", "function", "if", "implements", "import", "in", "instanceof", "interface", "let", "new", "null", "package", "private", "protected", "public", "return", "static", "super", "switch", "this", "throw", "true", "try", "typeof", "undefined", "var", "void", "while", "with", "yield", "{", Unicode letter number, Unicode lowercase letter, Unicode modifier letter, Unicode other letter, Unicode titlecase letter, Unicode uppercase letter or
[1-9] but "[" found.
WARNING: The @type tag does not permit a description; the description will be ignored. File: LoaderPlugin.js, line: 201
WARNING: The @type tag does not permit a description; the description will be ignored. File: LoaderPlugin.js, line: 208
ERROR: Unable to parse a tag's type expression for source file D:\wamp\www\phaser\src\math\random-data-generator\RandomDataGenerator.js in line 172 with tag title "param" and text "{*[]} seeds - The array of seeds: the `toString()` of each value is used.": Invalid type expression "*[]": Expected "!", "=", "?", "|" or end of input but "[" found.
ERROR: Unable to parse a tag's type expression for source file D:\wamp\www\phaser\src\math\random-data-generator\RandomDataGenerator.js in line 182 with tag title "param" and text "{*[]} seeds - The array of seeds: the `toString()` of each value is used.": Invalid type expression "*[]": Expected "!", "=", "?", "|" or end of input but "[" found.
ERROR: Unable to parse a tag's type expression for source file D:\wamp\www\phaser\src\physics\arcade\World.js in line 1786 with tag
title "param" and text "{*[]} arr": Invalid type expression "*[]": Expected "!", "=", "?", "|" or end of input but "[" found.
ERROR: Unable to parse a tag's type expression for source file D:\wamp\www\phaser\src\physics\arcade\World.js in line 1795 with tag
title "param" and text "{*[]} arr": Invalid type expression "*[]": Expected "!", "=", "?", "|" or end of input but "[" found.
ERROR: The @property tag requires a value. File: Scene.js, line: 10
ERROR: Unable to parse a tag's type expression for source file D:\wamp\www\phaser\src\scene\Settings.js in line 11 with tag title "property" and text "{(false|[type])} [files=false] - [description]": Invalid type expression "(false|[type])": Expected "!", "$", "'", "(", "*", ".", "...", "0", "?", "@", "Function", "\"", "\\", "_", "break", "case", "catch", "class", "const", "continue", "debugger", "default", "delete", "do", "else", "enum", "export", "extends", "false", "finally", "for", "function", "if", "implements", "import", "in", "instanceof", "interface", "let", "new", "null", "package", "private", "protected", "public", "return", "static", "super", "switch", "this", "throw", "true", "try", "typeof", "undefined", "var", "void", "while", "with", "yield", "{", Unicode letter number, Unicode lowercase letter, Unicode modifier letter, Unicode other letter, Unicode titlecase letter, Unicode uppercase letter or [1-9] but "[" found.
ERROR: Unable to parse a tag's type expression for source file D:\wamp\www\phaser\src\scene\Settings.js in line 11 with tag title "property" and text "{(false|[type])} [plugins=false] - [description]": Invalid type expression "(false|[type])": Expected "!", "$", "'", "(", "*", ".", "...", "0", "?", "@", "Function", "\"", "\\", "_", "break", "case", "catch", "class", "const", "continue", "debugger", "default", "delete", "do", "else", "enum", "export", "extends", "false", "finally", "for", "function", "if", "implements", "import", "in", "instanceof", "interface", "let", "new", "null", "package", "private", "protected", "public", "return", "static", "super", "switch", "this", "throw", "true", "try", "typeof", "undefined", "var", "void", "while", "with", "yield", "{", Unicode letter number, Unicode lowercase letter, Unicode modifier letter, Unicode other letter, Unicode titlecase letter, Unicode uppercase letter or [1-9] but "[" found.
ERROR: Unable to parse a tag's type expression for source file D:\wamp\www\phaser\src\scene\Settings.js in line 25 with tag title "property" and text "{(false|[type])} files - [description]": Invalid type expression "(false|[type])": Expected "!", "$", "'", "(", "*", ".", "...", "0", "?", "@", "Function", "\"", "\\", "_", "break", "case", "catch", "class", "const", "continue", "debugger", "default", "delete", "do", "else", "enum", "export", "extends", "false", "finally", "for", "function", "if", "implements", "import", "in", "instanceof", "interface", "let", "new", "null", "package", "private", "protected", "public", "return", "static", "super", "switch", "this", "throw", "true", "try", "typeof", "undefined", "var", "void", "while", "with", "yield", "{", Unicode letter number, Unicode lowercase letter, Unicode modifier letter, Unicode other letter, Unicode titlecase letter, Unicode uppercase letter or [1-9] but "[" found.
ERROR: Unable to parse a tag's type expression for source file D:\wamp\www\phaser\src\scene\Settings.js in line 25 with tag title "property" and text "{(false|[type])} plugins - [description]": Invalid type expression "(false|[type])": Expected "!", "$", "'", "(", "*", ".", "...", "0", "?", "@", "Function", "\"", "\\", "_", "break", "case", "catch", "class", "const", "continue", "debugger", "default", "delete", "do", "else", "enum", "export", "extends", "false", "finally", "for", "function", "if", "implements", "import", "in", "instanceof", "interface", "let", "new", "null", "package", "private", "protected", "public", "return", "static", "super", "switch", "this", "throw", "true", "try", "typeof", "undefined", "var", "void", "while", "with", "yield", "{", Unicode letter number, Unicode lowercase letter, Unicode modifier letter, Unicode other letter, Unicode titlecase letter, Unicode uppercase letter or [1-9] but "[" found.
ERROR: Unable to parse a tag's type expression for source file D:\wamp\www\phaser\src\scene\Settings.js in line 42 with tag title "property" and text "{(false|[type])} files - [description]": Invalid type expression "(false|[type])": Expected "!", "$", "'", "(", "*", ".", "...", "0", "?", "@", "Function", "\"", "\\", "_", "break", "case", "catch", "class", "const", "continue", "debugger", "default", "delete", "do", "else", "enum", "export", "extends", "false", "finally", "for", "function", "if", "implements", "import", "in", "instanceof", "interface", "let", "new", "null", "package", "private", "protected", "public", "return", "static", "super", "switch", "this", "throw", "true", "try", "typeof", "undefined", "var", "void", "while", "with", "yield", "{", Unicode letter number, Unicode lowercase letter, Unicode modifier letter, Unicode other letter, Unicode titlecase letter, Unicode uppercase letter or [1-9] but "[" found.
ERROR: Unable to parse a tag's type expression for source file D:\wamp\www\phaser\src\scene\Settings.js in line 42 with tag title "property" and text "{(false|[type])} plugins - [description]": Invalid type expression "(false|[type])": Expected "!", "$", "'", "(", "*", ".", "...", "0", "?", "@", "Function", "\"", "\\", "_", "break", "case", "catch", "class", "const", "continue", "debugger", "default", "delete", "do", "else", "enum", "export", "extends", "false", "finally", "for", "function", "if", "implements", "import", "in", "instanceof", "interface", "let", "new", "null", "package", "private", "protected", "public", "return", "static", "super", "switch", "this", "throw", "true", "try", "typeof", "undefined", "var", "void", "while", "with", "yield", "{", Unicode letter number, Unicode lowercase letter, Unicode modifier letter, Unicode other letter, Unicode titlecase letter, Unicode uppercase letter or [1-9] but "[" found.
ERROR: Unable to parse a tag's type expression for source file D:\wamp\www\phaser\src\structs\Map.js in line 19 with tag title "param" and text "{*[]} elements - [description]": Invalid type expression "*[]": Expected "!", "=", "?", "|" or end of input but "[" found.
ERROR: Unable to parse a tag's type expression for source file D:\wamp\www\phaser\src\structs\Map.js in line 37 with tag title "param" and text "{*[]} elements - [description]": Invalid type expression "*[]": Expected "!", "=", "?", "|" or end of input but "[" found.
ERROR: Unable to parse a tag's type expression for source file D:\wamp\www\phaser\src\time\Clock.js in line 152 with tag title "param" and text "{*[]} args - [description]": Invalid type expression "*[]": Expected "!", "=", "?", "|" or end of input but "[" found.
ERROR: Unable to parse a tag's type expression for source file D:\wamp\www\phaser\src\time\Clock.js in line 165 with tag title "param" and text "{*[]} args - [description]": Invalid type expression "*[]": Expected "!", "=", "?", "|" or end of input but "[" found.
ERROR: Unable to parse a tag's type expression for source file D:\wamp\www\phaser\src\time\TimerEvent.js in line 10 with tag title "property" and text "{*[]} [args] - [description]": Invalid type expression "*[]": Expected "!", "=", "?", "|" or end of input but "[" found.
```

I checked one of the new `@template` types and it doesn't understand what they are. It allows them through because the config setting is lenient, but this is what they come out as in the doclets:

```
    {
      "comment": "/**\r\n     * [description]\r\n     *\r\n     * @method Phaser.Structs.Map#set\r\n     * @since 3.0.0\r\n     *\r\n     * @param {object} key - [description]\r\n     * @param {T} value - [description]\r\n     *\r\n     * @return {Phaser.Structs.Map} This Map object.\r\n     */",
      "meta": {
        "filename": "Map.js",
        "lineno": 72,
        "columnno": 4,
        "path": "D:\\wamp\\www\\phaser\\src\\structs",
        "code": {}
      },
      "description": "[description]",
      "kind": "function",
      "name": "set",
      "since": "3.0.0",
      "params": [
        {
          "type": {
            "names": [
              "K"
            ]
          },
          "description": "[description]",
          "name": "key"
        },
        {
          "type": {
            "names": [
              "T"
            ]
          },
          "description": "[description]",
          "name": "value"
        }
      ],
      "returns": [
        {
          "type": {
            "names": [
              "Phaser.Structs.Map"
            ]
          },
          "description": "This Map object."
        }
      ],
      "memberof": "Phaser.Structs.Map",
      "longname": "Phaser.Structs.Map#set",
      "scope": "instance",
      "___id": "T000002R024355",
      "___s": true
    },
```

It doesn't appear to have a clue what the type is, so has just inserted it 'T' and 'K', which is a shame - I thought the point of a template was that it would map it back to an object again, like a TypeDef does.

Not really sure what to do about this, as it stands it will literally print the types of the Map.set method as T and K in the output, which is a bit nuts :)

orblazer:
This is logic out T and K type because is it generic type. But if you want we can rename Generic names Like `T > VALUE` and `K > KEY`

I have try search if we can make predefined generic like my exemple (`<G extends Phaser.GameObjects.GameObject>`) but i don't have found any solution for that :(

The generic type is pretty usefull for make TypeDefs. I search an method for make generic types for Actions methods for increase autocomplete when use Typescript.

orblazer:
Sorry for this commit : https://github.com/photonstorm/phaser/pull/3442/commits/63c9938d412107d7044c7c57aa1ba54f185a30d9
But i can't remove that commit ;(

photonstorm:
Could you use a custom tag? (something like `@generic T`) which the converter could then read and change the type accordingly? Then it will be useful in the TS Defs without making the JSDoc unreadable.

I just released 3.3, so I can close off this PR if you like and you could rebase against 3.4?

orblazer:
OK, seen that i rewrite all code we can close this PR.

PR #3444, opened 2018-03-22
Fix wrong types

orblazer:

* Documentation

Describe the changes below:

I have fix all types genereate error when run JSDoc.


photonstorm:
Thanks, I was just about to do this, so good timing!

Issue #3385, opened 2018-03-13
Rendertexture alpha not changing in WEBGL mode.


samid737:
This Issue is about

* A bug in the API (v3.2.1)

The (global? not sure) alpha of the rendertexture does not change in WEBGL mode. It does work in CANVAS.

CANVAS:
http://jsfiddle.net/e2e6w77r/14/

WEBGL:
http://jsfiddle.net/e2e6w77r/17/

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

PR #3445, opened 2018-03-22
Fix changing alpha in RenderTextureWebGLRenderer

ger1995:
* Nothing, it's a bug fix

Describe the changes below:
This is a fix for https://github.com/photonstorm/phaser/issues/3385. The tint of the texture was always set to 0xffffff and therefor the alpha values were ignored. The tint is now calculated using the alpha value.


photonstorm:
Lovely, thanks.

PR #3458, opened 2018-03-26
Added joint debug rendering to Matter Physics postUpdate

OmarShehata:
This PR changes

* Nothing, it's a bug fix

Describe the changes below:

This PR adds rendering of joints in the debug render of MatterPhysics.

I was experimenting with Matterjs physics and noticed only the bodies get drawn in the debug view, not anything else. I needed to view joints, so I added this and thought I'd share. I've ported over the debug rendering from the [Matter renderer](https://github.com/liabru/matter-js/blob/master/src/render/Render.js#L489-L557) with some changes (to use Phaser graphics instead of drawing directly to a canvas).

The `eslint` produces no errors, but I had to use `let` instead of `var` in the for loop (not sure about the status of that in the style guide) because otherwise it would complain about redefinition (and I was worried that just declaring `i` in the first for loop might be an issue in the future if we decide to add the ability to only render joints and not bodies, so that first for loop might not always execute).

### DIFF
--- a/src/physics/matter-js/World.js
+++ b/src/physics/matter-js/World.js
@@ -17,6 +17,8 @@ var MatterBody = require('./lib/body/Body');
 var MatterEvents = require('./lib/core/Events');
 var MatterWorld = require('./lib/body/World');
 var MatterTileBody = require('./MatterTileBody');
+var Vector = require('./lib/geometry/Vector');
+var Common = require('./lib/core/Common');
 
 /**
  * @classdesc
@@ -164,7 +166,9 @@ var World = new Class({
             debugShowVelocity: GetValue(config, 'debugShowVelocity', true),
             bodyDebugColor: GetValue(config, 'debugBodyColor', 0xff00ff),
             staticBodyDebugColor: GetValue(config, 'debugBodyColor', 0x0000ff),
-            velocityDebugColor: GetValue(config, 'debugVelocityColor', 0x00ff00)
+            velocityDebugColor: GetValue(config, 'debugVelocityColor', 0x00ff00),
+            debugShowJoint: GetValue(config, 'debugShowJoint', true),
+            jointDebugColor: GetValue(config, 'debugJointColor', 0x000000)
         };
 
         if (this.drawDebug)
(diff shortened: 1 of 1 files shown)


kvnndr0:
On Mar 26, 2018 8:24 AM, Omar Shehata <notifications@github.com> wrote:

This PR changes

  *   Nothing, it's a bug fix

Describe the changes below:

This PR adds rendering of joints in the debug render of MatterPhysics.

I was experimenting with Matterjs physics and noticed only the bodies get drawn in the debug view, not anything else. I needed to view joints, so I added this and thought I'd share. I've ported over the debug rendering from the Matter renderer<https://github.com/liabru/matter-js/blob/master/src/render/Render.js#L489-L557> with some changes (to use Phaser graphics instead of drawing directly to a canvas).

The eslint produces no errors, but I had to use let instead of var in the for loop (not sure about the status of that in the style guide) because otherwise it would complain about redefinition (and I was worried that just declaring i in the first for loop might be an issue in the future if we decide to add the ability to only render joints and not bodies, so that first for loop might not always execute).

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/photonstorm/phaser/pull/3458

Commit Summary

  *   Added joint debug rendering to Matter Physics postUpdate

File Changes

  *   M src/physics/matter-js/World.js<https://github.com/photonstorm/phaser/pull/3458/files#diff-0> (100)

Patch Links:

  *   https://github.com/photonstorm/phaser/pull/3458.patch
  *   https://github.com/photonstorm/phaser/pull/3458.diff

photonstorm:
Thanks for your PR. I like the look of the additions you've made. Please could you refactor it to not use any ES6 syntax, as v3 is ES5 and is not transpiled, so this will break older browsers if merged as-is. Once that is done I'll be happy to include it!

OmarShehata:
Thanks for the quick review @photonstorm ! I've been a huge fan of Phaser and your work for many years, so it's really cool to be able to contribute now. I've removed all uses of `let` in the last commit. I have one more question:

* Is there a straightforward way to update the Matterjs version in Phaser? 
  * I ask because my chain of joints is unstable in Phaser, but works in the Matterjs demo, and I believe they have the same settings, so I suspect it might be a version issue.

PR #3461, opened 2018-03-27
Fix many types and add many generics type

orblazer:

* Documentation

Describe the changes below:

## Questions :
1. We can remove `array` from `Phaser.Actions` ?
2. We can split object params, exemple in [Phaser.Cameras.Scene2D.Camera#28](https://github.com/photonstorm/phaser/blob/master/src/cameras/2d/Camera.js#L28) ?
3. For `Vector2Like`, `Vector3Like`... We replace by one type like this ?
```js
/**
 * @typedef Vector2Like
 *
 * @property {number} x - [description]
 * @property {number} y - [description]
 */
```
4. You have an idea for an date for release Typescript definitions ?

# Structures of `@generic` and `@genericUse`  :
```
@generic {[type]} [name] - [[property1],[property2],[return]]
```
```
@genericUse {[type2]} [name] - [[property1],[property2],[$return]]
```

Explain :
* `[type]` : Represent "extends [type]" of generic
* `[type2]` : Reprensent the new type
* `[name]` Represent "name" of the generic like `VALUE`
* `[[property1],[property2],[$return]]` : Represent the list of the property apply this generic. For replace `@type` use and `$type` and for `@return`, '$return`

__Exemple for `@generic` :__
Input : 
```js
/**
 * [...]
 *
 * @function Phaser.Actions.Angle
 * @since 3.0.0
 *
 * @generic {Phaser.GameObjects.GameObject[]} G - [items,$return]
 *
 * @param {(array|Phaser.GameObjects.GameObject[])} items - The array of items to be updated by this action.
 * [...]
 *
 * @return {(array|Phaser.GameObjects.GameObject[])} The array of objects that were passed to this Action.
 */
var Angle = function (items, value, step, index, direction) [...]
```
Output :
```ts
/**
 * [...]
 *
 * @function Phaser.Actions.Angle
 * @since 3.0.0
 *
 * @param {(array|Phaser.GameObjects.GameObject[])} items - The array of items to be updated by this action.
 * [...]
 *
 * @return {(array|Phaser.GameObjects.GameObject[])} The array of objects that were passed to this Action.
 */
export function Angle<G extends Array<Phaser.GameObjects.GameObject>> (items: G, value: number, step?: number, index?: integer, direction?: integer): G
```

__Exemple for `@genericUse` :__
Input : 
```js
/**
 * The objects that belong to this collection.
 *
 * @genericUse {T[]} T - [$type]
 *
 * @name Phaser.Structs.List#list
 * @type {Array.<*>}
 * @default []
 * @since 3.0.0
 */
```
Output :
```ts
/**
 * The objects that belong to this collection.
 *
 * @name Phaser.Structs.List#list
 * @type {T[]}
 * @default []
 * @since 3.0.0
 */
```

For more readable i propose add mapped types like this : https://blog.mariusschulz.com/2017/01/20/typescript-2-1-mapped-types (on generate code)
Exemple : for this type `?(object|Phaser.GameObjects.GameObject)` we are replaced by `Nullable<Phaser.GameObjects.GameObject>`


photonstorm:
1) Nope, because most Actions will work on any type of object, not just Game Objects.

2) That should be valid, jsdocs doesn't complain about it and the docblock it makes looks ok. So I think safe to use in some cases (like inside a TypeDef), but better to use a TypeDef in all other cases.

3) Should be fine: `@typedef {object} Vector2Like`

4) Nope. We've not tried re-generating them yet. Are yours generating properly yet?

I like the look of the `@generic` tag, nice and obvious. It did make me think, do you reckon they might add generics into a future version of ecmascript? In which case I wonder if `@generic` will become a standard jsdoc tag?! I guess if that happens it will be years away anyway :)

orblazer:
OK and for the TypeDef (making before my first PR and i have make by my hand :( ).

 I will be happy if they add Generic type in future like `@template` from Google Closure , this would allow a much more complete and intuitive JSDoc :)

Issue #3363, opened 2018-03-10
Phaser 2.x State / Phaser 3.x Scene - passing data 


pixelhijack:
following the Scene Manager documentation https://phaser.io/phaser3/api/scene-manager

**Setup**

```
const bootScene = {
    key: 'boot',
    active: true,
    init: (config) => {
        console.log('[BOOT] init', config);
    },
    preload: () => {
        console.log('[BOOT] preload');
    },
    create: (config) => {
        console.log('[BOOT] create', config);
    },
    update: () => {
        console.log('[BOOT] update');
    }
};

const gameScene = {
    key: 'game',
    active: true,
    renderToTexture: true,
    x: 64,
    y: 64,
    width: 320,
    height: 200,
    init: (config) => {
        console.log('[GAME] init', config);
    },
    preload: () => {
        console.log('[GAME] preload');
    },
    create: (config) => {
        console.log('[GAME] create', config);
    },
    update: () => {
        console.log('[GAME] update');
    }
};

const gameConfig = {
    type: Phaser.CANVAS,
    parent: 'phaser-example',
    width: 800,
    height: 600,
    scene: [ bootScene, gameScene ]
};

const game = new Phaser.Game(gameConfig);
game.scene.start('boot', { someData: '...arbitrary data' });
```

**Expected behaviour**

Based on Phaser 2.x States, what i want to achieve and expect is:

1. I have two scenes, both with unique keys i give ("boot", "game") their corresponding config data (JSON level, game config etc) which i want to pass down
2. In Phaser 2.x and supposedly in Phaser 3 this data is passed to the state/scene .init() and .create() methods
3. I can initialise first Boot scene, from where (if any key hit etc) i can start the Game scene

**Actual behaviour**

1. Scenes are initialised with the "default" fallback key
2. My data is never passed down, only a default { } empty object i got
3. Scenes are starting at the same time, their update() methods are called at the same time

![screen shot 2018-03-09 at 13 25 40](https://user-images.githubusercontent.com/5672651/37241066-7bba4cbc-2453-11e8-9129-38b3278c8491.png)

**Cause?**

```
for (var i = 0; i < this._pending.length; i++)
            {
                var entry = this._pending[i];

                if (entry.key === key)
                {
                    entry.autoStart = true;
                    entry.data = data;
                }
            }
```

[...]

PR #3470, opened 2018-03-29
[ISSUE #3363] Phaser Scene - passing data to scene init & create methods

pixelhijack:
**[ISSUE #3363] Phaser Scene - passing data to scene init & create methods**

Fixing / workaround for issue #3363 

As described in [issue #3363](https://github.com/photonstorm/phaser/issues/3363), in Phaser 3.x the ability of Phaser 2.x passing arbitrary initial data (config) to a starting (boot) scene was broken, as it is not working as expected in the examples either: https://labs.phaser.io/edit.html?src=src\scenes\boot%20data.js ({ wibble: 456 } is never logged out, as it was in Phaser 2.x). 

The issue is, the optional data parameter in the .start() method is copied to entry.data, but after it is added to this._pending as scene, which will re-add the newScene on start, it is not copied further.

This commit uses a workaround by copying this optional data param onto the scene (where it is not the proper place i guess), hopefully without causing regression. 

Also, the idea overlaps a bit with the config's optional 'extend' prop, making it a bit redundant, but now starting a scene like this should be working, available in .init() and .create() methods of the scene, instead of the default empty object: 

`game.scene.start('boot', { someData: '...arbitrary data from BOOT' });`

### DIFF
--- a/src/scene/SceneManager.js
+++ b/src/scene/SceneManager.js
@@ -180,7 +180,7 @@ var SceneManager = new Class({
         {
             entry = this._start[i];
 
-            this.start(entry);
+            this.start(entry, entry.scene.data);
         }
 
         this._start.length = 0;
@@ -219,7 +219,7 @@ var SceneManager = new Class({
             {
                 entry = this._start[i];
 
-                this.start(entry);
+                this.start(entry, entry.scene.data);
             }
 
             //  Clear the pending lists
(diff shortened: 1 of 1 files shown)


photonstorm:
Hmm, this now breaks any Scene that doesn't have `data` passed with it :)

Will look into it more now.

pixelhijack:
@photonstorm yeah it might need a default value and as i saw it behaves differently if scenes are defined in config, or directly, or ES6 extending the Phaser.Scene as a class or used with arrow lifecycle methods (i guess because of the `this` context bindings)

photonstorm:
No worries, I'm working on it at the moment.

PR #3474, opened 2018-03-30
[Issue 3473]:Animation support for varying image sizes

jp-gc:

* Nothing, it's a bug fix

Describe the changes below:
This patch allows the support for an animation to contain varying sizes of images and correctly anchoring them during the animation.


photonstorm:
The problem with doing it like this is that it ignores frames that have custom pivot points, such as those exported from Texture Packer where the pivot has been positoned carefully in order to prevent what you're seeing here.

photonstorm:
I think it'd be better to have the Animation component call the Texture component so it handles custom pivots. I'll have a look at it shortly.

Issue #3472, opened 2018-03-29
Bitmap Font letter spacing


nickaversano:
Hi, I was wondering if it was possible to adjust the letter spacing for Bitmap Fonts? I guess this could be applicable for all font objects too. I'm following [this example](https://github.com/photonstorm/phaser3-examples/blob/master/public/src/game%20objects/bitmaptext/dynamic/tinted%20retro%20text.js) to render a retro bitmap font and noticing that there's a 1px space between each character in a tightly-packed font sprite where all the characters are next to each other.

Any advice would be great!

photonstorm:
Thank you for submitting this feature request. This has been implemented and the feature has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

PR #3480, opened 2018-03-30
Feature Request #3472

wtravO:
This PR references Issue #3472 

* Documentation
* TypeScript Defs
* The public-facing API

Describe the changes below:
This PR will add a kerning property to BitmapText that allows developers to add additional spacing between characters.  Default val is 0.


photonstorm:
I like this, but I wonder if it's the wrong property name to use. Because `kerning` values are already specified inside the Bitmap Text data files and applied during rendering. You can see them extracted in `ParseXMLBitmapFont.js` and used in the renderer.

wtravO:
Yeah, I was thinking about that as well. Perhaps `charPadding`?

harriscory730:
So Mr Richard what's is next

wtravO:
I believe I had Kerning confused with Letter Spacing in this scenario.  So, in this case, I believe it should be named `letterSpacing`.  Should avoid confusion and match up better to common typography terminology.

Referencing this article : https://graphicdesign.stackexchange.com/questions/2606/difference-between-kerning-vs-letter-spacing

photonstorm:
`letterSpacing` is much better, thanks.

wtravO:
It's just when I created the branch. :) Had multiple PR's in at the same time.

wtravO:
Not sure what happened there.  It pulled the creation of the branch in as a commit it seems?  Can be reverted and squashed perhaps.  Should fix it.

PR #3494, opened 2018-04-02
JSdoc, changed properties marked with [?...] to just be [...]

erd0s:

* Documentation
* TypeScript Defs

Describe the changes below:

Changed some jsdoc definitions that were causing weird property names in the d.ts file. I think the intention here would have been to mark these as _nullable_ as per http://usejsdoc.org/tags-type.html but the [] already indicates optional... I think the ? can be safely removed.


erd0s:
Closed because of comments from https://github.com/photonstorm/phaser3-docs/issues/12

photonstorm:
I'm happy to merge this because config object properties are never nullable, only class properties are.

erd0s:
Hey mjrosencrantz, [someVar] should denote a variable as optional in JSdoc, ? on the _type_ should indicate that it’s nullable, but I don’t think it’s valid to have both [] and ? on the variable, are you sure this is causing your issue?

mjrosencrantz:
Perhaps there is a bug with whatever code is converting the jsdoc to typescript definitions.  I'll look for it.  Previously in the typescript definition file the fields of GameConfig came out like: `parent?: any;` Now they're just: `parent: any;`

photonstorm:
This is already being resolved in the new parser branch in the docs repo. No need for anyone to do anything, we're already on it.

PR #3504, opened 2018-04-04
jsdoc / tsdef: fix correct type reference on Scene#load

ampled:

* TypeScript Defs

Describe the changes below:

TypeScript definition errored at Phaser.Scenes.Scene.load as it tried to reference Phaser.Loader.LoadPlugin (should be Phaser.Loader.LoaderPlugin)


photonstorm:
👍

PR #3509, opened 2018-04-05
Line.PointA&B fix

Tomas2h:

* Nothing, it's a bug fix

Describe the changes below:
It fixes the 2 methods line.pointA and line.pointB as the object Vector2 can only handle .set(..).
This also fixes the issues 41 42 and 44 of the phaser3-example repo.


photonstorm:
Well spotted, thanks :) I've actually added Vector2.setTo as an alias for Vector2.set, but I'll merge this anyway as the code is a little shorter and more direct.

PR #3511, opened 2018-04-05
Fixed renderCanvas() call in Graphics.generateTexture()

hexus:
Generating textures from graphics objects stopped working due to commit (c232642a) altering the signature of the `GraphicsCanvasRenderer`.

I've just updated the call to the renderer to pass null for the parent transform (the newly introduced parameter), which I assume is the correct thing to do here.

Glad I caught this before release. :)


photonstorm:
👍

PR #3512, opened 2018-04-05
filling in most of the jsdocs for GetFastValue and SpriteSheetFromAtlas

icbat:
This PR changes:

* Documentation

Describe the changes below:

Filled in jsdocs to the best of my ability for a few more classes.


photonstorm:
Thanks, I'll merge this in. Please note that `@memberOf` is correct though, I use the camel-case versions of jsdoc tags, so they don't need replacing to lower-case.

icbat:
Oh, thanks for the feedback! Erred on the side of jsdocs's docs, but I should've just left it. Will do so in the future. Thanks!

PR #3522, opened 2018-04-07
documenting UppercaseFirst

icbat:

* Documentation

Describe the changes below:

Adds documentation to a utils/string/UppercaseFirst.js.

Submitting this by itself as I'm not 100% sure on how `@example` should be used. [JSDocs](http://usejsdoc.org/tags-example.html) has it with a separate `@example` tag per example, but [utils/array/NumberArrayStep.js](https://github.com/photonstorm/phaser/blob/master/src/utils/array/NumberArrayStep.js) has one tag with several examples underneath.

Unsure how this renders to make a decision. Let me know if I should change it!

### DIFF
--- a/src/utils/string/UppercaseFirst.js
+++ b/src/utils/string/UppercaseFirst.js
@@ -5,14 +5,23 @@
  */
 
 /**
- * [description]
+ * Capitalizes the first letter of a string if there is one.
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

PR #3534, opened 2018-04-09
Ellipse random area calculation fixed

budda:
Ellipse.random() didn't return a point anywhere in the Ellipse area.

Full details and screenshots (before & after) can be found on https://github.com/photonstorm/phaser-ce/issues/522 assuming the issue is the same in Phaser 3.

* Nothing, it's a bug fix

Describe the changes below:

Points should be returned up to the circumference of the defined ellipse,  instead, the points returned where within the centre, with a large padding around the ellipse edge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ being passed to the simulation. The default value is 1 to remain consistent with
 * Tween.restart handles removed tweens properly and reads them back into the active queue for the TweenManager (thanks @wtravO)
 * Tween.resume will now call `Tween.play` on a tween that was paused due to its config object, not as a result of having its paused method called. Fix #3452 (thanks @jazen)
 * LoaderPlugin.isReady referenced a constant that no longer exists. Fix #3503 (thanks @Twilrom)
+* Ellipse.random() returned a subset of possible points within the ellipse area. (thanks @budda)
 
 ### Updates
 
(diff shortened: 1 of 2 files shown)


photonstorm:
Before your change:

ellipse1
ellipse2

After your change:

ellipse1-after
ellipse2-after

Test code:

```
var config = {
    width: 800,
    height: 600,
    type: Phaser.AUTO,
    parent: 'phaser-example',
    scene: {
        create: create
    }
};

function create ()
{
    var ellipse = new Phaser.Geom.Ellipse(400, 300, 200, 400);

    var graphics = this.add.graphics({ lineStyle: { width: 2, color: 0x00aaaa }, fillStyle: { color: 0xff0000 }});

    graphics.strokeEllipseShape(ellipse);

    for (var i = 0; i < 2000; i++)
    {
        var p = ellipse.getRandomPoint();

        graphics.fillRect(p.x, p.y, 2, 2);
    }
}
```

budda:
It was a bit of a shot in the dark based on the Phaser CE bug I was submitting that could be seen in https://codepen.io/budda/pen/jzdboZ and didn't have a chance to write some test code earlier.

Issue to close down then.

photonstorm:
Actually, I think that codepen is showing a different issue. The ellipse has been defined with a size of 200x300, but the `graphics.drawEllipse` has drawn it at 400x600. So the random points are correct, it's the outline that looks wrong to me!

PR #3540, opened 2018-04-10
Ensure body is not null as a result of `Impact.World.remove(object)`

iamDecode:
## This PR changes

* Nothing, it's a bug fix

## Describe the changes below:

Calling `Impact.World.remove(body)` to remove an object from the world during an update call would crash on line 637 with error: 
```
Cannot read property 'skipHash' of undefined
```

This happens because `this.bodies.entries` is changed, while `len` in this scope is already defined.
By using forEach rather than C-style loops, the removed (and undefined) body will not end up in the loop, preventing this crash.

## Discussion: 

I ran some crude benchmarks which showed no significant performance loss between for and forEach. However, if this PR still turns out to have an undesired impact on performance, this could be solved by explicit null checks on body instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,7 @@
 * Tween.restart handles removed tweens properly and reads them back into the active queue for the TweenManager (thanks @wtravO)
 * Tween.resume will now call `Tween.play` on a tween that was paused due to its config object, not as a result of having its paused method called. Fix #3452 (thanks @jazen)
 * LoaderPlugin.isReady referenced a constant that no longer exists. Fix #3503 (thanks @Twilrom)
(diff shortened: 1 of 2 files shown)


photonstorm:
Thank you for your PR and explanation. I'm going to go with a simpler `if body` check to avoid a bind and anonymous function creation in a hot loop. I've pushed my change and you've been credited with the fix in the change log.

However, I think this highlights the need for a bigger refactor. Even using a `forEach` approach it's still not going to cope well with significant additions or, or deletions from, the bodies array during the update loop. Other systems in v3 use a Process Queue approach instead, so the main array is _never_ modified during key processing, and I think Impact would benefit from it here too. Will add it to our task list.

PR #3542, opened 2018-04-11
[fixes #3535] Fix constructed functions in RTree

### LINKED ISSUE #3535: Arcade Physics collisions aren't occuring
- <http://labs.phaser.io/edit.html?src=src\physics\arcade\collider%201.js>
- <http://labs.phaser.io/edit.html?src=src\physics\arcade\sprite%20vs%20group.js>

This Issue is about

* A bug in the API: 3.4.0 (12d27518da8975dbcea38569bc7eb9513b7e7c01)

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/56860255-arcade-physics-collisions-aren-t-occuring?utm_campaign=plugin&utm_content=tracker%2F283654&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283654&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jdotrjs:
**Change type:** Bug fix, #3535

The previous patch didn't take into account that the accessors passed in as the format array were dot-prefixed due to the previous eval-based construction.

The only two uses of `RTree` that I found were in `World.js`:

```javascript
this.tree = new RTree(this.maxEntries, [ '.left', '.top', '.right', '.bottom' ]);
// ...
this.staticTree = new RTree(this.maxEntries, [ '.left', '.top', '.right', '.bottom' ]);
```

It's likely that this could be updated to just not pass dotted attribute names but I wasn't super comfortable that they weren't needed in this form elsewhere despite a quick search. I'd honestly say that it might be a better change if we remove the dots before merge but I'll leave that up to the discretion of the reviewers/merger.

I'm not super familiar with Phaser's style but I passed lint sooo :D

Buyer beware: I didn't verify this is still within the twitch CSP (it should be). I tested the following cases using slight variations of the repro code I left in the bug:

1. this.physics.world.collide(sprite, group)
1. this.physics.world.collide(group, sprite)
1. this.physics.world.collide(sprite, sprite)
1. this.physics.world.collide(group, group)

### DIFF
--- a/src/structs/RTree.js
+++ b/src/structs/RTree.js
@@ -474,9 +474,9 @@ rbush.prototype = {
         format = format.map(
             function (f)
             {
-                return f.substring(1)
+                return f.substring(1);
             }
-        )
+        );
 
         // Do not use string-generated Functions for CSP policies
         // Instead a combination of anonymous functions and grabbing properties
@@ -486,12 +486,12 @@ rbush.prototype = {
 
         var mkCompareFn = function(attr) {
           return function(a, b) {
-            return a[attr] - b[attr]
-          }
-        }
+            return a[attr] - b[attr];
+          };
+        };
 
-        this.compareMinX = mkCompareFn(format[0])
-        this.compareMinY = mkCompareFn(format[1])
+        this.compareMinX = mkCompareFn(format[0]);
+        this.compareMinY = mkCompareFn(format[1]);
 
         this.toBBox = function(a)
         {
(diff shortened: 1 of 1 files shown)


photonstorm:
Thank you for this. I see no reason why we need to keep the dot prefix. We can also make the default accessors array be [left, top, right, bottom] one as well. Going to merge this PR for now though, can always merge another one to modify that.

PR #3543, opened 2018-04-11
Add 'destroy' event of sound object

rexrainbow:

* The public-facing API

Describe the changes below:

Add 'destroy' event of sound object. A possible use-case is to free the reference of this destroyed sound object in other object.


photonstorm:
Sure, sounds like a good idea.

PR #3551, opened 2018-04-12
Allow tween onComplete listeners to affect tween state

Cudabear:
Set the state of the tween to PENDING_REMOVE before calling the onComplete listener of the tween to allow the `onComplete` listener to affect the tween state and not be immediately overwritten upon finishing execution.

This is useful for conditionally restarting the tween upon its completion.

Example:

```
        var me = this;
        this.myTween = this.add.tween({
            targets: this.mySprite,
            x: 300,
            duration: 3000,
            onComplete: function(){
                if(someConditional){
                    me.myTween.restart();
                }
            }
        });
```

Before this change, the `restart` call's state change would be overwritten immediately after leaving execution of the on complete.

### DIFF
--- a/README.md
+++ b/README.md
@@ -372,6 +372,7 @@ A special mention must go to @orblazer for their outstanding assistance in helpi
 * PathFollower.resume has been renamed to `resumeFollow` to avoid conflicting with the Animation component.
 * PathFollower.stop has been renamed to `stopFollow` to avoid conflicting with the Animation component.
 * BaseSound.setRate has been renamed to `calculateRate` to avoid confusion over the setting of the sounds rate.
+* Phaser.Tweens.Tween now allows restarting the tween in the `onComplete` listener
 
 Please see the complete [Change Log]((https://github.com/photonstorm/phaser/blob/master/CHANGELOG.md)) for previous releases.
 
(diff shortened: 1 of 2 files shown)


snowbillr:
I think another option may be to change the condition in the [`Tween#restart` method](https://github.com/photonstorm/phaser/blob/master/src/tweens/tween/Tween.js#L412) to check for both `REMOVED` and `PENDING_REMOVE`.

Cudabear:
The problem with changing the state at all inside the `onComplete` listener is the `complete` `nextStep` and `update` functions all have this line:

`this.state = TWEEN_CONST.PENDING_REMOVE` 

which will undo anything you do to change the state.

photonstorm:
I'm going to leave this PR for now as I'll change how tween callbacks work very soon (re: the discussion about it in the issues) - I'll keep this open so I can refer to it then.

Cudabear:
I'll close this because it seems as though this feature request was resolved as of 3.16.2.

Cudabear:
Nevermind, totally lied.  Still seeing this potential improvement in `3.16.2`

photonstorm:
Just to say that I've implemented this feature into 3.19.0 directly. Any time the Tween state changes, it's done before a Tween event or callback is dispatched.

PR #3559, opened 2018-04-13
Add missing letterSpacing to DynamicBitmapText

### LINKED ISSUE #3558: BitmapText renders only first letter in WEBGL and reports wrong bounds
This Issue is about
* A bug in the API 3.4.0

This happens with both Dynamic and ~Static BitmapText~.
The issue is probably just with the bounds calculation and not actual rendering.

EDIT: Seems to be just the Dynamic actually.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/57031951-bitmaptext-renders-only-first-letter-in-webgl-and-reports-wrong-bounds?utm_campaign=plugin&utm_content=tracker%2F283654&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283654&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Antriel:
This PR changes
* Nothing, it's a bug fix

Newly added `letterSpacing` in BitmapText was missing in the Dynamic variant, which broke bounds calculation due to letterSpacing being undefined. Which subsequently broke rendering in WebGL and aligning.
Fixes #3558

### DIFF
--- a/src/gameobjects/bitmaptext/dynamic/DynamicBitmapText.js
+++ b/src/gameobjects/bitmaptext/dynamic/DynamicBitmapText.js
@@ -116,6 +116,16 @@ var DynamicBitmapText = new Class({
          */
         this.fontSize = size || this.fontData.size;
 
+        /**
+         * Adds/Removes spacing between characters
+         * Can be a negative or positive number
+         *
+         * @name Phaser.GameObjects.DynamicBitmapText#letterSpacing
+         * @type {number}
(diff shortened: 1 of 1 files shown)


photonstorm:
I wonder if this ought to be Static only, but I guess it won't hurt to have it in Dynamic too.

PR #3561, opened 2018-04-13
Passing this.engine to MatterEvents.off

pixelscripter:

* Nothing, it's a bug fix

Describe the changes below:

Passing this.engine to MatterEvents.off, because else we get an error in Matter.Events line 47 (Uncaught TypeError: Cannot set property 'events' of undefined).


pixelscripter:
This fix removes the error, but because the this.setEventsProxy() is only called in the World constructor, these events are not bound again.

PR #3580, opened 2018-04-17
Delete Texture only applicable on renderer.gl

jmcriat:
* Fixes a bug

Describe the changes below:

deleteTexture only applicable on renderer.gl, There is no deleteTexture on this.renderer when running on canvas, so destroying tilesprites (scene shutting down) would throw error.


photonstorm:
👍

PR #3605, opened 2018-04-20
Update GetBounds.getBounds() JSDoc

zilbuz:
This PR

* Updates the Documentation

Describe the changes below:

Update GetBounds.getBounds() JSDoc so that @generic matches with @param and @return. It seems that this comment wasn't in sync with the rest of the block. Moreover since this comment is used to generate the TS definitions, I wasn't able to use `getBounds()` in TypeScript since the compiler tried to force a `Vector2` for the result.


photonstorm:
👍

PR #3625, opened 2018-04-27
Fixed loader isn't passed to JSONFile

iamchristopher:
This update aims to fix an error when loading JSON files with `this.load.json` caused by the loader not being passed to `JSONFile`.


photonstorm:
👍

PR #3628, opened 2018-04-28
Fix camera bound typings

PaNaVTEC:
This PR:

* Updates the Documentation / Typings

Describe the changes below:

Bounds where not generated with that syntax so I created an entity to
explictly define the camera bounds.

Bounds are there only when the internal property 'useBounds' is there,
so it can be undefined in some cases.


photonstorm:
👍

PR #3630, opened 2018-04-29
Fix JSDoc parameter type in Actions.GridAlign

ksmai:
This PR
* Updates the Documentation

Describe the changes below:
`cellWidth` should be an integer instead of a boolean, as described by the default value and the description text.


photonstorm:
👍

PR #3635, opened 2018-04-30
arcade colliders take gameobjects, not arcade bodies

snowbillr:

* Updates the Documentation

Describe the changes below:
Adding an Arcade Physics Collider was throwing an error because it expected the passed objects to be Arcade Bodies (which is incorrect since you can pass in groups, arrays, tilemap layers, etc). This PR updates the param types for that method (and a couple of others) to match the param types of the Arcade World's `collideObjects` function, which takes in GameObjects instead.


snowbillr:
Ah hold up a sec, they actually take `Phaser.GameObjects.GameObject | Phaser.GameObjects.GameObject[]`, i'll update the PR in a little bit.

PR #3638, opened 2018-05-01
Change PathFollower base class from Sprite to Container

johwiese:

* Updates the Documentation
* Adds a new feature

Describe the changes below:

This PR enables Phaser to move any game object along a path not only a single sprite. The base class of PathFollower has changed from Sprite to Container. 

Some of the doc-strings my need a rework because of the new base class. Containers can't be animated?

This breaks all old versions using the PathFollower. This will break all PahtFollower phaser3-examples.

The API changed from 
> PathFollower (scene, path, x, y, texture, frame)

To:
> PathFollower (scene, x, y, path, children)

Reason:
Every's GA parameter list starts with scene, x, y. So, why not Pathfollower?
Replaced texture, frame with children

From:
>startFollow: function (config, startAt)
To:

>startFollow: function (config)

Reason: 
startAt was not used.

From:
 >preUpdate: function (time, delta)

To:
>preUpdate: function ()

Reason:
not used any more 

Example 
```
var config = {
    type: Phaser.AUTO,
    width: 800,
    height: 600,
    backgroundColor: '#2d2d2d',
    parent: 'phaser-example',
    scene: {
        preload: preload,
        create: create
    }
};

function preload ()
{
    this.load.image('ball', 'assets/sprites/shinyball.png');
}

function create ()
{
    var points = [ 50, 400, 200, 200, 350, 300, 500, 500, 700, 400 ];

    var curve = new Phaser.Curves.Spline(points);

    var graphics = this.add.graphics();

    graphics.lineStyle(1, 0xffffff, 1);

    curve.draw(graphics, 64);

    graphics.fillStyle(0x00ff00, 1);

    for (var i = 0; i < points.length; i++)
    {
        graphics.fillCircle(points[i].x, points[i].y, 4);
    }
    var follower = this.add.follower(50, 400, curve);

    for(var j = 0; j < 4; j++){
        var ball = this.add.sprite(0, 50 * j , 'ball');
        follower.add(ball);  
    }

      follower.startFollow({
        duration: 5000,
        yoyo: true,
        repeat: -1,
        rotateToPath: true,
        rotationOffset: 90
    });
}
```


hexus:
I like how minimal yet impactful this change is.

I agree that Path Followers shouldn't be limited to Sprites. I'm not sure whether it should be bound to being a Container, but that certainly is _one_ way to enable any Game Object to be a Path Follower.

rexrainbow:
How about adding a parameter to change the default base class?

src/gameobjects/pathfollower/PathFollower.js
```diff
@@ -234,10 +232,9 @@ var PathFollower = new Class({
      *
      * @return {Phaser.GameObjects.PathFollower} This Game Object.
      */
-    startFollow: function (config, startAt)
+    startFollow: function (config)
```
hexus: It'd be worth removing this from the JSDoc also.

hexus:
> How about adding a parameter to change the default base class?

My previous comment was likely in the hopes that there would be a way to extend `GameObject`, or even [decorate](https://en.wikipedia.org/wiki/Decorator_pattern) it, but JS makes that kind of tricky, because `GameObject`s have all kinds of methods that we can't magically forward to.

As such, a `Container` is the best approach I can think of at the moment.

Once the JSDoc has updated according to your API changes we could look at merging this. :+1:

hexus:
Having reviewed this a little further, I think `PathFollower` could be simplified to use augmentation similar to the way [`MatterGameObject`](https://github.com/photonstorm/phaser/blob/15c4334/src/physics/matter-js/MatterGameObject.js) works.

This would involve moving `PathFollower` properties and methods to a Game Object Component class instead. I'm going to look into implementing this.

johwiese:
Yes, on the js side, this would probably be a better way to achieve this.

But on the ts side, I have concerns about the definition. As you can see,
the MatterGameObject strips all typescript definitions down to a plain
GameObject. You throw in a Sprite and get a GameObject in return. Wouldn't
it be better to return an interface reflecting the added features?

Am 16.08.2018 2:17 nachm. schrieb "Chris Andrew" <notifications@github.com>:

Having reviewed this a little further, I think PathFollower could be
simplified to use augmentation similar to the way MatterGameObject
<https://github.com/photonstorm/phaser/blob/15c4334/src/physics/matter-js/MatterGameObject.js>
works.

This would involve moving PathFollower properties and methods to a Game
Object Component class instead. I'm going to look into implementing this.

—
<https://github.com/photonstorm/phaser/pull/3638#issuecomment-413525668>,
or mute the thread

hexus:
See #3936 for a demonstration of what I had in mind. It's not quite finished, but gets the idea across.

hexus:
> Wouldn't it be better to return an interface reflecting the added features?

To elaborate on my previous comment, I totally agree here. The problem is that the provided Game Object could be just that: any type of Game Object. I would assume there isn't a way to dynamically map return types based on given parameter types in TypeScript.

Another option would be making a mixin type def for every type of Game Object: `{PathFollowerSprite|PathFollowerContainer|PathFollower*}`. But that seems a bit unfriendly for custom Game Objects.

The true solution to this kind of problem is an [Entity Component System](https://en.wikipedia.org/wiki/Entity%E2%80%93component%E2%80%93system), but Phaser isn't designed this way.

photonstorm:
Thank you for taking the time to open this PR, however, I'm going to close it off. I've since moved all PathFollower properties and methods to the new `PathFollower` Game Object component, which can be applied to any custom Game Object now.

PR #3641, opened 2018-05-02
Update Tiled Base64Decode function

tamagokun:

* Fixes a bug

Describe the changes below:

The array length should be 1/4 the size of the raw binary string, otherwise you end up with a bunch of undefined array values at the end.


photonstorm:
👍

PR #3647, opened 2018-05-08
Update JSDocs

zilbuz:
This PR:
* Updates the Documentation

Describe the changes below:
- Update GameConfig to explicit the `render` subobject
- GameConfig#render.powerPreference is documented as string instead of boolean
- Add comments to the members of the `Config` object

I used `@const` for the members of `Config` so that they are marked "readonly" in the generated TS definitions and the comment is still only one line. Otherwise I have to do: 
```js
/**
 * @readonly
 * @member {type} name
 */
```

If you prefer the longer version or a version similar to the properties of `Game` with `@name`, `@type`, `@readonly` and `@since`, I can update my PR.

### DIFF
--- a/src/boot/Config.js
+++ b/src/boot/Config.js
@@ -79,16 +79,16 @@ var ValueToColor = require('../display/color/ValueToColor');
  * @property {string} [banner.text='#ffffff'] - [description]
  * @property {string[]} [banner.background] - [description]
  * @property {FPSConfig} [fps] - [description]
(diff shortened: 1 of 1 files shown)


photonstorm:
The issue is the spacing. If you update them to this:

```
/**
 * @const {(integer|string)} Phaser.Boot.Config#width - [description]
 */
...
```

and make sure there is an empty line before the next docblock then it'll be fine.

PR #3650, opened 2018-05-09
Missing loader parameter in File.call

TadejZupancic:

* Fixes a bug

Describe the changes below:

File needs to be called with 2 parameters. First parameter - loader was missing.


photonstorm:
Thanks, it was also missing in PluginFile too, which is now fixed as a result of this PR.

PR #3659, opened 2018-05-12
Correct JSDoc type for GameObject#body

samme:
This PR

* Updates the Documentation/TypeScript

### DIFF
--- a/src/gameobjects/GameObject.js
+++ b/src/gameobjects/GameObject.js
@@ -148,7 +148,7 @@ var GameObject = new Class({
          * If this Game Object is enabled for physics then this property will contain a reference to a Physics Body.
          *
          * @name Phaser.GameObjects.GameObject#body
-         * @type {?object}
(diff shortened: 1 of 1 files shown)


src/gameobjects/GameObject.js
```diff
@@ -148,7 +148,7 @@ var GameObject = new Class({
          * If this Game Object is enabled for physics then this property will contain a reference to a Physics Body.
          *
          * @name Phaser.GameObjects.GameObject#body
-         * @type {?object}
+         * @type {?object|Phaser.Physics.Arcade.Body|Phaser.Physics.Impact.Body}
```
photonstorm: Pretty sure it needs to be {?(object|Phaser.Physics.Arcade.Body|Phaser.Physics.Impact.Body)}`

samme: Fixed.

PR #3677, opened 2018-05-18
Update Config.js

GunvantSr:
Changed variables from var to const!


photonstorm:
Thank you for your PR, but I can’t merge this in as it will break all non-ES6 browsers, which are sadly a not insignificant number.

Issue #3651, opened 2018-05-09
Restarting tween right after initialization makes it stuck


Antriel:
Example code:
```js
var tween = this.tweens.add({
    targets: { v:0 },
    v: 100,
    onComplete: function() { console.log("never fires"); }
});
tween.restart();
```

This causes the tween to have a `TWEEN_CONST.ACTIVE` state, but it's still in `TweenManager`'s `_add` because it checks for state of `TWEEN_CONST.PENDING_ADD`.
A possible fix would be not restarting tween that has such state or changing the check in `TweenManager`.

zacharysarette:
I just created a PR for this here #3680 .

It does look like it checks the state in the restart function to see if it is TWEEN_CONST.REMOVED but doesn't check if it is TWEEN_CONST.PENDING_ADD.

I added it to the if statement and tested it. It works. 

We'll see if it gets pulled in or not. If someone has a better idea on how to fix the issue I would love to see how the solution. Any advice would be appreciated since I am new to contributing to Phaser.js. Cheers!

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

PR #3680, opened 2018-05-20
This fixes an issue with the tween restart function in phaser.js.

zacharysarette:
This PR 

* Fixes a bug #3651 

Describe the changes below:

It now checks to see if this.state === TWEEN_CONST.PENDING_ADD in the tween restart function

If true, sets this.seek(0) and this.parent.makeActive(this)

This fixes the issue that makes the tween stuck when restarting the new tween directly after adding a the new tween.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -184,6 +184,69 @@
         "long": "3.2.0"
       }
     },
+    "@webpack-contrib/config-loader": {
+      "version": "1.1.1",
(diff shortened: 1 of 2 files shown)


photonstorm:
Thanks for this PR. Could you please remove your changes to package.json and the lock file, so I can merge it.

zacharysarette:
Okay, I think I just put a new pull request in.

PR #3682, opened 2018-05-20
Add `allowDrag`, `allowGravity`, `allowRotation` to Arcade.PhysicsGroup config

samme:
This PR

* Adds a new feature

[To save looping](http://www.html5gamedevs.com/topic/37529-overlap-between-sprite-and-staticgroup-and-not-working/?tab=comments#comment-214347).


photonstorm:
Nice one. I'm going to tidy up the Body methods slightly, but otherwise this is sound.

PR #3687, opened 2018-05-22
Update type definitions

thomastanck:
* Updates the Documentation

Describe the changes below:

Fixes param types for
* Phaser.Physics.Arcade.Components.Bounce#setCollideWorldBounds (number -> boolean)
* Phaser.Physics.Arcade collide/overlap functions (Body -> Body|Group)
* Phaser.Physics.Arcade.Factory staticGroup/group config object (array -> object|object[])
* Phaser.GameObjects.Group.* (return type GameObject -> any)
* Phaser.GameObjects.Components.* (return type `this` when returning this)

Fixes member types for
* Phaser.Physics.Arcade.Sprite.body (inherit Phaser.GameObjects.Sprite.body type -> specifying in ArcadeSprite.js)

The last one is iffy.

Feel free to revert any of my changes without my input since types are pretty subjective in most of these cases.

### DIFF
--- a/src/gameobjects/components/Alpha.js
+++ b/src/gameobjects/components/Alpha.js
@@ -81,7 +81,7 @@ var Alpha = {
      * @method Phaser.GameObjects.Components.Alpha#clearAlpha
      * @since 3.0.0
      * 
-     * @return {Phaser.GameObjects.GameObject} This Game Object instance.
+     * @return {this} This Game Object instance.
      */
     clearAlpha: function ()
     {
--- a/src/gameobjects/components/Alpha.js
+++ b/src/gameobjects/components/Alpha.js
@@ -103,7 +103,7 @@ var Alpha = {
      * @param {float} [bottomLeft] - The alpha value used for the bottom-left of the Game Object. WebGL only.
      * @param {float} [bottomRight] - The alpha value used for the bottom-right of the Game Object. WebGL only.
      * 
-     * @return {Phaser.GameObjects.GameObject} This Game Object instance.
+     * @return {this} This Game Object instance.
      */
     setAlpha: function (topLeft, topRight, bottomLeft, bottomRight)
     {
--- a/src/gameobjects/components/BlendMode.js
+++ b/src/gameobjects/components/BlendMode.js
@@ -100,7 +100,7 @@ var BlendMode = {
      *
      * @param {(string|Phaser.BlendModes)} value - The BlendMode value. Either a string or a CONST.
      *
-     * @return {Phaser.GameObjects.GameObject} This Game Object instance.
+     * @return {this} This Game Object instance.
      */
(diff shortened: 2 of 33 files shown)


thomastanck:
I believe the very last commit is a very good idea, it makes the following code work
```typescript
		let platforms = this.physics.add.staticGroup();
		(<Phaser.Physics.Arcade.Sprite>platforms.create(400, 568, 'ground')).setScale(2).refreshBody();
```
because although `.setScale(` doesn't have any idea it's actually a `Physics.Arcade.Sprite` (`.setScale)` is defined under `GameObjects.Components`), its return type annotation is `this`, and so typescript infers that the return type is actually a `Physics.Arcade.Sprite`.

Without this solution (e.g. having its return type be `GameObject.Sprite` for example), `.refreshBody(` would fail, so you would need the following uglier code:
```typescript
		let platforms = this.physics.add.staticGroup();
		(<Phaser.Physics.Arcade.Sprite>(<Phaser.Physics.Arcade.Sprite>platforms.create(400, 568, 'ground')).setScale(2)).refreshBody();
```

photonstorm:
Well, this sucks. I tried swapping a few component methods to `returns {this}` to see what happens, and JSDoc just doesn't care. It treats it as a reserved word. The output doesn't contain the correct data type at all :(

I can see this was requested as a feature https://github.com/jsdoc3/jsdoc/issues/991 but was closed almost immediately, so it appears it's something that TypeScript can do, but JSDoc itself can't. Not entirely sure what to do. I don't want the docs to suffer just for the benefit of TS devs. Will have a look to see if we could implement this in our own jsdoc plugin, but I cannot merge this until we've solved it :(

thomastanck:
Hmm, I’ve read through the issue and they seem to propose having JSdoc convert return this into whatever type is being displayed (eg Type A), instead of the current behaviour of simply showing “Type this”. Personally I like the current behaviour (Type this) more, but oh well.

Well as I said, feel free to revert any changes you don’t like. I’m on mobile right now so I cant revert that change yet. What about the other commits?

photonstorm:
Outside of TypeScripts, IDEs like Webstorm use the jsdoc types to provide intellisense, so unless an actual type is returned they would break, which isn't a change I'm comfortable with. I don't think it would take much to create a plugin for this, will have a look at it.

photonstorm:
After several hours of wrangling with jsdoc I managed to create a plugin that will convert `return {this}` into the relevant data type in the generated docs, while keeping `{this}` in the source code, which should solve both issues. The plugin is in the docs repo and appears to work fine, so I will review this PR again tomorrow and it should be good to merge in as-is.

samme:
> Phaser.GameObjects.Group.* (return type GameObject -> any)

Can I revert this?

I think it must be `GameObject` or a subclass.

PR #3689, opened 2018-05-22
Add offset from follow target to Camera

hermbit:
This PR

* Adds a new feature

Adds a "followOffset" object to the Camera, which, if set, will offset the position of the camera relative to its follow target.

### DIFF
--- a/src/cameras/2d/Camera.js
+++ b/src/cameras/2d/Camera.js
@@ -310,6 +310,16 @@ var Camera = new Class({
          * @since 3.0.0
          */
         this._follow = null;
+        
+        /**
+         * Offset position from follow target.
+         *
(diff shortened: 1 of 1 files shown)


photonstorm:
Thank you for submitting this PR. Although I'm not going to merge it, the feature did get implemented into the Camera class, so `followOffset` is now a property you can use, and we credited it to you in the Change Log.

hermbit:
Awesome, that's great to hear! 

Is there anything that I did wrong in my PR that I can improve on in the future? I'm always happy to contribute so I want to make sure I'm doing it right.

photonstorm:
It turned out to be quicker to just have it as a zeroed vec2 by default than to add in 2 ternaries in the main loop. That's about the only difference between what I implemented and your original.

PR #3693, opened 2018-05-23
Fixes tweens getting stuck after being restarted in bug #3651

zacharysarette:
This PR 

Fixes a bug #3651
Describe the changes below:
The tween animation was getting stuck when calling restart on it after being added.

It now checks to see if this.state === TWEEN_CONST.PENDING_ADD in the tween restart function

If true, sets this.seek(0) and this.parent.makeActive(this)

This fixes the issue that makes the tween stuck when restarting the new tween directly after adding a the new tween.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -184,69 +184,6 @@
         "long": "3.2.0"
       }
     },
-    "@webpack-contrib/config-loader": {
-      "version": "1.1.1",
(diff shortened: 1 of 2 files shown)


photonstorm:
The fail isn't to do with your PR, so you can ignore it. If you view the details you'll see it's elsewhere in the API.

PR #3703, opened 2018-05-26
Deleted `console.log`.

rblopes:


hexus:
Thanks!

For your information, the build error is unrelated to this change.

PR #3712, opened 2018-05-28
Fixed endless loop in `TweenManager.getTweensOf()`

cyantree:
This PR

* Fixes a bug

Describe the changes below:

Passing mutiple targets to `TweenManager.getTweensOf()` resulted in an endless loop as the inner loop never finished.


photonstorm:
👍

PR #3718, opened 2018-05-30
Particle Emitter Manager in Container

TadejZupancic:

* Fixes a bug

Describe the changes below:

Added Alpha and ScrollFactor Mixin to Particle Emitter Manager otherwise it crashes when added to the Container.


photonstorm:
Thanks for this. I've merged it in and emitters now work inside containers, but it's also added a bunch of methods (like `setAlpha`) which don't do anything, so I think I'll tweak it to resolve that (you'll still get credit in the change log though)

PR #3723, opened 2018-05-31
Don't re-add a user-specified Canvas to DOM

NQNStudios:
This PR fixes a bug.

## Changes

While it was possible to pass an existing canvas element into the Game constructor, the AddToDom() method was moving the element onto document.body regardless of whether it already had a parent. This was causing my canvas to be repositioned unexpectedly, defeating the purpose of passing my own in (custom positioning of the canvas among other HTML elements).

Now, if GameConfig.parent is undefined but the canvas already has one, it won't be moved.


photonstorm:
Happy to merge this, but please remove your changes to gitignore and package-lock first, thanks.

NQNStudios:
That's done. Thanks!

PR #3729, opened 2018-06-04
Matterjs pe loader

CodeAndWeb:
This PR adds a new feature:

Load complex physics shapes (polygon, circle, compounds) from a shape definition file created with PhysicsEditor.

### DIFF
--- a/src/physics/matter-js/PhysicsEditorLoader.js
+++ b/src/physics/matter-js/PhysicsEditorLoader.js
@@ -17,19 +17,22 @@ var PhysicsEditorLoader = {
 
     loadBody: function (x, y, w, h, config)
     {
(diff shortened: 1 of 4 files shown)


photonstorm:
Sorry, but I can’t merge this because it contains ES6 code and will break backward compatibility. Phaser 3 is currently 100% ES5 so it builds without the need of a transpiler. All merged code must follow that structure too.

CodeAndWeb:
@photonstorm Any update?

photonstorm:
2 small changes:

1) Can you rename it `PhysicsEditorParser` please, because it doesn't actually load anything (and sounds like it should be part of the Loader class currently)

2) Can you please add in jsdocs. There are only a handful of functions so it shouldn't take long. You can copy the docblock from another file to get the correct structure.

PR #3736, opened 2018-06-07
Tweens can now have duration 0

TadejZupancic:
* Fixes a bug

Describe the changes below:

If the tween duration was 0, it crashed because of division by zero when calculating progress. Now it sets progress to 1 in that case.

### DIFF
--- a/src/tweens/tween/Tween.js
+++ b/src/tweens/tween/Tween.js
@@ -1214,7 +1214,7 @@ var Tween = new Class({
                 }
 
                 var forward = (tweenData.state === TWEEN_CONST.PLAYING_FORWARD);
-                var progress = elapsed / duration;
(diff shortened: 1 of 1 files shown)


photonstorm:
Why would the duration ever be zero unless the dev had made a configuration error when creating it? 

Surely this is something we can trap during the init of the tween, rather than checking it every single frame.

TadejZupancic:
It could be used to just set the values at the time the tween is supposed to be executed. It's more elegant than writing a callback function every time you want to just set some values. Greensock has a set method that can be used for that, but I didn't find something like that in phaser.

photonstorm:
I still there must be a cleaner way of doing it than checking for such an edge-case every single frame, for every single tween.

TadejZupancic:
Yes, I agree. I'll think of some other way of doing it.

PR #3747, opened 2018-06-13
[fix] sceneManager .run will queueOp if scene not ready

rook2pawn:
This PR 

* Fixes a bug

Fixes `.run` in game sceneManager to `queueOp` an unready scene. Since we don't yet know if the key is valid, we look for it first in the `_pending` queue, unlike in the `scenePlugin` where we already know it is valid.

### DIFF
--- a/src/scene/SceneManager.js
+++ b/src/scene/SceneManager.js
@@ -140,7 +140,7 @@ var SceneManager = new Class({
                 });
             }
         }
-        
+
         game.events.once('ready', this.bootQueue, this);
     },
 
@@ -1011,7 +1011,7 @@ var SceneManager = new Class({
 
     /**
      * Runs the given Scene, but does not change the state of this Scene.
-     * 
+     *
      * If the given Scene is paused, it will resume it. If sleeping, it will wake it.
      * If not running at all, it will be started.
      *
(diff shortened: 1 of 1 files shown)


src/scene/SceneManager.js
```diff
@@ -1032,6 +1032,19 @@ var SceneManager = new Class({
...
+            {
+                if (this._pending[i].key === key)
+                {
+                    isFound = true;
+                    break;
+                }
+            }
+            if (isFound)
+            {
+                this.queueOp('start', key, data);
```
hexus: This line could be moved in place of line 1040, `isFound = true;`. Avoids the extra conditional and variable. :)

rook2pawn: thanks! good catch

src/scene/SceneManager.js
```diff
@@ -1032,6 +1032,14 @@ var SceneManager = new Class({
 
         if (!scene)
         {
+            for (var i = 0; i < this._pending.length; i++)
+            {
+                if (this._pending[i].key === key)
+                {
+                    this.queueOp('start', key, data);
```
hexus: Nice!

PR #3748, opened 2018-06-14
[feat] scene.run can now pass data to .wake and .resume if it needs

rook2pawn:

* Adds a new feature

Describe the changes below:

scene.run can now pass data to .wake and .resume if it needs to invoke those methods from .run. This gives the ability to listen i.e. `on.wake(ref, data)`

### DIFF
--- a/src/scene/SceneManager.js
+++ b/src/scene/SceneManager.js
@@ -140,7 +140,7 @@ var SceneManager = new Class({
                 });
             }
         }
-        
+
         game.events.once('ready', this.bootQueue, this);
     },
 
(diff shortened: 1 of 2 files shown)


hexus:
Might worth updating the `data` parameter description too. This change goes against that description. :)

photonstorm:
I'll merge this in, but I think if we're going to allow this, it needs to be done properly. As it stands with this PR you can _only_ pass data when waking or resuming a Scene if you use `run`, but not if you use the normal approach (i.e. `this.scene.wake()`). which means the Systems is now dispatching an event with `undefined` as the 2nd argument most of the time. I'll have a play with it now.

PR #3749, opened 2018-06-14
Fixed DataManager changedata event emits original value instead of new value

iamchristopher:
* Fixes a bug

Describe the changes below:

The DataManager `changedata` event currently emits the originally set value instead of the newly assigned one.


photonstorm:
👍

PR #3755, opened 2018-06-15
Add rounded rectangle to Graphics

TadejZupancic:
* Adds a new feature

Describe the changes below:

Added fillRoundedRect and strokeRoundedRect functions to Phaser.GameObjects.Graphics, that draw a filled/stroked rounded rectangle.

Got some code for different radii per corner from [here](https://stackoverflow.com/a/3368118/5299434)

### DIFF
--- a/src/gameobjects/graphics/Graphics.js
+++ b/src/gameobjects/graphics/Graphics.js
@@ -549,6 +549,100 @@ var Graphics = new Class({
     },
 
     /**
+     * Fill a rounded rectangle with the given position, size and radius.
+     *
+     * @method Phaser.GameObjects.Graphics#fillRoundedRect
+     * @since 3.11.0
+     *
+     * @param {number} x - The x coordinate of the top-left of the rectangle.
+     * @param {number} y - The y coordinate of the top-left of the rectangle.
+     * @param {number} width - The width of the rectangle.
+     * @param {number} height - The height of the rectangle.
+     * @param {number} [radius = 20] - The corner radius; It can also be an object to specify different radii for corners
+     * @param {number} [radius.tl = 0] Top left
+     * @param {number} [radius.tr = 0] Top right
+     * @param {number} [radius.br = 0] Bottom right
+     * @param {number} [radius.bl = 0] Bottom left
+     *
+     * @return {Phaser.GameObjects.Graphics} This Game Object.
+     */
+    fillRoundedRect: function (x, y, width, height, radius)
+    {
+        if (typeof radius === 'number') {
+            radius = {tl: radius, tr: radius, br: radius, bl: radius};
+        } else if (typeof radius === 'object') {
+            var defaultRadius = {tl: 0, tr: 0, br: 0, bl: 0};
+            for (var side in defaultRadius) {
+                radius[side] = radius[side] || defaultRadius[side];
+            }
+        }
+        else {
+            radius = {tl: 20, tr: 20, br: 20, bl: 20};
+        }
+
+        this.beginPath();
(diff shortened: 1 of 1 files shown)


src/gameobjects/graphics/Graphics.js
```diff
@@ -548,6 +548,110 @@ var Graphics = new Class({
...
+     */
+    fillRoundedRect: function (x, y, width, height, radius)
+    {
+        if (typeof radius === 'number')
+        {
+            radius = {tl: radius, tr: radius, br: radius, bl: radius};
+        }
+        else if (typeof radius === 'object')
+        {
+            var defaultRadius = {tl: 0, tr: 0, br: 0, bl: 0};
```
iamchristopher: Wondering if a loop is unnecessary overhead here? Direct assignment may be better: `radius.tl = radius.tl || 0`

src/gameobjects/graphics/Graphics.js
```diff
@@ -548,6 +548,110 @@ var Graphics = new Class({
...
+            for (var side in defaultRadius)
+            {
+                radius[side] = radius[side] || defaultRadius[side];
+            }
+        }
+        else
+        {
+            radius = {tl: 20, tr: 20, br: 20, bl: 20};
```
iamchristopher: The above condition defines different default values than this. Perhaps one default should be define and referenced in each case?

src/gameobjects/graphics/Graphics.js
TadejZupancic: Yes, that would probably be better.

src/gameobjects/graphics/Graphics.js
TadejZupancic: I was thinking and changing it one way and another before making a pull request and then decided it for this way. It is a bit confusing, but I think that if you want rounded rectangle, then the corners should be rounded if you omit the parameter. But if you specify a parameter for at least one of the corners, then you have already set it as rounded and maybe you don't want others to be rounded - in that case you don't have to set all other parameters to 0. If it's too confusing, then we can make the default 20 in any case.

iamchristopher: That's a valid point, I think it makes sense to keep it the way you have it then :)

However, you may want to consider adding the default radius in the constructor and then update `setDefaultStyles` to accept an override.

TadejZupancic: setDefaultStyles sets fill and stroke style that each Graphics object uses. I don't think default radius for one specific Graphics type should be there. Otherwise Ellipse's smoothness would probably also be there already.

photonstorm:
This is a lovely feature, thank you. My only comment is that I find it confusing that if you don't provide a radius, it defaults to 20, but if you provide a blank radius object, it defaults to zero. I'll merge it anyway and fix it here :)

photonstorm:
@TadejZupancic  Just wondering, but any idea why this doesn't render correctly in canvas? It _almost_ works, see here: http://labs.phaser.io/view.html?force=Canvas&v=dev&src=src/game%20objects/graphics/fill%20rounded%20rectangle.js

TadejZupancic:
It works in 3.11, but not in dev - maybe it has something to do with the graphics pipeline change? I'll see if I can find something.

TadejZupancic:
Please check Graphics.arc function. endAngle is changed there and that's why it is drawn incorrectly. I think `endAngle -= startAngle;` line is there my mistake and should be deleted.

photonstorm:
Yeah it's the overshoot calculation causing it again - thanks, will fix.

Issue #3759, opened 2018-06-16
.setInteractive() has a bad/confusing return


t1gu1:
It could be nice if you could change the return of setInteractive() like the other methods to chain it all.

**E.G**:
**I can do:** object.setX(x).setY(y).setInteractive();
**I can't do:** object.setInteractive().setX(x).setY(y)

This is cause by the fact that the setInteractive() method doesn't return the same thing. (it's just confusing for nothing)

Thank for your works!
We makes something nice with your phaser engine.
If you want to test it :package:  -> [https://greatcoffeebreak.com/](https://greatcoffeebreak.com/)

Try to complete it haha. it's hard to know what to do but it's desired to be like that

photonstorm:
That method is absolutely chainable: https://github.com/photonstorm/phaser/blob/master/src/gameobjects/GameObject.js#L356

t1gu1:
I use typescript. Maybe it's in the typing file?

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

PR #3762, opened 2018-06-17
Fix GameObject typings when returning itself

dvdbrink:

* Fixes a bug

Describe the changes below:

I noticed the same problem described in #3759. I believe this should fix the return type for GameObjects that extend the base GameObject.


photonstorm:
👍

PR #3782, opened 2018-06-23
Account for position of tilemapLayer when culling tiles

Upperfoot:

* Fixes a bug

Describe the changes below:

This fixes the problem of culling tiles when we change the tilemapLayer position, without this hotfix tiles will not appear if the position of the tilemapLayer is not the default 0,0 position.


photonstorm:
👍

PR #3791, opened 2018-06-27
Line#getPointB documentation typo fix: "start" is used where "end" was intended

mikewesthad:
Simple doc typo fix


photonstorm:
👍

PR #3816, opened 2018-07-06
Fix Tiled obj property transfer onto spriteConfig

georgzoeller:
* Fixes a bug

Describe the changes below:

_Observed:_ 

Tiled properties are an array of objects (e.g. properties":[{"name":"ObjId","type":"int","value":335},{"name":"frame","type":"int","value":1256}]}) rather than a simple JSON object that the code currently assumed.

Because the function extends spriteConfig with the tiled object.properties object, spriteConfig gains a bunch of index properties ([0],[1],[2],) based on the array rather than the expected object keys and values. This prevents setting for .example frame for the spriteConfig from Tiled directly.

_Fix:_
Transfer tiled object properties onto spriteConfig correctly using the name and value properties on the object

### DIFF
--- a/src/tilemaps/Tilemap.js
+++ b/src/tilemaps/Tilemap.js
@@ -543,7 +543,13 @@ var Tilemap = new Class({
 
             if (found)
             {
-                var config = Extend({}, spriteConfig, obj.properties);
+                var props = {}
+                for (var j=0; j < obj.properties.length; j++)
+                {
+                    props[obj.properties[j].name] = obj.properties[j].value
+                }
+
+                var config = Extend({}, spriteConfig, props);
 
                 config.x = obj.x;
                 config.y = obj.y;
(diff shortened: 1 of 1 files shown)


photonstorm:
Thanks for the PR. This will break maps made in a slightly older version of Tiled though, where the tile properties are _not_ in an array. See below for example:

```
         "tileproperties":
            {
             "135":
                {
                 "bonus":"100"
                },
             "29":
                {
                 "start":"true"
                },
             "72":
                {
                 "speed":"200"
                },
             "99":
                {
                 "goal":"true"
                }
            },
```

I can't merge this PR because I don't want to break it working with older maps, but I will merge it if you can add in a conditional to ensure it only takes the new approach if an array is found, otherwise it defaults to the current method.

georgzoeller:
Will do.

photonstorm:
Can you please check to see if this isn't related to #3832 ? As it sounds like it may have just been a Tiled foobar which they since fixed.

georgzoeller:
Yea they now reverted the change. It loses the data type but for Javacript that’s not much of an issue.

Withdrawing the PR

PR #3821, opened 2018-07-08
Adding console warning to notify users of invalid frame config

andygroff:
* Fixes a bug

Describe the changes below:
It doesn't exactly fix a bug, but hopefully helps users discover their bugs. I have several times now set a frame height that was taller than my horizontal spritesheet. This results in zero frames and when you try to use the animation you get a console error that's pretty tough to track down:
image

I added a console warning to notify users of this, in the hopes that they will waste less time than I have tracking down why their animation won't play. I also considered forcing the row or column to 1 when they're at zero so that users would see their animation and realize the settings were off, but I wasn't sure if there might be some valid use case to having zero frames, and didn't want to break anything.

### DIFF
--- a/src/textures/parsers/SpriteSheet.js
+++ b/src/textures/parsers/SpriteSheet.js
@@ -58,9 +58,9 @@ var SpriteSheet = function (texture, sourceIndex, x, y, width, height, config)
     var column = Math.floor((height - margin + spacing) / (frameHeight + spacing));
     var total = row * column;
 
-    if(total === 0)
+    if (total === 0)
     {
-        console.warn('TextureManager.SpriteSheet: Frame config produces zero frames. Check frameWidth and frameHeight.');
+        console.warn('SpriteSheet frame dimensions will result in zero frames.');
     }
 
     if (startFrame > total || startFrame < -total)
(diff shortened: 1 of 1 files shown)


src/textures/parsers/SpriteSheet.js
```diff
@@ -58,6 +58,11 @@ var SpriteSheet = function (texture, sourceIndex, x, y, width, height, config)
     var column = Math.floor((height - margin + spacing) / (frameHeight + spacing));
     var total = row * column;
 
+    if(total === 0)
```
photonstorm: Can you add a space here please (to keep it consistent with the rest of the codebase)

src/textures/parsers/SpriteSheet.js
```diff
@@ -58,6 +58,11 @@ var SpriteSheet = function (texture, sourceIndex, x, y, width, height, config)
...
+    {
+        console.warn('TextureManager.SpriteSheet: Frame config produces zero frames. Check frameWidth and frameHeight.');
```
photonstorm: I'm happy with this addition, but I'd like to make the warning shorter as it doesn't get minified in the production bundle. Perhaps we could convey the same message with fewer words?

andygroff:
Thanks for the feedback. I've updated the code.

PR #3830, opened 2018-07-14
Update AudioFile.js

kelostrada:
This PR Fixes a bug.

Describe the changes below:

When variable `e` is undefined or null then whole game won't start. In my case some bigger mp3 file on safari cannot load (which is another issue) and javascript breaks because of that. This is a small change which at least allows the error to be shown and the game to be started anyway.


photonstorm:
👍

Rickrod82:
e

PR #3834, opened 2018-07-14
Major Optimization for Culling Tilemaps

tarsupin:
This PR

* Optimizes an existing feature.

Describe the changes below:

Refine the loop to only scan through the drawn boundaries that are being tested against (except when skipping cull). This eliminates a huge bottleneck / lookup time from the original loop each frame.

The update should affect tilemaps exponentially based on their size (and # of layers). It reduced my culling time to 1/70th for 1000x1000 tilemaps, and by about 1/20th on 250x250 tilemaps. It also reduced my requestAnimationFrame time from over 16ms per frame to ~1ms for the larger map.

### DIFF
--- a/src/tilemaps/components/CullTiles.js
+++ b/src/tilemaps/components/CullTiles.js
@@ -27,10 +27,10 @@ var CullTiles = function (layer, camera, outputArray)
 
     outputArray.length = 0;
     
-	var y = 0;
-	var x = 0;
-	var tile = null;
-	
+    var y = 0;
+    var x = 0;
+    var tile = null;
+    
     var tilemapLayer = layer.tilemapLayer;
 
     var tileW = Math.floor(layer.tileWidth * tilemapLayer.scaleX);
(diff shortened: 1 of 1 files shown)


photonstorm:
Very nice PR. The ternaries style ought to match the rest of the API, but I'll make that change here as there's no ESLint rule for it.

photonstorm:
Shouldn't the `drawLeft` (etc) values use the `tileW/H` values, i.e. the _scaled_ tile size? For example:

```
drawLeft = Math.max(0, boundsLeft / layer.tileWidth);
```

`boundsLeft` was calculated using the scaled tile size, but it's being divided by an un-scaled tile size.

tarsupin:
Ah, thanks, I'm still not familiar with the styles used here. Had to redesign a few times just to pass the checker.

And you're correct about the scaled tiles, I had overlooked that. Do you want me to open another PR, or are you modifying that?

photonstorm:
I've updated it. Could you please give it a go and check it still works for your maps?

photonstorm:
I'm also wondering if the `tilesTotal` value ought to be the current one, i.e. the map w*h, or should it be the total of all tiles tested by the for loops (and `tilesDrawn` is just those that get rendered)

tarsupin:
Yep, I'll test it out and confirm everything is usable. I'm in the middle of refactoring and some things are broken, so it might be later today, but I'll get to it.

I had assumed tilesTotal was meant for full map w*h, but its definition is also a little vague. "The total number of tiles in this layer. Updated every frame."

So either the definition needs to be updated or the value does.

tarsupin:
Nevermind, I misunderstood the definition. The value updates every frame, it's not using that to indicate the number that's being updated/scanned every frame. So the definition makes sense as it is.

tarsupin:
Yep, the tilemaps I'm using are still working great. Performance on my standard 500x450 map (with two layers) indicates the Culling function at 0.31% compared to 30.64% at for 3.10

PR #3847, opened 2018-07-19
Major Optimization for Tilemap Collision Indexing

tarsupin:
This PR optimizes an existing feature (Tilemap Collision Indexing).

This optimization will apply to [nearly] everyone using procedural generation tilemaps, infinite tilemaps, multiplayer tilemaps, particularly large tilemaps (especially those dyanmic in nature) or who otherwise intend to index collisions before the tiles are loaded.

Benchmarking:
Using the standard approach, indexing collisions for a 500x450 tileset took 2.53 seconds. By pre-building the collision index array in its entirety, I was able to reduce that to 259ms. However, this implementation reduces it to 8 ms. Larger tilemaps would be exponentially affected.

Considerations:
There are better implementations of if we deprecate some existing code. The advantage would be that the standard approach itself would be optimized, (thus avoiding the sort of 2500+ms costs that most code would passively implement). The parameters I added could also be defaulted to true to force this optimization to reduce its obscurity, but would require existing code to be updated.


photonstorm:
Can you think of any examples where you _wouldn't_ want to do this?

photonstorm:
Finally merging this, having fixed a small bug in it!

Issue #3837, opened 2018-07-15
Add ability to play Animations in reverse


qzix13:
Trying to play an animation in reverse. The forward property does not help.

This variant is not working. `Play `function set `forward `to true
```
spr.anims.forward = false;
spr.anims.play();
```

I hoped this would work. But no luck.
```
spr.anims.play();
spr.anims.forward = false;
```

photonstorm:
It's an internal property and should have been flagged as private, as it doesn't do what you assumed it would do. I'll rename this issue and turn it into a feature request.

photonstorm:
Thank you for submitting this feature request. We have implemented this and the feature has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

Ben-Millions:
I just pulled 3.12.0 Beta 3 Release with the intention of testing out the playReverse. And it appears to disregard the setting for repeat when yoyo: true, and it appears to repeats infinitely. I suspect it's line 645 of Animation.js, but I know I don't fully understand the class yet, and I'm not sure when I'll have time to dig around and rebuild and test it.

khaleb85:
@Ben-Millions can you send us a sample of the bug?

photonstorm:
@Ben-Millions I just tried your codepen against 3.12 Beta 4 and I can confirm that setting `yoyo` to `true` makes it repeat forever, regardless of the `repeat` value. So yes, definitely a bug in the playReverse changes.

photonstorm:
Re-opening until `yoyo` is fixed.

Glerek:
Should I resubmit my pull request? Back then I tested the full matrix: yoyo, repeat and forward/backward so maybe it would be easier for you to merge it?

khaleb85:
@photonstorm I'll try to solve it this weekend, I don't know what can cause this problem...

PR #3853, opened 2018-07-21
Add ability to play animation reverse(#3837) and some comment to TextureManager.js

Edwin222:

* Updates the Documentation
* Adds a new feature
* ~~Fixes a bug~~

Describe the changes below:

* Add a feature to play animations in reversed sequence (issue #3837)
I added a attribute `reverse` in `AnimationConfig`. default value is false. 
if reverse is true, that animation's sequence is reversed. it seems to works in my test case. Feedback please.

* Add some descriptions to TextureManager.js


khaleb85:
I`m trying to solve this in a different way, you basically are reversing the frame orders, I am trying to create a **revert** function that can reverse the animation at any time, also a **playReverse** function that runs the animation in reverse mode.

I don´t know which solution is the best one  @photostorm can tell us?

Edwin222:
I also thought about that. However, too many functions and classes are associated and all of that should be modifed to implement that way. It can cause many errors so I was trying to find a solution that doesn't have to touch original source code.

If you can implement that function safely, your solution will be better than mine.

photonstorm:
Thanks for the PR @Edwin222 ! I think I'd rather do proper reversed animation support than just flipping the array. However, the comments you added into the Texture Manager are great so I'd like to merge those! Would you mind removing the reverse change, then I'll gladly pull in the rest. Every little bit helps when it comes to the docs!

Edwin222:
Thank you for feedback. I change it.

photonstorm:
Thank you :) Just to say that reverse animation support will be coming in 3.12!

PR #3859, opened 2018-07-23
Pass `data` value to global plugins

samme:
This PR

* Adds a new feature

When starting a global plugin when the game is created, use the `data` property of the entry in the plugins array.

When starting a global plugin manually, use the new argument to `plugins.install()`.

The plugin receives the data in its `init` method.


photonstorm:
Nice. Let's add it.

PR #3860, opened 2018-07-23
Added an optional reverse parameter to the play function. (#3837)

Glerek:
This PR:
* Updates the Documentation
* Adds a new feature
* Fixes a bug

I've added an optional "reverse" parameter to the play function. The default value is false. It works with both repeat mode and yoyo mode. I also tweaked the restart function to set the initial reverse animation value and the original first animation frame (the optional startFrame parameter).

### DIFF
--- a/src/animations/Animation.js
+++ b/src/animations/Animation.js
@@ -626,7 +626,7 @@ var Animation = new Class({
 
                 this.getNextTick(component);
             }
-            else if (component.repeatCounter > 0 )
+            else if (component.repeatCounter > 0)
             {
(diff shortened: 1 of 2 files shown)


photonstorm:
Thank you for this PR, it was a nice clean and elegant solution too! However, we had already merged a PR from @khaleb85 which added a `playReverse` method on the Animation component, and a `reverse` method to reverse an already playing animation. It's quite unusual to get multiple PRs at the same time for the same feature!

Again, thank you for spending time doing this, but alas I need to close it off as the features are already in master.

PR #3863, opened 2018-07-25
Empty params reset to current texture size

tarsupin:
This PR adds a new feature: Update a physics body (arcade) to match the existing frame of a texture using empty values on setSize();

This applies to instances where you are modifying frames during gameplay and need the physics body to match the new frame.


tarsupin:
Whoops, sorry, I was going to update per your request but I was swamped for the last few days. Thanks for handling it.

PR #3864, opened 2018-07-25
Allow XHR status 0 as success too

### LINKED ISSUE #3464: Loader can't load from file:// [$50]
This Issue is about a bug in the API (v3.3.0):

I have tried wrapping a Phaser app in Cordova/ React Native.
It mostly worked but there are issues with loading files. Neither `images` nor `json` can be loaded, they fail without a proper error.

After close debugging I noticed that the "Request Type" comes up as `XHR` whereas with Phaser 2.x it's image/ json like it should be.

I have set up a very basic example `textures/sprite sheet.js` and a more complex project, both fail.

Could this be related to #2670  or  #2485? Did this fix get lost in 3.x?

Any help is very appreciated, did anybody manage to wrap a Phaser 3 project with Cordova?

Cheers!

p.s.: [www.zip](https://github.com/photonstorm/phaser/files/1854589/www.zip) the content of www within the Cordova app

screen shot 2018-03-28 at 7 22 35

UPDATE 29.03:
I have looked into this more. It seems like the issue has to do with the Phaser 3 loader not being able loading resources from `file://`. When I open the basic texture examples from Phaser 2 without being served from a server they work just fine, the Phaser 3 ones don't. Seems the loader does something differently so CORS becomes an issue.

<bountysource-plugin>

---
There is a **[$50 open bounty](https://www.bountysource.com/issues/56443560-loader-can-t-load-from-file?utm_campaign=plugin&utm_content=tracker%2F283654&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283654&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Ithamar:

* Adds a new feature
  * Allow opening games locally (without web server) when browser allows it
* Fixes a bug
  * This fixes #3464

Describe the changes below:

Normally only status 200 would be accepted as success, but 0 is returned when a file is loaded from the local filesystem (file://). This happens for example when opening the index.html of a (Phaser) game in a browser directly, or, as it turns out, when using Cordova on iOS.

### DIFF
--- a/src/loader/File.js
+++ b/src/loader/File.js
@@ -315,7 +315,10 @@ var File = new Class({
      */
     onLoad: function (xhr, event)
     {
-        var success = !(event.target && (event.target.status !== 200 && event.target.status !== 0));
+        var localFileOk = xhr.responseURL.indexOf('file://') == 0 &&
+            event.target.status === 0;
+
+        var success = !(event.target && event.target.status !== 200) || localFileOk;
 
         //  Handle HTTP status codes of 4xx and 5xx as errors, even if xhr.onerror was not called.
         if (xhr.readyState === 4 && xhr.status >= 400 && xhr.status <= 599)
--- a/src/loader/File.js
+++ b/src/loader/File.js
@@ -315,7 +315,7 @@ var File = new Class({
      */
     onLoad: function (xhr, event)
     {
-        var success = !(event.target && event.target.status !== 200);
+        var success = !(event.target && (event.target.status !== 200 && event.target.status !== 0));
 
         //  Handle HTTP status codes of 4xx and 5xx as errors, even if xhr.onerror was not called.
         if (xhr.readyState === 4 && xhr.status >= 400 && xhr.status <= 599)


photonstorm:
0 is also the code used for an XMLHttpRequest error too, which means we can't do it as simply as this I'm afraid. A full fix would check for a local file load, consider the presence of Cordova and then a response of zero.

Ithamar:
ok, will do further investigation and follow your suggestions for my next version ;)

Ithamar:
I've now added a check to only allow status 0 for file:// URLs, I considered.checking for Cordova (using window.cordova) but in its current form this would also make local development without web server work, and I think that is worth it.
Tested and saw that missing files will trigger onError() instead of onLoad() properly, so this should be safe enough I hope.

Ithamar:
@photonstorm would love to hear your opinion about my changes ;)

photonstorm:
I haven't had a chance to test this for myself, but I'm going to merge it anyway and hope some of those using Cordova can confirm one way or another (as it shouldn't harm them at all)

karkb:
How I can include this changes in my phaser.min.js file?

photonstorm:
You have to pull down the master branch and then build it (using webpack, use `npm run help` to see a list of build options)

karkb:
Thanks! Tested on iOS Cordova, it works!
Appreciate that.

karkb:
I just build phaser.min.js that's and replace it with my old phaser file.
Nothing more~

photonstorm:
@karkb no I meant what did you have to do in your Corodva setup? The html / meta tags / config.xml, CSP, that kind of stuff.

PR #3866, opened 2018-07-25
fix: Corrects the wrong JSdoc comments

tfelix:
Wrong comments lead to a faulty Typescript Definition file missing
the depth attribute.

* Updates the Documentation

Describe the changes below:

Depth property was missing in generated type script definition file. Correcting these paths fixes this error.


photonstorm:
👍

PR #3877, opened 2018-07-30
generateTexture fix

TadejZupancic:
* Fixes a bug

Describe the changes below:

generateTexture sets viewport of Graphics TargetCamera, which eventually calls updateSystem(). If scene is not set, this.config and this.sceneManager are undefined, which leads to a crash inside updateSystem().


photonstorm:
Thanks for your PR. This fix is already in master, though :) Sorry for not spotting the PR sooner.

PR #3907, opened 2018-08-09
Emit previous value with DataManager's `changedata` event

iamchristopher:

* Adds a new feature

Describe the changes below:

This updates the payload for `DataManager`'s `changedata` event to include the previous value for convenience. Previously it was not possible to access the previous value once set without caching in userland.


photonstorm:
Sure, why not. Merged.

Issue #3841, opened 2018-07-17
timeline loop ignores delay


Antriel:
```
this.tweens.timeline({
	targets: image,
	loop: 10,
	tweens: [{
		alpha: 0,
	},
	{
		alpha: 1,
		delay: 1000
	}]
});
```
The first time this plays, the delay of the second tween is applied. The remaining loops it is ignored and alpha tweens from 0 back to 1 immediately.

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

PR #3923, opened 2018-08-13
Fix a bug that tween delay doesn't applied after one loop of timeline(issue #3841)

Edwin222:

* ~~Updates the Documentation~~
* ~~Adds a new feature~~
* Fixes a bug

Describe the changes below:

Fix a bug described in issue #3841 .

After one loop of timeline, function callstack seems like to:

Timeline.nextState() => Timeline.resetTweens() => Timeline.resetTweens() => Tween.play() => Tween.resetTweenData().

And that function doesn't init the tweenData's delay if resetFromLoop is true. So I fixed it and it seems to be worked well.


photonstorm:
👍

PR #3926, opened 2018-08-13
Replace DistanceBetween with DistanceSquared in closest() and furthest() functions (ArcadePhysics.js)

Mursaat:
This PR contains just a minor performance change

Describe the changes below:
In `closest` and `furthest` functions (`ArcadePhysics.js`), distances are calculated with `DistanceBetween` using a square root, but the same result can be obtained without this operation, using `DistanceSquared`.


photonstorm:
👍

PR #3936, opened 2018-08-16
WIP: PathFollower Game Object Component

hexus:
This is an experimental implementation of a Game Object Component that implements the same behaviour as the `PathFollower` Game Object.

It's intended to allow any Game Object to become a Path Follower, as opposed to forcing the usage of Containers like #3638. This also doesn't currently introduce any breaking API changes.

I wanted this to behave similarly to `MatterGameObject`, but it was tricky keeping plain properties in the Component definition and using the same approach to mixins. I thought that perhaps these properties should be simply moved to the factory, but documenting properties there, separate from the component definition, might be a bit strange.

So, I improvised the mixin code a little. This is what I think still needs some work.

I tested the changes against the `Paths > Followers > Basic Follower` example and it worked well.

Some feedback on cleaning this up would be great.

### DIFF
--- a/src/gameobjects/components/Path.js
+++ b/src/gameobjects/components/Path.js
@@ -1,0 +1,418 @@
+/**
+ * @author       Richard Davey <rich@photonstorm.com>
+ * @copyright    2018 Photon Storm Ltd.
+ * @license      {@link https://github.com/photonstorm/phaser/blob/master/license.txt|MIT License}
+ */
+
+var DegToRad = require('../../math/DegToRad');
+var GetBoolean = require('../../tweens/builders/GetBoolean');
+var GetValue = require('../../utils/object/GetValue');
+var TWEEN_CONST = require('../../tweens/tween/const');
+var Vector2 = require('../../math/Vector2');
+
+/**
+ * Settings for a PathFollower.
+ *
+ * @typedef {object} PathConfig
+ *
+ * @property {number} duration - The duration of the path follow.
+ * @property {number} from - The start position of the path follow, between 0 and 1.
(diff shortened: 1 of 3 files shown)


rexrainbow:
[Here](https://github.com/rexrainbow/phaser3-rex-notes/blob/master/plugins/behaviors/pathfollower/Pathfollower.js) is my simple path follower behavior, which only provides property `t` for controlling the position of game object on a path. This path follower behavior does not have tween task.

[Here](https://codepen.io/rexrainbow/pen/GXKPKB?editors=0010) is a demo of this path follower : 

1 Add path follower behavior  
```javascript
pathFollower = pathFollowerPlugin.add(gameObject, {path: path});
```

2. Tween `pathFollower.t`  
```javascript
scene.tweens.add({
    targets: pathFollower,
    t: 1,
    ease: 'Linear', // 'Cubic', 'Elastic', 'Bounce', 'Back'
    duration: 1000,
    repeat: -1,
    yoyo: true
});
```

This simple path follower could work with any kind of game object which has properties `x`, `y`.

I will add more control logic for setting the angle of game object, or position offset, like official PathFollower game object does.

hexus:
I like this approach. It keeps the path following behaviour and the game object separate.

photonstorm:
Thanks for the PR. I've made my own PathFollower Component, which any custom GO can use if it wants.

hexus:
Awesome! :tada:

GKLL:
> Thanks for the PR. I've made my own PathFollower Component, which any custom GO can use if it wants.

How to add physical attributes to pathFollow to detect collisions?Thanks you

PR #3937, opened 2018-08-16
Use correct frame name in pixel perfect hit test

goldfire:
`gameObject.frame.key` doesn't exist and was passing an undefined value to `getPixelAlpha`. This just changes it to the correct `gameObject.frame.name` value.


photonstorm:
👍

PR #3941, opened 2018-08-17
Resizing of the SVG in the preload

ysraelJMM:
One of the main qualities of SVG images, is its scalability, not being able to use this, is to reduce it to be a bit more image. An SVG file has an XML format, so it can be modified with the DOM.
In order to collaborate, I have put the option to modify the parameters of the SVG and in this way make it adaptable.
The modification gives the option of being able to choose the size in the preload to which you want to have the image before rasterizing it; And so avoid the subsequent scaling that will not have the same quality.
You can see the example here: 
https://www.youtube.com/watch?v=Kd7pNjY-FLk
tested in version 3.11 It does not interfere with the original code.

* Updates the Documentation
* Adds a new feature
* Fixes a bug

Describe the changes below:

### DIFF
--- a/src/loader/filetypes/SVGFile.js
+++ b/src/loader/filetypes/SVGFile.js
@@ -92,8 +92,8 @@ var SVGFile = new Class({
         {
             var xml = null;
             var parser = new DOMParser();
-            xml = parser.parseFromString( text , 'text/xml');
-            var svgXML = xml.getElementsByTagName('svg')[ 0 ];
+            xml = parser.parseFromString(text , 'text/xml');
+            var svgXML = xml.getElementsByTagName('svg')[0];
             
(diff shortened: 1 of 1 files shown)


photonstorm:
I'm happy to include this new feature, but the implementation needs tweaking first. It would make a lot more sense if the width/height were provided as a config object, just like when loading a Sprite Sheet, and then either used or not used, based on what the dev provides.

If you can update the PR to get the values from there instead (you can copy the way Sprite Sheet File does it) then I'll be happy to merge this in. Also, don't forget to update the `FileTypesManager.register` too, so it's included in there (if you need help with updating the docs in English that's fine, just let me know)

ysraelJMM:
object svgConfig {width: number, height:number}

example: 
screenshot
https://ibb.co/n5cMse

ysraelJMM:
:) good

Issue #3946, opened 2018-08-18
TypeError: undefined is not an object for CanvasInterpolation and CanvasPool


padme-amidala:
* Phaser Version: 3.11
* Operating system: Mac OS X High Sierra 10.13.6
* Browser: Safari 11.1.2

## Description

`Phaser.Display.Canvas.CanvasInterpolation` and `Phaser.Display.Canvas.CanvasPool` are always `undefined`. I 'm trying to create a new `HTMLCanvasElement` with `Phaser.Display.Canvas.CanvasPool.create(...)` and get following error: `TypeError: undefined is not an object`.

## Example Test Code

Try to create a new HTMLCanvasElement element e. g. in TypeScript `let htmlCanvasElement = Phaser.Display.Canvas.CanvasPool.create(this, 48, 48, Phaser.CANVAS, true);`.

## Additional Information

The example code works, if we correct the names of the exports in `src/display/canvas/index.js`.

photonstorm:
Thanks for opening this issue, and for submitting a PR to fix it. We have merged your PR into the `master` branch and attributed the work to you in the [Change Log](https://github.com/photonstorm/phaser/blob/master/CHANGELOG.md). If you need to tweak the code for whatever reason please submit a new PR.

PR #3947, opened 2018-08-18
Fixed an issue where module exports are not named correct for CanvasInterpolation and CanvasPool

padme-amidala:
* Fixes a bug #3946

Fixed an issue where `TypeError: undefined is not an object` was thrown in TypeScript as soon as you tried to access a function of CanvasInterpolation or CanvasPool.


photonstorm:
👍

PR #3961, opened 2018-08-21
RenderTexture: Texture is always destroyed

TadejZupancic:

* Fixes a bug

Describe the changes below:

Texture is now destroyed only if it was not saved to the TextureManager, so it can be reused.


photonstorm:
Please fix the lint errors and I'll merge this.

TadejZupancic:
Ok, done.
I have just one short question. Is there any way to get/create the RenderTexture from Texture stored in the TextureManager? I would like to edit the same texture later, but I don't have the RenderTexture instance anymore. If it doesn't exist, I'll look into making it, because it would come handy.

PR #3962, opened 2018-08-21
Update RenderTexture fill

TadejZupancic:

* Adds a new feature
* Fixes a bug

Describe the changes below:

Add parameters to enable filling a rectangle instead of a whole texture. Besides that alpha was ignored for canvas.

I don't know much about gl, so please check if that is ok or if there's any better way.
Please check default values for width and height in @param line, so they don't mess with your documentation generation.

### DIFF
--- a/src/gameobjects/rendertexture/RenderTexture.js
+++ b/src/gameobjects/rendertexture/RenderTexture.js
@@ -355,19 +355,28 @@ var RenderTexture = new Class({
     },
 
     /**
-     * Fills the Render Texture with the given color.
+     * Fills the whole/part of the Render Texture with the given color.
      *
      * @method Phaser.GameObjects.RenderTexture#fill
      * @since 3.2.0
      *
      * @param {number} rgb - The color to fill the Render Texture with.
      * @param {number} [alpha=1] - The alpha value used by the fill.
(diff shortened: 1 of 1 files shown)


src/gameobjects/rendertexture/RenderTexture.js
```diff
@@ -378,17 +386,23 @@ var RenderTexture = new Class({
             this.renderer.setFramebuffer(this.framebuffer);
 
             var gl = this.gl;
-    
+            
+            gl.enable(gl.SCISSOR_TEST);
```
photonstorm: Don't do this unless the w/h !== the canvas w/h, it just adds un-needed gl ops.

TadejZupancic: ok, I'll add the check for w/h. I've checked a few examples and SCISSOR_TEST was always enabled at the start of the canvas draws. I also didn't find any disable command in the phaser.js. Is it necessary to enable it? I mean is it possible that it is not enabled?

photonstorm: Yes it needs enabling here.

TadejZupancic:
I've changed it, so I hope it's ok now. My guess is that disabling scissor test is not needed, because it is never disabled in the whole phaser.js source.

Issue #3779, opened 2018-06-21
Bug: isDown state won't update when mouse released outside of game canvas


rexrainbow:
[Here](https://codepen.io/rexrainbow/pen/wXmXJq?editors=0010) is a sample code of this issue (v3.10.1)
1. Press left-key of mouse inside  game canvas, 
1. Drag outside game canvas,
1. Release left-key outside game canvas,
1. Move cursor back to game canvas.

No mouse key pressed after last step, but the `scene.input.activePointer.isDown` is still `true`.

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

PR #3964, opened 2018-08-21
Reset Pointer.buttons to 0 on mouseup

SonnyCampbell:

* Fixes a bug

Describe the changes below:
The leftButtonDown(), rightButtonDown(), etc functions currently don't return false after releasing the button on the mouse. There's a check in Pointer.up() for:

`if (event.buttons) { this.buttons = event.buttons; }`

However event.buttons is always 0 on mouseup, so this.buttons never gets reset to 0.

### DIFF
--- a/src/input/Pointer.js
+++ b/src/input/Pointer.js
@@ -383,11 +383,7 @@ var Pointer = new Class({
      */
     up: function (event, time)
     {
-        if (event.buttons)
-        {
-            this.buttons = event.buttons;
-        }
-
+        this.buttons = event.buttons;
         this.event = event;
 
         //  Sets the local x/y properties


xRahman:
Hello,

there is still one problem even after applying your fix.

If you press a button (inside canvas), move the mouse out of canvas, release that button and move the mouse back into the canvas, button is reporeted in pointer events as 'down' even though it's not down anymore.

rexrainbow:
@xRahman Related issue: #3779

photonstorm:
Thanks for this PR. I've implemented a fix into the Pointer class locally which resolves this issue, so although I'm not going to merge the PR, the fix is in and attributed to you in the change log.

PR #3968, opened 2018-08-22
Fix Arcade.World#disable

samme:
This PR

* Fixes a bug

It was passing the wrong argument to disableBody, so failing.


photonstorm:
👍

PR #3979, opened 2018-08-25
Fix graphics moveFxTo and lineFxTo not work

pixelpicosean:
This PR

* Fixes a bug

Describe the changes below:

Fixes the issue that `Graphics.moveFxTo` and `Graphics.lineFxTo` not working on `master` branch, by adding missing commands operation to renderer.


photonstorm:
Thanks for this. Pretty sure it needs further tweaks somewhere else as it ignores the line colors at the moment (and renders little gaps between the triangle list) - may be easier to just remove the functions entirely to be honest.

PR #3982, opened 2018-08-26
Add CursorKeys type to Phaser.Input.Keyboard namespace

spontoreau:

* Updates the Documentation

Describe the changes below:
The CursorKeys type wasn't inside a namespace. This PR just add the Phaser.Input.Keyboard namespace to the CursorKeys type. 

Because it's only a type in the code documentation, it doesn't have any impacts on the Phaser source code but it fixes the TypeScript definition generated during tsgen execution.


photonstorm:
👍

PR #3989, opened 2018-08-30
Tiled editor changed syntax of json export

martin-pabst:
Tiled editor changed syntax of json export, see 
http://www.html5gamedevs.com/topic/38729-unexpected-tiles-properties-syntax-in-the-latest-tiled/?tab=comments#comment-220919
this fix enables phaser 3 to read the new file format.

This PR fixes a bug

Description of changes:
Inside json export files generated by Tiled (https://www.mapeditor.org/) the data structure of element tilesets[i].tiles has changed from map to array. This fix recreates the map structure so that phaser can find the tiledata for a given tile index.

### DIFF
--- a/src/tilemaps/Tileset.js
+++ b/src/tilemaps/Tileset.js
@@ -214,9 +214,11 @@ var Tileset = new Class({
     {
         if (!this.containsTileIndex(tileIndex)) { return null; }
 
-        if(!this.tileIndexMap){
+        if (!this.tileIndexMap)
+        {
             this.tileIndexMap = {};
-            for(var i = 0; i < this.tileData.length; i++){
+            for (var i = 0; i < this.tileData.length; i++)
+            {
                 this.tileIndexMap[this.tileData[i]['id']] = this.tileData[i];
             }
         }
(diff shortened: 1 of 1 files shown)


photonstorm:
I'll only merge this if you can confirm it will still work with Tiled 1.1.6 (the current official release), because this has only changed in the Beta version of Tiled (2018.08.22). There was a previous official release that swapped to using an array but it was dropped and they reverted back to the map again, which is what you'll find in 1.1.6.

If it doesn't work with 1.1.6 then please modify the PR so it'll handle both formats.

photonstorm:
Thanks for this PR. I've merged it manually and made some modifications (as it didn't declare a property) but it's now in the master branch.

martin-pabst:
Thank you!

photonstorm:
Sadly this now breaks lots of the tilemap examples :(

Going to have to look at another way to handle it.

photonstorm:
Phew.. took a bit of work, but updated the Parser and it's all working now with 1.1 and 1.2 JSON structures! :)

Issue #3928, opened 2018-08-14
Game Resize / Image Resize issue in 3.12 Beta


Mursaat:
Hello,

On that example, i'm using Phaser 3.11.0 : https://jsfiddle.net/s3La9crn/
When you resize the window at the bottom right, you can see that the sprite fits the maximum available space in the window. 

Now check the same example, with Phaser 3.12.0 : https://jsfiddle.net/d08e4v7q/
The behavior is different and strange.

Maybe I'm not doing this the right way, but it sounds more like an issue.

(I have the same problem with a tilemap instead of a sprite)

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

PR #3993, opened 2018-08-31
Modify resize() to change canvas style attribute too

Edwin222:

* Fixes a bug

Describe the changes below:

I modified resize() function WebGLRenderer and CanvasRenderer to solve issue #3928 .

Reason of that issue is original resize() function only change canvas size, not style's one.

This resize() change not only canvas size, but also canvas.style size.


photonstorm:
Thanks for this PR. The change seems to just negate the `autoResize` check entirely, as you get the same end result regardless of what you set that config property to. So it makes more sense to just force that to be `true` (which I've done in the latest commit) - I'll leave this in place until the Scale Manager takes over responsibility for this.

Edwin222:
Thank you for replying. I am looking forward to the next release.

PR #4001, opened 2018-09-01
Fixed animation playReverse with yoyo flag

khaleb85:
Fixed a bug found by @Ben-Millions (#3837) that after "playReverse" implementation, the yoyo flag run the animation infinitely.

Fixed playReverse repeat, now playReverse respect the repeat number.

I tested it with @Ben-Millions [sample ](https://codepen.io/ben-millions/pen/XPpmZW) and my sample in [phaser3 examples](https://labs.phaser.io/view.html?src=src\animation\reverse%20animation.js&v=dev).

### DIFF
--- a/src/animations/Animation.js
+++ b/src/animations/Animation.js
@@ -583,14 +583,7 @@ var Animation = new Class({
             //  Yoyo? (happens before repeat)
             if (component._yoyo)
             {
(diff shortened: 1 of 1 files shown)


photonstorm:
Thanks for the quick turn-around on this.

PR #4012, opened 2018-09-04
Allow matter js bodies to render their debug graphics individually

jf908:

This PR

* Fixes a bug

Describe the changes below:

This small change allows the debug graphics to be turned off and on individually for matter js bodies. Before this commit, if a body has their `render.visible` value set to false, it stops all subsequent debug graphics from rendering.


photonstorm:
👍

PR #4014, opened 2018-09-04
fixed incorrect properties assignment in setTileScale method

alexeymolchan:

* Fixes a bug

Describe the changes below:

setTileScale method of the TileSprite game object now correctly set tileScale instead of tilePosition


photonstorm:
👍

PR #4019, opened 2018-09-10
Add support for setting float array uniforms in the WebGLPipeline / WebGLRenderer

Mattykins:
This PR
* Adds a new feature

Describe the changes below:

Added support for setting float array uniforms. There are 4 new methods on both the WebGLRenderer and WebGLPipeline

* setFloat1v
* setFloat2v
* setFloat3v
* setFloat4v

P.S.
My editor decided to remove some trailing spaces in the files I edited. According to the contribution guide we don't like trailing spaces anyways, so I left the changes in :) If you would like me to revert those lines back to how they were for any reason please let me know and I can take care of that.

### DIFF
--- a/src/renderer/webgl/WebGLPipeline.js
+++ b/src/renderer/webgl/WebGLPipeline.js
@@ -546,7 +546,7 @@ var WebGLPipeline = new Class({
     /**
      * Set a uniform value of the current pipeline program.
      *
-     * @method Phaser.Renderer.WebGL.WebGLPipeline#setFloat1
+     * @method Phaser.Renderer.WebGL.WebGLPipeline#setFloat1v
      * @since 3.2.0
      *
      * @param {string} name - [description]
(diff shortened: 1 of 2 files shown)


photonstorm:
I'm happy to merge this in, although the constant creation of new Float32Arrays in the `setFloat` methods is a bit jarring, although there's no easy way around it (beyond perhaps passing them in, but they still have to be created at some point).

The trailing space editor config is for code, not comments. Trailing spaces on comment lines are fine. Removing them won't hurt though.

Mattykins:
We could do something like

```js
    setFloat1v: function (program, name, arr)
    {
        this.setProgram(program);
        this.gl.uniform1fv(this.gl.getUniformLocation(program, name), (arr instanceof Float32Array) ? arr : new Float32Array(arr));
        return this;
    },
```

Thoughts?

photonstorm:
Nah that'd probably even worse. Would be better to make you pass in a Float32Array instead of casting it each time, then it's up to you if you pay the cost or not. Or at least when.

Mattykins:
Hmmm. Think it’s worth changing to accept a Flot32Array rather than instantiate one? I feel like this method is more niche anyways so if the API is more complex in this area it’s not a huge detriment to end users.

photonstorm:
Probably worth changing, yes. As you're likely to want to change the uniform values constantly.

Mattykins:
Good call. I'll update the PR tonight.

PR #4024, opened 2018-09-11
Fix game object memory leak

goldfire:
When calling `destroy` directly on a game object, the `shutdown` event for this game object on the scene never gets turned off, thus leading to a memory leak. I wasn't 100% sure if this is where/how this should be done, but it seems to be working for us in our testing.

### DIFF
--- a/src/gameobjects/GameObject.js
+++ b/src/gameobjects/GameObject.js
@@ -560,6 +560,8 @@ var GameObject = new Class({
 
         //  Tell the Scene to re-sort the children
         sys.queueDepthSort();
+        
+        this.scene.sys.events.off('shutdown', this.destroy, this);
 
         this.active = false;
         this.visible = false;


goldfire:
So, it turns out there is an issue with this. After deploying this, we discovered that there were random lag spikes, which turned out to be when several dozen game objects were destroyed in a single frame. Is there any reason this is using events for `shutdown` instead of just looping through the `list` when shutting the scene down? Seems like that would be much more efficient (for example, our main scene has 4-5k game objects, and even though most of them are inactive, this still causes thousands of idle events to be setup).

photonstorm:
The original issue (not clearing the `shutdown` event list) has been resolved in the master branch.

The follow-up comment sounds more like a re-structuring / feature request, so best opened as such as an issue, rather than lost to time in a PR comment.

PR #4032, opened 2018-09-14
Wrong filling rectangle for camera with background

aaronfc:

* Fixes a bug

Describe the changes below:
After some changes related to scissor the camera background is displaced.
Bug can be reproduced in https://labs.phaser.io/view.html?src=src\camera\minimap%20camera.js from 3.12.0 version.


photonstorm:
👍

photonstorm:
Thanks. This fixes the background, but the example still doesn't render properly, likely because of the scissor values. Will have a look.

PR #4038, opened 2018-09-17
Fixes issue with tweens not restarting when triggered in a row.

rgk:
This PR

* Fixes a bug

Describe the changes below:

https://labs.phaser.io/edit.html?src=src/tweens/restart.js

In this example just add an additional tween.restart(); under the existing tween.restart(); to see the issue.

This fix allows multiple restarts to work in a row.

### DIFF
--- a/src/tweens/tween/Tween.js
+++ b/src/tweens/tween/Tween.js
@@ -417,6 +417,7 @@ var Tween = new Class({
         else
         {
             this.stop();
+            this.parent.makeActive(this);
             this.play();
         }
     },


photonstorm:
I don't mind merging this, but why on earth would you ever restart a tween multiple times in a row? Sounds like a fix for a copy paste error in userland.

rgk:
I am using websockets with my phaser game, which calls events to be ran in the game. If the game is minimized during the events, they stack and get called multiple times. I started noticing only events with tweens I restart fail. I do agree this might be a bit of a design flaw on my end.

photonstorm:
Thanks for your PR. I have fixed this via another method, but you can now call `restart` multiple times without it messing up.

Issue #4000, opened 2018-09-01
Feature suggestion : arcade physics engine `body.setMaxSpeed(speed)` 


rexrainbow:
Currently arcade physics engine has `body.setMaxVelocity(x, y)` to limit velocity  in x and y part separated.
For example, `body.setMaxVelocity(200)`
- If acceleration is (200, 0) every tick, the velocity will be (200, 0) -> max speed is `200`. 
- If acceleration  is (200, 200) every tick, the velocity will be (200, 200) -> max speed is `282.8427`.... 

How about adding `body.setMaxSpeed(speed)` to limit the `speed` (vector length of velocity) in a steady value?

photonstorm:
Yeah, worth adding at some point I think.

rexrainbow:
Set MaxVelocity will also limit the final angle of moving direction when speed-up continuously. 
For example  -- [asteroids movement](https://labs.phaser.io/view.html?src=src\physics\arcade\asteroids%20movement.js), moving direction of ship will become 45/135/225/315 degrees (maxVelocity is (200,200) here) if initial angle of ship is not 0.

Edwin222:
Is someone working on this feature? If not, I'd like to try to do this.

PR #4044, opened 2018-09-20
Add setMaxSpeed() function.

Edwin222:

* Adds a new feature

Describe the changes below:

I added setMaxSpeed() function requested on #4000 issue.

This function limits vector length of arcade physics body's velocity. I implemented it by checking vector length of new velocity value when Physics.Arcade.World  computes velocity.

if it has maxSpeed value and new vector length is greater than max speed value, it decrease new velocity's X and Y value to fit max speed. formula used for this is based on trigonometric function so change of vector's direction from change of value is applied as normally.

### DIFF
--- a/src/physics/arcade/Body.js
+++ b/src/physics/arcade/Body.js
@@ -421,6 +421,17 @@ var Body = new Class({
         this.maxVelocity = new Vector2(10000, 10000);
 
         /**
+         * The Body's maximum speed.
+         * 
+         * This limits Body's scalar value of speed. Always positive value.
+         * 
+         * Negative value means that there is no limitation on speed.
+         * 
+         *  
+         */
+        this.maxSpeed = -1;
+
+        /**
          * If this Body is `immovable` and in motion, this the proportion of this Body's movement received by the riding body on each axis, relative to 1.
          * The default value (1, 0) moves the riding body horizontally in equal proportion and vertically not at all.
          *
@@ -1515,6 +1526,22 @@ var Body = new Class({
     setMaxVelocity: function (x, y)
     {
         this.maxVelocity.set(x, y);
+
+        return this;
+    },
+
+    /**
+     * Sets the Body's maximum speed(Scalar value).
+     * 
+     * @method Phaser.Physics.Arcade.Body#setMaxSpeed
+     * @since 3.13.0
+     * 
+     * @param {number} s - The max speed value, in pixels per second. 
+     * 
+     * @return {Phaser.Physics.Arcade.Body} This Body Object.
+     */
+    setMaxSpeed: function (s){
+        this.maxSpeed = s;
 
         return this;
     },
--- a/src/physics/arcade/Body.js
+++ b/src/physics/arcade/Body.js
@@ -427,7 +427,9 @@ var Body = new Class({
          * 
          * Negative value means that there is no limitation on speed.
          * 
-         *  
+         * @name Phaser.Physics.Arcade.Body#maxSpeed
+         * @type {number}
+         * @since 3.13.0 
          */
         this.maxSpeed = -1;
 
--- a/src/physics/arcade/Body.js
+++ b/src/physics/arcade/Body.js
@@ -423,7 +423,7 @@ var Body = new Class({
         /**
          * The Body's maximum speed.
          * 
-         * This limits Body's scalar value of speed. Always positive value.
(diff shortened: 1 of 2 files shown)


rexrainbow:
The initial value of maxVelocity is a very larger number `10000`
```javascript
this.maxVelocity = new Vector2(10000, 10000);
```
Maybe initial value of `maxSpeed` could be set to another larger number like `20000` to prevent `(maxSpeed >= 0)` comparison, and have similar style with `maxVelocity`.

src/physics/arcade/World.js
```diff
@@ -1354,7 +1356,16 @@ var World = new Class({
         velocityX = Clamp(velocityX, -maxX, maxX);
         velocityY = Clamp(velocityY, -maxY, maxY);
 
-        body.velocity.set(velocityX, velocityY);
+        newSpeed = Math.sqrt(velocityX * velocityX + velocityY * velocityY);
```
samme: This calculation should be skipped if maxSpeed < 0.

Edwin222: Thank you for opinion. I will apply it.

samme:
Just in case you're interested, this can also be done with the vector methods:

```javascript
if (body.velocity.length() > maxSpeed)
{
    body.velocity.normalize().scale(maxSpeed);
}
```

Edwin222:
It will be better and more nice to look. Thanks for comments.

src/physics/arcade/Body.js
```diff
@@ -438,6 +438,21 @@ var Body = new Class({
...
+         * 
+         * @name Phaser.Physics.Arcade.Body#maxSpeed
+         * @type {number}
+         * @since 3.13.0 
+         */
+        this.maxSpeed = -1;
+
+        /**
+         * If this Body is `immovable` and in motion, this the proportion of this Body's movement received by the riding body on each axis, relative to 1.
+         * The default value (1, 0) moves the riding body horizontally in equal proportion and vertically not at all.
```
njt1982: This comment part looks wrong to me? Is it intended to replace the first two lines? Or did the base comment change and this PR is accidentally reverting/prepending it back?

src/physics/arcade/World.js
```diff
@@ -1355,6 +1357,9 @@ var World = new Class({
         velocityY = Clamp(velocityY, -maxY, maxY);
 
         body.velocity.set(velocityX, velocityY);
+        if(maxSpeed >= 0 && body.velocity.length() > maxSpeed){
```
njt1982: It looks like existing code stype should:
a) Put a space after the opening if and before the `(`
b) Put the opening curly brace on the next line.

(Look at some of the `if`'s above)

njt1982: Ironically, I personally prefer curly braces on the same line, like this (although I like spaces before them).

njt1982:
I couldn't apply the patch of this PR to the latest Phaser 3 release, so had to manually apply it but I am getting this:

`TS2339: Property 'setMaxSpeed' does not exist on type 'Image'.`

```
export class MainScene extends Phaser.Scene {
  player: Phaser.Physics.Arcade.Image
...
... 
  create(): void {
    this.player = this.physics.add.image(0, 0, 'player')
    this.player.setMaxSpeed(200)
  }
}
```

I'm wondering if we also need to add the method to the Velocity component too? (to map it from the component to the physics body)?

src/physics/arcade/World.js
```diff
@@ -1270,6 +1270,8 @@ var World = new Class({
         var maxY = body.maxVelocity.y;
 
         var speed = body.speed;
+        var maxSpeed = body.maxSpeed;
+        var newSpeed = 0;
```
njt1982: Is this an obsolete variable?

src/physics/arcade/Body.js
```diff
@@ -1548,6 +1563,22 @@ var Body = new Class({
...
+     * Sets the maximum scalar value of Body's velocity.
+     * 
+     * @method Phaser.Physics.Arcade.Body#setMaxSpeed
+     * @since 3.13.0
+     * 
+     * @param {number} s - The max speed value, in pixels per second. 
+     * 
+     * @return {Phaser.Physics.Arcade.Body} This Body Object.
+     */
+    setMaxSpeed: function (s){
```
njt1982: `{` should be on the next line

PR #4065, opened 2018-09-25
Matter.js Force Component Docs

rgk:

* Updates the Documentation

Describe the changes below:

Doc updates, checking if I did this well before I start jumping into multiple files.

### DIFF
--- a/src/physics/matter-js/components/Gravity.js
+++ b/src/physics/matter-js/components/Gravity.js
@@ -18,7 +18,7 @@ var Gravity = {
      * @method Phaser.Physics.Matter.Components.Gravity#setIgnoreGravity
      * @since 3.0.0
      *
-     * @param {boolean} value - Flag toignore gravity.
+     * @param {boolean} value - Flag to ignore gravity.
      *
      * @return {Phaser.GameObjects.GameObject} This Game Object.
      */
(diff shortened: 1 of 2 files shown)


src/physics/matter-js/components/Gravity.js
```diff
@@ -5,20 +5,20 @@
...
+ * A component to manipulate gravity for Matter.js bodies.
  *
  * @name Phaser.Physics.Matter.Components.Gravity
  * @since 3.0.0
  */
 var Gravity = {
 
     /**
-     * [description]
+     * Ignore gravity.
```
photonstorm: This is a bit too short for such an important method really. It should at least explain where the gravity comes from (the world simulation) and that it causes this body to ignore that completely. It can also be toggled in real-time.

src/physics/matter-js/components/Gravity.js
```diff
@@ -5,20 +5,20 @@
...
      *
      * @method Phaser.Physics.Matter.Components.Gravity#setIgnoreGravity
      * @since 3.0.0
      *
-     * @param {boolean} value - [description]
+     * @param {boolean} value - Flag to ignore gravity.
```
photonstorm: Should be something like:

Set to `true` to ignore the effect of world gravity, or `false` to include it.

photonstorm:
Yeah generally it's fine, have added my comments.

PR #4074, opened 2018-09-28
Create RenderTexture from existing texture/frame

TadejZupancic:

* Adds a new feature

Describe the changes below:

Now a RenderTexture can be (re)created from an existing texture or a specific frame within that texture. This allows one to e.g. create and draw GameObjects to RenderTexture and later modify it.


TadejZupancic:
I use it in the following way. I create one RT, save it as texture and prerender frames needed to display history for different games to that texture - one frame is usually made from a combination of images and some graphics objects.  This frames are then used to display the history to the user. When the history for one game changes, I just "edit" the specific frame (create a RT from that frame and redraw it). In that case I use only one texture for all the frames and can edit only the specific frame. And since the texture and frame do not change, none of the images using them need to be changed.

I could create a RT and draw a new frame to it each time something needs to be changed, but in that case I would have to use one texture for each game or draw all the other history frames to it each time. And that texture would probably be a different instance each time something changed, so I would also have to change all the references in the images containing that texture.

As I see it this gives me way less work to do and is probably also faster than recreating everything from scratch each time one game's history changes.

photonstorm:
You can edit a saved RT at any point, without having to re-create them. So I think, what you're saying, is that you use an RT like an atlas, and want to 'create' a new RT using just a frame of an existing RT, and draw only to that area of it? Hence all of the scissor operations in your PR, etc.

TadejZupancic:
Yes, I'm using it like atlas and need to change only the specific frame when I receive new data. I use it in different scenes and I don't have the existing RT instance available - only the texture and I haven't found a way to get it from the texture.

photonstorm:
Based on your PR, if you created a new RT using a frame from an existing RT, it wouldn't actually edit the original RT at all, but would create a brand new texture with the frame drawn into it - yes? I.e. if you edit the new RT, the old one isn't updated.

PR #4077, opened 2018-09-30
Fix null game.context after WebGLRenderer init

samme:
This PR

* Fixes a bug

`game.context` was always null even after `game.renderer.gl` was created.

I guess WebGLRenderer#init() used to be run later, but now it's run immediately during construction.


photonstorm:
👍

PR #4087, opened 2018-10-04
Tween inside Timeline fix

TadejZupancic:

* Fixes a bug

Describe the changes below:

The stop method in Tween assumed, that tween's parent was TweenManager. If the Tween has been added to the Timeline, that was not true and the stop method crashed.


photonstorm:
👍

Issue #4088, opened 2018-10-04
First frame takes on appearance of second frame in Spritesheets created from trimmed Texture Atlas Frames


Cirras:
## Version

* Phaser Version: v3.14.0-FB

* Operating system: Windows 10

* Browser: Not applicable

## Description

In SpriteSheets created from a texture atlas with `TextureManager.addSpriteSheetFromAtlas()`, the first frame becomes identical to the second frame. In all usage, including animations, the first frame will appear as the second frame instead.

## Example Test Code

The following code is pulled directly from the following [Phaser 3 Example](http://labs.phaser.io/edit.html?src=src\textures\sprite%20sheet%20from%20atlas.js). I have modified all the animation configurations to only use the first 2 frames in order to demonstrate that the animations derived from the trimmed texture atlases will appear to "freeze" because both frames are identical due to this bug, while their untrimmed counterparts will rapidly display the first 2 frames as expected.

```javascript
var config = {
    type: Phaser.AUTO,
    parent: 'phaser-example',
    width: 800,
    height: 600,
    scene: {
        preload: preload,
        create: create
    }
};

var game = new Phaser.Game(config);

function preload ()
{
    this.load.path = 'http://labs.phaser.io/assets/atlas/trimsheet/';

    this.load.atlas('testanims', 'trimsheet.png', 'trimsheet.json');
}

function create ()
{
    var t1 = this.textures.addSpriteSheetFromAtlas(
        'boom1',
        {
            atlas: 'testanims',
            frame: 'explosion-notrim',
            frameWidth: 64,
            frameHeight: 64,
            endFrame: 23
        });


    var t2 = this.textures.addSpriteSheetFromAtlas(
        'boom2',
        {
            atlas: 'testanims',
            frame: 'explosion',
            frameWidth: 64,
            frameHeight: 64,
            endFrame: 23
        });

    var b1 = this.textures.addSpriteSheetFromAtlas(
        'bubble1',
        {
            atlas: 'testanims',
            frame: 'bubble-notrim',
            frameWidth: 34,
            frameHeight: 68
        });


    var b2 = this.textures.addSpriteSheetFromAtlas(
        'bubble2',
        {
            atlas: 'testanims',
            frame: 'bubble',
            frameWidth: 34,
            frameHeight: 68
        });

    var config1 = {
        key: 'explode1',
        frames: this.anims.generateFrameNumbers('boom1', { start: 0, end: 1 }),
        frameRate: 20,
        repeat: -1
    };

[...]

PR #4089, opened 2018-10-04
Make Texture.add() firstFrame check more explicit (Fix issue #4088)

Cirras:
Fixes issue #4088 where a SpriteSheet created from a trimmed texture atlas frame returns the second frame when the first frame is requested.

This PR
* Fixes a bug

Describe the changes below:
-Made simple change to `Texture.add()` that will prevent the second frame from being selected as the `firstFrame` in the event that __BASE texture does not exist.


photonstorm:
This is now fixed (albeit indirectly) because the sprite sheet parser now automatically creates a `__BASE` entry, where it didn't before. However, I like this checking approach better anyway, so am going to merge it anyway.

Issue #4096, opened 2018-10-06
Improper sizing on single-row/single-column SpriteSheets created from trimmed Texture Atlas frames


Cirras:
## Version

* Phaser Version: v3.14.0-FB

* Operating system: Windows 10

* Browser: Not applicable

## Description

In SpriteSheets created from a (trimmed) texture atlas frame with `TextureManager.addFromTextureAtlas()`, if the resulting spritesheet is single-row or single-column then the spritesheet frame sizes will be calculated improperly and show parts of the surrounding textures in the texture atlas when used.

## Example Test Code

In the following example, we will meet Priestly the Slime. Unfortunately, we'll also meet portions of the texture(s) below Priestly in the texture atlas due to this bug. Have a look.

```javascript
import 'phaser';

// NOTE: Below is a patch to fix issue #4088 for the purposes of this example.
// It can be commented out without issue, but the first frame of the animation will be skipped.
{
    var Texture = require('phaser/src/textures/Texture');
    var Frame = require('phaser/src/textures/Frame');

    Texture.prototype.add = function (name, sourceIndex, x, y, width, height)
    {
        var frame = new Frame(this, name, sourceIndex, x, y, width, height);

        this.frames[name] = frame;

        //  Set the first frame of the Texture (other than __BASE)
        //  This is used to ensure we don't spam the display with entire
        //  atlases of sprite sheets, but instead just the first frame of them
        //  should the dev incorrectly specify the frame index
        if (this.firstFrame === '__BASE')
        {
            console.log('firstName', this.firstFrame);
            this.firstFrame = name;
        }

        this.frameTotal++;

        return frame;
    }
}

const config = {
    type: Phaser.AUTO,
    width: 500,
    height: 500,
    pixelArt: true,
    roundPixels: true,
    scene: {
        preload: preload,
        create: create
    }
};

var game = new Phaser.Game(config);

function preload()
{
    this.load.path = 'https://game.bones-underground.org/';
    this.load.multiatlas('trimmedMultiAtlas', 'gfx/6/all.json');
}

function create()
{
    this.cameras.main.setZoom(3);

    this.textures.addSpriteSheetFromAtlas(
        'Priestly', 
        {
            atlas: 'trimmedMultiAtlas',
            frame: '1818',
            frameWidth: 45,
            frameHeight: 52
        });

    this.anims.create({
        key: 'PriestlyAnimation',
        frames: this.game.anims.generateFrameNumbers('Priestly'),
        frameRate: 2,
[...]

PR #4097, opened 2018-10-06
Improved trim handling for Spritesheets created from trimmed Texture Atlas Frames

Cirras:
This PR
* Fixes a bug

Describe the changes below:
Fixes issue #4096 by modifying the calculation method for spritesheet frame sizes in `Parsers.SpriteSheetFromAtlas()`.
Previously the calculation method did not handle the case where `topRow && bottomRow` or `leftRow && rightRow` for a single frame, causing problems when a spritesheet was single-row or single-column. One more corner case covered, eh?

Also fixed an apparent typo on line 155 where the `rightRow` bool was being added to `frameX` instead of the correct `rightWidth` value. (Trivial)


photonstorm:
👍

PR #4101, opened 2018-10-09
Text style setStyle/setFont fix

DotTheGreat:

* Updates the Documentation
* Fixes a bug

Describe the changes below:

When using a font string instead of setting ``fontFamily``, ``fontSize``, and ``fontStyle`` in either setStyle or setFont, the fields  ``fontFamily``, ``fontSize``, and ``fontStyle`` wouldn't get set. This isn't a problem while creating the text object because TextStyle.update won't get called, which sets TextStyle._font to said fields. (``this._font = [ this.fontStyle, this.fontSize, this.fontFamily ].join(' ').trim();``)


photonstorm:
Nice one 👍

PR #4111, opened 2018-10-13
Fix Phaser.GameObjects.Shape#setStrokeStyle JSDoc

darkwebdev:

* Updates the Documentation
* Fixes a bug

Describe the changes below:

This should fix TypeScript definitions, that at the moment lead to an unexpected behavior (cause color is passed instead of a line width).

At the moment definition looks like this:
```setStrokeStyle(color?: number, alpha?: number): this;```

and hopefully will look like this:
```setStrokeStyle(lineWidth?: number, color?: number, alpha?: number): this;```

At the moment inside IDE:
image


photonstorm:
👍

PR #4113, opened 2018-10-15
Simple fix for a reference error when culling tilemaps

Siyalatas:

* Fixes a bug

Describe the changes below:
Panning the camera to an area that did not have tile info would throw an Uncaught Reference error, as mapData[y] does not exist.


photonstorm:
Nice 👍

PR #4118, opened 2018-10-16
Fix issue with null config in Arcade PhysicsGroup constructor

pierpo:
# This PR

* Fixes a bug

# Describe the changes below:

This code was crashing:

```javascript
    const monsterGroup = this.physics.add.group(this.monsters);
```

with the following stacktrace:

image

It turns out the case were the config is null was not properly handled.

The execution reached this instruction without going through any of the if cases! So `config` was simply `undefined`, which leads to a crash.

image

### DIFF
--- a/src/physics/arcade/PhysicsGroup.js
+++ b/src/physics/arcade/PhysicsGroup.js
@@ -121,6 +121,14 @@ var PhysicsGroup = new Class({
                 singleConfig.removeCallback = this.removeCallbackHandler;
             });
         }
+        else
+        {
+            // config is not defined and children is not a plain object nor an array of plain objects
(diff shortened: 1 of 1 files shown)


samme:
I just tried `this.physics.add.group(null)` and `this.physics.add.group(null, null)` and neither gave an error.

pierpo:
@samme Yes, because this case is already handled. Both `children` and `config` are null so you get through the first `if` condition.

The problem is when you have `children` that is an array of non plain objects (not representing a config object) and no `config`.

photonstorm:
As unusual as this is, it's a valid edge-case I think.

PR #4119, opened 2018-10-17
Pick up animation data from Tiled 1.2+

nkholski:

* Fixes a bug

Describe the changes below:

Animation data was lost in the parser of Tiled 1.2+. I'm unaware what the objectgroup property is and if it can coexist with the animation property so I played safe. If it can't coexist the test in the if could be removed:

// Copy animation data
if(tile.animation) {
tiles[tile.id] = { animation: tile.animation };    
}


nkholski:
BTW it's tested. The bug made my Animated Tiles Plugin to break, but after the above fix makes it work as expected again without any changes in the plugin itself.

photonstorm:
No idea if it can co-exist or not, so let's leave the check in!

PR #4121, opened 2018-10-18
Event emits "loaderror" opposed to camel cased "loadError" as said in…

jestarray:


photonstorm:
Thanks for the PR, but it's the docs that should change, not the event name. All events are lower-case through-out the API.

jestarray:
@photonstorm ahh I see. Alright, fixed the docs to reflect emitted event names.

PR #4136, opened 2018-10-29
Array of dead particles is now being filled up with dead particles.

Waclaw-I:
* Fixes a bug

Describe the changes below:

Hi!
Currently, when using particle emitters, dead particles pool is not functioning properly, here is the example:
[click](https://codepen.io/anon/pen/PygvGW?editors=0010)

It is caused by not catching value that `this.dead.concat(rip);` is returning.

Best regards.


photonstorm:
👍

photonstorm:
Ok. I've recoded how dead particles are handled and it's working properly now :)

PR #4144, opened 2018-11-04
fixed - cameraFilter bitmask doesn't work for values < 0

stuartkeith:

* Fixes a bug

Describe the changes below:

The `cameraFilter` bitmask is only checked if its value is more than 0. A full bitmask is equal to -1, so I've changed the check to specifically check for !== 0.

I discovered this when coming up with a way of easily only showing something on one camera - `foo.cameraFilter = ~camera.id`, which for a camera id of 1 returns -2.


photonstorm:
Thanks. I've also made the change to the Game Object class, as it's more useful there!

PR #4148, opened 2018-11-06
Fix Passing in Data to the Scene via Add

kainage:

* Fixes a bug

Describe the changes below:

Data was not being passed in to the scene manager when adding a scene via `add`.


photonstorm:
👍

PR #4152, opened 2018-11-08
Fix Tile.tileset to return just the containing tileset instead of all…

### LINKED ISSUE #4095: Tile.tileset incorrectly holds an array
## Version
Phaser Version: 3.14.0

## Description

Tile.tileset now incorrectly holds an array of tilesets instead of a tileset object. This breaks Tile.getCollisionGroup and Tile.getTileData.

quadrupleslap:
Fixes #4095, but I'm not sure if it's the right way to do it.

### DIFF
--- a/src/tilemaps/Tile.js
+++ b/src/tilemaps/Tile.js
@@ -761,8 +761,9 @@ var Tile = new Class({
     },
 
     /**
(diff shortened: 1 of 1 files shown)


photonstorm:
I've fixed some linting errors, but otherwise, this seems fine to me.

PR #4156, opened 2018-11-10
Fix TS definition of Set#iterate, callbackScope is an optional param

RoryO:

* Updates the Documentation

Describe the changes below:

Fixes TS definition of Set#iterate to `callbackScope` as optional.


photonstorm:
👍

PR #4159, opened 2018-11-11
Added RotateTo Effect to Camera

jan1za:
* Adds a new feature

I have added a new RotateTo Effect on the 2D Camera. The effect works like the others Pan, Zoom etc. It receives the target rotation value in radians.
If the value is positive the rotation is performed in a clockwise direction else the rotation is done anti-clockwise.
There is added parameter called shortestPath which if it is set to true the effect calculates if a clockwise rotation is faster than an anti-clockwise and performs which ever is fastest.

Thank you again for a very cool framework.


xBradley:
ty

photonstorm:
A little late, perhaps, but may as well get this merged! 👍

PR #4165, opened 2018-11-15
fix tileLayer.setCollisionFromCollisionGroup() to work with multiple tilesets

mgrubinger:

* Fixes a bug

Describe the changes below:

Fixes [#4095](https://github.com/photonstorm/phaser/issues/4095) by looping over all tilesets of a tile to get the CollisionGroups required for SetCollisionFromCollisionGroup().

Tile.getCollisionGroup() returns an array for each tileset.
SetCollisionFromCollisionGroup() iterates through this array.

Please let me know if there are implications of doing this which I don't understand since I'm pretty new to the Phaser code.

### DIFF
--- a/src/tilemaps/Tile.js
+++ b/src/tilemaps/Tile.js
@@ -335,7 +335,14 @@ var Tile = new Class({
      */
     getCollisionGroup: function ()
     {
-        return this.tileset ? this.tileset.getTileCollisionGroup(this.index) : null;
+        if (!this.tileset) { return null; }
+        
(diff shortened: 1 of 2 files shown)


photonstorm:
Thanks for your PR! However, I'm pretty sure we don't need this now, as Tile.tileset is now properly returning the correct tileset the tile is from, meaning these methods should return to their previous working state. Please test the master branch to confirm if you can.

mgrubinger:
Great to hear it's fixed! Thanks!

Issue #4168, opened 2018-11-15
StaticBody#setCircle offset not working and also not rendering as circle in debug


maikthomas:
## Version

* Phaser Version:3.11.0

* Operating system:MacOS 10.13.6

* Browser: Not specific, have tested on Chrome and Firefox

## Description

I have identified two problems:

1. A StaticBody whose body has been set to a circle with `setCircle` does not render as a circle in debug. However the collision body is in fact a circle.
2. The offset parameters appear not to actually offset the body. In this case I observe that the render and collision body are both not offsetting.

## Example Test Code

I made a codepen with a few bodies.
There are two staticbodies on the left. 
The top one is square, the bottom is circle. They both render as square in debug.
I added moving bodies to collide with them. You can see the bottom moving body collides with the circle and bounces off at an angle, confirming the circular collision.

Also please try changing the offset for the static bodies, it has no effect.

https://codepen.io/maikthomas/pen/OaMRGb

## Additional Information
Related forum post I made, has screenshots and code but not complete sample for running: http://www.html5gamedevs.com/topic/41088-setcircle-doesnt-work-on-a-staticgroup-game-object/
It at least shows how for non static bodies the debug render works as expected.

P.s. many thanks for all your effort, I'm new to Phaser and it is great! 👏

maikthomas:
Think these are linked https://github.com/photonstorm/phaser/issues/4213
As far as I can tell no static body offset methods are working correctly.

photonstorm:
Hiya, thank you for opening this (all those months ago!) - static bodies now debug draw properly thanks to your PR and `setOffset` definitely now works for Static bodies, too, as tested in 3.20.

Rayanfer32:
I tried using `setOffset` on this [repl](https://repl.it/@s3/korisukka) ...still doesnt work!!!

samme:
@Rayanfer32 

![Screen Shot 2020-04-03 at 11 29 35 AM](https://user-images.githubusercontent.com/435681/78393390-854e3880-759e-11ea-851e-3051e2b0fd9b.png)

PR #4170, opened 2018-11-16
[issue/4168] draw circular StaticBody as circle in drawDebug

maikthomas:

* Fixes a bug

Describe the changes below:
As described in #4168, circular staticbodies were not drawn in drawDebug as a circle.
I have test this in the example `public/src/physics/arcade/static group.js` by adding the following:
```
    group.children.entries.forEach(ball => {
        ball.body.setCircle(16);
    });
```
We can clearly see the fix working for the balls in the second screenshot below.
Before fix:
image
After fix: 
image

### DIFF
--- a/src/physics/arcade/StaticBody.js
+++ b/src/physics/arcade/StaticBody.js
@@ -788,10 +788,21 @@ var StaticBody = new Class({
     {
         var pos = this.position;
 
+        var x = pos.x + this.halfWidth;
+        var y = pos.y + this.halfHeight;
+
         if (this.debugShowBody)
         {
             graphic.lineStyle(1, this.debugBodyColor, 1);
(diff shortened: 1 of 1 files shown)


maikthomas:
I see the travis built failed for this PR, but they are not my changes and I didn't want to fix them as I assumed that would cause conflict as someone else will probably fix them soon.
https://travis-ci.com/maikthomas/phaser/builds/91689887?utm_medium=notification&utm_source=email

photonstorm:
Nice one 👍

maikthomas:
@photonstorm thanks!
Unfortunately this doesn't solve this other problem reported in #4168 about the offset not working,
I spent like 3 hours messing around with the offset in StaticBody setCircle and couldn't get any effect whatsoever.

PR #4173, opened 2018-11-18
Fix: Cannot read property 'index' of undefined at GetTileAt and RemoveTileAt

WaSa42:
Fixes a bug: `Cannot read property 'index' of undefined at GetTileAt`

A tile was undefined and the method failed because there were no case handling undefined.
I assumed that the behavior was the same as `tile === null`.


photonstorm:
👍

PR #4179, opened 2018-11-21
Allow set framerate when forceSetTimeOut is true

pavels:
* Fixes a bug

when `forceSetTimeOut` is true, allow setting target frame rate by setting `fps.target` value in config.

### DIFF
--- a/src/dom/RequestAnimationFrame.js
+++ b/src/dom/RequestAnimationFrame.js
@@ -82,6 +82,16 @@ var RequestAnimationFrame = new Class({
          */
         this.lastTime = 0;
 
+        /**
+         * The target rate in [ms].
+         *
+         * @name Phaser.DOM.RequestAnimationFrame#target
+         * @type {number}
+         * @default 0
+         * @since 3.16.0
+         */
+        this.target = 0;
+
         var _this = this;
 
         /**
@@ -116,7 +126,7 @@ var RequestAnimationFrame = new Class({
         {
             var d = Date.now();
 
-            var delay = Math.max(16 + _this.lastTime - d, 0);
+            var delay = Math.min(Math.max(_this.target * 2 + _this.tick - d, 0),_this.target);
 
             _this.lastTime = _this.tick;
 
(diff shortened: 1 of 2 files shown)


src/dom/RequestAnimationFrame.js
```diff
@@ -116,7 +126,7 @@ var RequestAnimationFrame = new Class({
         {
             var d = Date.now();
 
-            var delay = Math.max(16 + _this.lastTime - d, 0);
+            var delay = Math.min(Math.max(_this.target * 2 + _this.tick - d, 0),_this.target);
```
photonstorm: Unless `target` is given a default value > 0, there will never be a delay set, because this will always return zero.
Also, space after commas in argument lists.

src/dom/RequestAnimationFrame.js
```diff
@@ -82,6 +82,16 @@ var RequestAnimationFrame = new Class({
          */
         this.lastTime = 0;
 
+        /**
+         * The target rate in [ms].
+         *
+         * @name Phaser.DOM.RequestAnimationFrame#target
```
photonstorm: The property name feels ambiguous. It's not clear from the name (or jsdoc) that it only impacts setTimeout.

src/dom/RequestAnimationFrame.js
pavels: I have fixed the comma

the `target` is set from `start` method always - i can set so sane default for example 16 (that will be 60 fps) but that will anyway be overwritten in `start` method - what do you think?

src/dom/RequestAnimationFrame.js
pavels: added more docs

photonstorm:

https://phaser.discourse.group/t/phaser-3-performance/1322/7

mdenti:
@photonstorm, @pavels, we'd love to have this working for our project, what's stopping it?

photonstorm:

https://phaser.discourse.group/t/how-to-limit-fps-with-phaser-3/275/11

Thundros:
When is this coming?!

Thundros:
?

Thundros:
@photonstorm When is this coming?!

photonstorm:
Nothing is stopping anyone pulling this file into their own fork if they need it urgently.

photonstorm:
This feature is now part of the master branch and will be in 3.21. I couldn't merge this PR, but all of your changes are in there anyway.

PR #4180, opened 2018-11-21
Fix 'collide' events triggered by overlap only

samme:
This PR

* Fixes a bug

If you had an Arcade Body with onCollide=true and onOverlap=false it could receive a 'collide' event in response to an overlap.


photonstorm:
Thanks. I've merged this fix locally and attributed it to you.

PR #4181, opened 2018-11-21
Correct docs for Phaser.Boot.Config#audio

samme:
This PR

* Updates the Documentation

Boot.Config#audio was the wrong type (I think).

### DIFF
--- a/src/boot/Config.js
+++ b/src/boot/Config.js
@@ -32,15 +32,13 @@ var ValueToColor = require('../display/color/ValueToColor');
 /**
  * Config object containing various sound settings.
  *
- * @typedef {object} SoundConfig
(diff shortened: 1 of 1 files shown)


pavle-goloskokovic:
This change is not correct. 
`SoundConfig` is used internally and for passing sound objects their configuration. 
`AudioConfig` is used for passing sound manager config to the game and should be defined as a separate type.

samme:
There's still a definition for SoundConfig in <https://github.com/photonstorm/phaser/blob/v3.15.1/src/sound/index.js#L15>.

I think the one I removed was a duplicate, so it should be safe.

photonstorm:
This PR is correct, cheers.

PR #4183, opened 2018-11-22
setFont fix (also fixes #4141)

### LINKED ISSUE #4141: toJSON() on Text doesn't save the font style when using the "font" shorthand
I'm using a Google font in my project, loaded like this:

```
<!DOCTYPE html>
<html>
	<head>
		<style>
			@font-face {
				font-family: 'GameFont';
				src: url( 'assets/LuckiestGuy-Regular.ttf' )
			}
			
			.fontPreload {
				font-family: 'GameFont';
				position: absolute;
				left:-100px;
			}
		</style>
	</head>
	<body>
		<div class="fontPreload">.</div>
	</body>
</html>
```

In game, I'm displaying a Text field like so:

```
const t = scene.add.text( 100, 100, `This is some game text. and some more here and here`, {
	font: "20px GameFont",
	color: "#ffffff"
})
```

It displays correctly in game.

If I print out the result of `toJSON()`, it's missing the font name and size:

```
{
  ...
  "data": {
    "autoRound": true,
    "text": "This is some game text. and some more here and here",
    "style": {
      "fontFamily": "Courier",
      "fontSize": "16px",
      "fontStyle": "",
      "backgroundColor": null,
      "color": "#ffffff",
      "testString": "|MÃ‰qgy",
      ...
    },
    ...
}
```

If, when creating the Text object, I don't use the `font` shorthand, the JSON result is correct:

```
const t = scene.add.text( 100, 100, `This is some game text. and some more here and here`, {
	fontFamily: "GameFont",
        fontSize: 20,
	color: "#ffffff"
})
```

```
{
  ...
  "data": {
    "autoRound": true,
    "text": "This is some game text. and some more here and here",
    "style": {
      "fontFamily": "GameFont",
      "fontSize": "20px",
      ...
      }
    },
    ...
  }
}
```

DottieDot:

* Fixes a bug

Describe the changes below:

When entering a font string in `setFont`, `fontFamily` would hold all the data (which isn't an issue when drawing the text because the font properties get merged into 1 string anyways), this is something I noticed when I checked if my previous PR (#4101) fixed #4141 (which it did if it weren't for that mistake).


photonstorm:
Thanks for the update!

PR #4195, opened 2018-11-27
Fix for from/to in PathFollower.js

kensleebos:
* Fixes a bug

Describe the changes below:
Fixed path follower from/to value. It was overwritten by default.


photonstorm:
As the comment above those lines implies, these were overridden on purpose for a reason, although it was so long ago it escapes me what it was. I'll merge this PR for now and make a note to undo it if it breaks anything.

Issue #4197, opened 2018-11-28
RND(RandomDataGenerator) is null.


UYEONG:
## Version
* Phaser Version: v3.16.0 Beta 3-FB
* Operating system: MacOS
* Browser: Chrome

## Description

Version 3.15.x has rnd in const.js.
```
// math/const.js
// ...
    /**
     * An instance of the Random Number Generator.
     * 
     * @name Phaser.Math.RND
     * @type {Phaser.Math.RandomDataGenerator}
     * @since 3.0.0
     */
    RND: new RND()
};
```

But Version 3.16.0 does not have rnd in const.js.

```
// math/const.js
// ...
    /**
     * An instance of the Random Number Generator.
     * 
     * @name Phaser.Math.RND
     * @type {Phaser.Math.RandomDataGenerator}
     * @since 3.0.0
     */
    RND: null
};
```

So this problem occurs in the demo "[Matter Detect Collision With Tile](https://labs.phaser.io/view.html?src=src\game%20objects\tilemap\collision\matter%20detect%20collision%20with%20tile.js)".

```
matter detect collision with tile.js:66 Uncaught TypeError: Cannot read property 'integerInRange' of null
    at Scene.<anonymous> (matter detect collision with tile.js:66)
    at InputPlugin.emit (index.js:182)
    at InputPlugin.processDownEvents (InputPlugin.js:804)
    at InputPlugin.update (InputPlugin.js:548)
    at EventEmitter.emit (index.js:203)
    at Systems.step (Systems.js:364)
    at SceneManager.update (SceneManager.js:558)
    at Game.step (Game.js:552)
    at TimeStep.step (TimeStep.js:560)
    at step (RequestAnimationFrame.js:104)
```

that the code is missing in the refactored.

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

bigswede74:
@photonstorm I'm using phaser 3 typescript and from looking at the [phaser 3 typescript github page](https://github.com/digitsensitive/phaser3-typescript) the package.json is using v3.18.1. How can I get the updated 3.19.0 if I'm using typescript?

```
  "dependencies": {
    "phaser": "3.18.1"
  },
```

photonstorm:
[...]

PR #4198, opened 2018-11-28
Add RND in math/const.js #4197

UYEONG:
See https://github.com/photonstorm/phaser/issues/4197


photonstorm:
Thanks for your PR, but this is now done in the Game Config.

PR #4219, opened 2018-12-08
removed legacy onUpdate animation description

jestarray:
the feature was removed since 3.14 or so


photonstorm:
The section about "modify the animation timing" is still true, so worth keeping in.

PR #4223, opened 2018-12-10
Add facebook plugin to Phaser.Scene for JSDoc

bryanwood:
This PR adds the facebook plugin to `Phaser.Scene` for JSDoc.


photonstorm:
👍

Issue #4235, opened 2018-12-15
Object disappears when Tween duration is zero


BigZaphod:
## Version

* Phaser Version: 3.15.1

* Operating system: macOS 10.14.2

* Browser: Safari 12.0.2

## Description

When a tween's duration is 0, the object being animated disappears.

## Example Test Code

Using https://labs.phaser.io/edit.html

```
var config = {
    type: Phaser.AUTO,
    width: 800,
    height: 600,
    parent: 'phaser-example',
    scene: {
        preload: preload,
        create: create,
    }
};
var game = new Phaser.Game(config);

function preload () {
    this.load.image('block', 'assets/sprites/block.png');
}

function create () {
    var image = this.add.image(200, 300, 'block');
    this.tweens.add({
        targets: image,
        x: 300,
        duration: 0
    });
}
```

Change the duration to anything greater than 0, and it works - but when duration is 0, the sprite just disappears instead. I'd expect it to just jump to the end of the tween instantly if the duration is 0.

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

PR #4240, opened 2018-12-16
Fix for issue #4235.

AndrewDRX:
This PR:
* Fixes a bug

Description:
Fix for issue #4235 . The issue occurs due to the use of a NaN result produced from a zero denominator equation.


photonstorm:
Thanks for the PR but I've resolved this issue via capping the duration during the `init` of the Tween.

Issue #4271, opened 2019-01-02
Leaderboard getScores doesn't take count and offset in parameter.


Oramy:
## Version

* Phaser Version: 3.10.1

* Operating system: Windows

* Browser: Firefox

## Description
This bug is simple. The count and offset parameters of the getScores method of the Leaderboard class are fixed to their default value. Inspecting the code shows that they are not passed as parameters to the Facebook API.
I tried to show the first place of my leaderboard but the next results showed up.

## Example Test Code
I cannot really provide an example code as to be reproduced it needs a Facebook Leaderboard.

## Additional Information
The solution is to add arguments to getEntriesAsync in the Leaderboard method. 
I will try to make a pull request.

Thanks for making this engine.

photonstorm:
Thanks for opening this issue, and for submitting a PR to fix it. We have merged your PR into the `master` branch and attributed the work to you in the [Change Log](https://github.com/photonstorm/phaser/blob/master/CHANGELOG.md). If you need to tweak the code for whatever reason please submit a new PR.

PR #4272, opened 2019-01-02
Fixed Issue #4271

Oramy:
* Fixes a bug

Fixes the Leaderboard.getScores method which didn't take its arguments into account. See issue #4271


photonstorm:
👍

Issue #4256, opened 2018-12-24
Two way collision between circles, with arcade physics


JeSuisUnCaillou:
## Version
* Phaser Version: 3.15.1
* Operating system: Windows 10
* Browser: Not specific.

## Description

This concerns the **arcade** physics engine.

If you have two Circles colliding "sideways" while both having a velocity, the bounce will be extremely weird, like if they are "bouncing back".

## Example Test Code

This code is almost exactly the same than this example : http://labs.phaser.io/view.html?src=src\physics\arcade\circular%20body.js&v=3.15.1

I just changed the initial velocities to make them do the weird collide bug.

```
var config = {
    type: Phaser.AUTO,
    width: 800,
    height: 600,
    parent: 'phaser-example',
    physics: {
        default: 'arcade',
        arcade: {
            debug: true,
            gravity: { y: 100 }
        }
    },
    scene: {
        preload: preload,
        create: create
    }
};

var game = new Phaser.Game(config);

function preload ()
{
    this.load.image('wizball', 'assets/sprites/wizball.png');
}

function create ()
{
    var ball1 = this.physics.add.image(100, 240, 'wizball');
    var ball2 = this.physics.add.image(700, 240, 'wizball');

    ball1.setCircle(46);
    ball2.setCircle(46);

    ball1.setCollideWorldBounds(true);
    ball2.setCollideWorldBounds(true);

    ball1.setBounce(1);
    ball2.setBounce(1);

    // here are the changed velocities 
    ball1.setVelocity(150, 0);
    ball2.setVelocity(-150, 0);

    this.physics.add.collider(ball1, ball2);
}

```

## Additional Information

I'm currently building a game of "bumper ships", where the collisions should work like a billard game. This weird bounce is the only thing that is holding me back.

Billard games are very common, and I wonder if anyone has tried to make one with the arcade physics engine. If you did, I'll be glad to hear how you managed the collisions.

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

Issue #579, opened 2014-03-15
WebGL render in Internet Explorer 11


cub:
With Internet Explorer 11 I get a background transparent and an overlight effect : 
Exemple with the Invaders game exemple
![ie11](https://f.cloud.github.com/assets/1578812/2429617/249eb012-ac8e-11e3-86f5-39a99e70f452.jpg)

By the way this exemple don't work correctly : the invaders/shoot don't collide correctly ;)

In the changelog I see : 
We now force IE11 into Canvas mode to avoid a Pixi bug with pre-multiplied alpha. Will remove once that is fixed, sorry, but it's better than no game at all, right? :(

The test don't work for me.
game.device.trident is false.

When I test manually the regexp /Trident\/(\d+.\d+); rv:(\d+.\d+)/ @Device.js:582 on navigator.userAgent I get no match.
navigator.userAgent @IE11 : 
"Mozilla/5.0 (Windows NT 6.1; WOW64; Trident/7.0; SLCC2; .NET CLR 2.0.50727; .NET CLR 3.5.30729; .NET CLR 3.0.30729; .NET4.0C; .NET4.0E; Media Center PC 6.0; .NET CLR 1.1.4322; rv:11.0) like Gecko"
There is no " rv:" directly after the Trident version but at the end.

I'm not very good with regexp to propose a path..

But.. I tested http://www.goodboydigital.com/pixijs/examples/15/indexAll.html on IE11 and everything works. Pixi.js support now fully the WebGL @ IE11 ?

PR #4279, opened 2019-01-05
Fix circles collision #4256

akuskis:
This PR:

* Fixes a bug

Describe the changes below:

According to the issue #4256 , math of Circle-Circle Collision was updated.

Before:
original_1

After:
fixed_1

Before:
original_2

After:
fixed_2

### DIFF
--- a/src/physics/arcade/World.js
+++ b/src/physics/arcade/World.js
@@ -1504,11 +1504,6 @@ var World = new Class({
         GetOverlapX(body1, body2, false, bias);
         GetOverlapY(body1, body2, false, bias);
 
-        var dx = body2.center.x - body1.center.x;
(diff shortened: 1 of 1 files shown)


samme:
@akuskis You're welcome to attack https://github.com/photonstorm/phaser-ce/issues/579 as well. 😃

akuskis:
@samme
With suggested changes:

* I can see bug explained in #579. When balls are grouped and 'physics.collide' is called on that group from 'update' function.
* I do not see any issues. When balls are created  and collision is set between all of them.

Need to take a look, why behaviour is different, but I think that it's a separate issue.

photonstorm:
Lovely PR, thanks 👍

PR #4286, opened 2019-01-09
Fix WebGLRenderer to apply transparent bg in config

gomachan7:
This PR 
* Fixes a bug

When configured as `Phaser.Boot.Config.transparent = true`, canvas's bg is still colored on WebGLRenderer.  But on running by CanvasRenderer, it is applied correctly.

So I fixed WebGLRenderer to clear color as transparent on `transparent` is `true`.


photonstorm:
👍

Issue #4284, opened 2019-01-08
Arcade Physics step


photonstorm:
During collision or overlap the colliding bodies are updated multiple times per step.

samme:
If it's the same as #4154 I can make a PR.

photonstorm:
Almost certainly is. You can see it manifest clearly in the scenes/sub game demo (if you rotate while colliding with the edge of a planet) and in the physics/mass collision test where lots of collisions are missed (and in 3.11 it works fine)

Issue #4154, opened 2018-11-09
Arcade Body deltas can be inconsistent during collisions


samme:
## Version
* Phaser Version: 3.15.1

## Description

When a body collides with several bodies during a single step, it has correct `deltaX()` and `deltaY()` values only during the first collision. On any subsequent collisions during the step, `deltaX()` and `deltaY()` are 0.

In the example, this makes friction work correctly only for the first colliding body.

## Example Test Code

<https://codepen.io/samme/pen/oQxaVb?editors=0010>

See the platform's deltaX values in console.

## Additional Information

I think this is caused by calling the body's postUpdate method during separation (14d3f6b, #3777).

samme:
Here's another example. Box collides with 5 balls simultaneously but only 2 balls are moved:

<https://codepen.io/samme/pen/vQGvgQ>

photonstorm:
Agreed, it's the reason the mass collide demo doesn't work properly anymore either. Will look into it.

PR #4287, opened 2019-01-09
Queue 'late' colliding bodies for a second update

samme:
This PR

* Fixes a bug (#4154, #4284)

Describe the changes below:

Arcade.World#separate no longer calls postUpdate on separated bodies. Instead, if Arcade.World#update has already happened, it records these bodies in a new set and then calls postUpdate on each one exactly once, during Arcade.World#postUpdate.

There are two new private properties, Arcade.World#late and Arcade.World#_late. I wish I had better names for them.

### DIFF
--- a/src/physics/arcade/World.js
+++ b/src/physics/arcade/World.js
@@ -222,6 +222,28 @@ var World = new Class({
         this.pendingDestroy = new Set();
 
         /**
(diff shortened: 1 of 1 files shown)


photonstorm:
Thanks, works nicely and fixes the mass collision test, along with others.

PR #4288, opened 2019-01-10
animation{event}-key was not firing on the Animation event emitter

jestarray:
fixes this issue: 
```javascript
var config = {
    type: Phaser.AUTO,
    parent: 'phaser-example',
    width: 800,
    height: 600,
    scene: {
        preload: preload,
        create: create
    }
};

function preload ()
{
    this.load.spritesheet('boom', 'assets/sprites/explosion.png', { frameWidth: 64, frameHeight: 64, endFrame: 23 });
}

function create ()
{
    var config = {
        key: 'explode',
        frames: this.anims.generateFrameNumbers('boom', { frames: [ 0, 1, 2, 1, 2, 3, 4, 0, 1, 2 ] }),
        frameRate: 20,
        repeat: -1
    };

    this.anims.create(config);

    //DOES NOT WORK
    let animation = this.anims.get("boom").on('animationupdate-explode', ()=> {
        console.log("play sound or something")
    }); 

    let booomie = this.add.sprite(400, 300, 'boom').play('explode');

    //WORKS FINE
    boomie.on("boom").on('animationupdate-explode', ()=> {
        console.log("play sound or something222")
    }); 
}

```
https://labs.phaser.io/edit.html?src=src/animation/generate%20frames.js (reproducible if you can get it running on latest dev release)

### DIFF
--- a/src/animations/Animation.js
+++ b/src/animations/Animation.js
@@ -808,6 +808,8 @@ var Animation = new Class({
 
                 this.emit('repeat', this, frame);
 
+                component.currentAnim.emit('animationrepeat-' + this.key, this, frame, component.repeatCounter, parent);
+
                 parent.emit('animationrepeat-' + this.key, this, frame, component.repeatCounter, parent);
 
                 parent.emit('animationrepeat', this, frame, component.repeatCounter, parent);
--- a/src/gameobjects/components/Animation.js
+++ b/src/gameobjects/components/Animation.js
@@ -676,6 +676,8 @@ var Animation = new Class({
 
         anim.emit('start', anim, frame);
 
+        anim.emit('animationstart-' + key, anim, frame, gameObject);
+
         gameObject.emit('animationstart-' + key, anim, frame, gameObject);
 
         gameObject.emit('animationstart', anim, frame, gameObject);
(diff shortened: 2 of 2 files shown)


photonstorm:
I don't really understand the need for this change. The key-based events were never meant to be emitted from the Animation objects, because an Animation can only ever have one key, so it's pointless emitting both `animationupdate` and `animationupdate-key`, as they come from the same base object.

The keyed events are meant for Game Objects only.

jestarray:
Must have misunderstood the changelog:
> The Animation class now extends the Event Emitter and dispatches events itself. This allows you to listen for events from a specific Animation, rather than via a Game Object. This is handy, for example, if you had an explosion animation that you wanted to trigger a sound effect when it started. You can now listen for the events from the Animation object directly.

Won't all gameObjects have their own copy of the listener `animationstart-explode` ? Since I always want the explode animation to play a sound afterwards for all gameObjects, I was thinking I could just apply the listener once on the Animation object rather than have every gameObject cloning it.

photonstorm:
You can listen for the events directly on the Animation, they're still emitted. It just doesn't need to emit the `-key` version of the event on the animation _as well_, because Animations only have one key, so listen for `Animation.on('animationcomplete')` is enough.

jestarray:
Ahhhh, :man_facepalming: I get it now. :+1:

PR #4301, opened 2019-01-16
Adds invertAlpha flag to GeometryMask

tfelix:
Similiar to the invertAlpha flag on Bitmap Mask this flag if set to true
will esentially invert the function of the stencil buffer: non drawn
shapes will become invisible and drawn shapes will be visible.

* Adds a new feature

Describe the changes below:

As said in the commit message this flag will invert the function of the stencil buffer when used on the GeometryMask. Its nice when you want to show only a small rectangle for example.


photonstorm:
Why not, looks useful - thanks :)

PR #4311, opened 2019-01-23
Updated tilemap jsdocs

jestarray:


src/tilemaps/mapdata/LayerData.js
```diff
@@ -29,7 +29,7 @@ var LayerData = new Class({
         if (config === undefined) { config = {}; }
 
         /**
-         * [description]
+         * The name of the layer(specified in Tiled)
```
photonstorm: Can you add a space before the open brace please. In fact, I think just `The name of the layer, if specified in Tiled.` would be fine.

src/tilemaps/mapdata/LayerData.js
```diff
@@ -56,7 +56,7 @@ var LayerData = new Class({
         this.y = GetFastValue(config, 'y', 0);
 
         /**
-         * [description]
+         * The width in tile units of the layer.
```
photonstorm: `The width in tiles of the layer` (there's no such thing as a tile unit really)

src/tilemaps/mapdata/LayerData.js
```diff
@@ -65,7 +65,7 @@ var LayerData = new Class({
         this.width = GetFastValue(config, 'width', 0);
 
         /**
-         * [description]
+         * The height in tiles units of the layer.
```
photonstorm: See width.

src/tilemaps/mapdata/MapData.js
```diff
@@ -18,7 +45,7 @@ var GetFastValue = require('../../utils/object/GetFastValue');
  * @constructor
  * @since 3.0.0
  *
- * @param {object} [config] - [description]
+ * @param {MapDataConfig} [config] - [description]
```
photonstorm: 👍

src/tilemaps/mapdata/MapData.js
```diff
@@ -145,7 +172,7 @@ var MapData = new Class({
          * An array with all the layers configured to the MapData.
          * 
          * @name Phaser.Tilemaps.MapData#layers
-         * @type {array}
+         * @type {Array.<LayerData|ObjectLayer>}
```
photonstorm: Typed arrays in jsdoc are represented like this: `LayerData[]` and when you've got a type that can be more than one thing you wrap them in brackets, so: `{(LayerData[]|ObjectLayer)}`.

PR #4322, opened 2019-01-26
fix: GetValue not checking for the existence of '.' correctly

kyranet:

* Fixes a bug

Describe the changes below:

[`String.prototype.indexOf()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/indexOf) returns `-1` when the passed string is not found, which is truthy, so the branch was always executed regardless of its presence.


photonstorm:
👍

Issue #4291, opened 2019-01-11
Align 'justified' in text gameobject


andrewbaranov:
## Version

* Phaser Version: last Phaser 3 Sandbox

* Operating system: Windows 10 64 bit

* Browser: Chrome 71.0.3578.98 (64 bit)

## Description

Justified style not work in Text GameObject.
In Phaser docs:
https://photonstorm.github.io/phaser3-docs/Phaser.GameObjects.Text.html
setAlign(align)
Set the text alignment.
Expects values like 'left', 'right', 'center' or 'justified'.

## Example Test Code

1. Open Phaser3 labs
http://labs.phaser.io/edit.html?src=src/game%20objects/text/static/multi%20line%20text%20with%20alignment.js
2. Replace text style with    align: 'justified'
3. Align becomes like align left
Expected text bocomes justified

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

andrewbaranov:
Works perfect.
Usage: set text style { fontFamily: 'Arial', color: '#00ff00', align: 'justify' }
Pay attention for align: '**justify**' NOT '**justified**'

PR #4332, opened 2019-01-29
Issue/4291 Align 'justified' in text gameobject

Donerkebap13:

* Adds a new feature

Describe the changes below:
Issue #4291 
Added the 'justified' align property to Text object

Usage:
this.add.text(100, 100, content, { fontFamily: 'Arial', color: '#00ff00', align: **'justified'** });

### DIFF
--- a/src/gameobjects/text/static/Text.js
+++ b/src/gameobjects/text/static/Text.js
@@ -1198,6 +1198,25 @@ var Text = new Class({
             {
                 linePositionX += (textSize.width - textSize.lineWidths[i]) / 2;
             }
+            else if (style.align === 'justified')
+            {
+                var minimumLengthToApplyJustification = 0.85;
+                if (textSize.lineWidths[i] / textSize.width >= minimumLengthToApplyJustification)
+                {
+                    var extraSpace = textSize.width - textSize.lineWidths[i];
+                    var spaceSize = context.measureText(' ').width;
+                    var extraSpaceCharacters = Math.floor(extraSpace / spaceSize);
+                    var array = lines[i].split(' ');
+    
+                    for (var s = 0, idx = 0; s < extraSpaceCharacters; s++)
+                    {
+                        array[idx] += ' ';
+                        idx = (idx + 1) % array.length;
+                    }
+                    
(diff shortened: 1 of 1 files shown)


src/gameobjects/text/static/Text.js
```diff
@@ -1198,6 +1198,27 @@ var Text = new Class({
             {
                 linePositionX += (textSize.width - textSize.lineWidths[i]) / 2;
             }
+            else if (style.align === 'justified')
+            {
+                var minimumLengthToApplyJustification = 0.85;
```
photonstorm: Where does the 0.85 value come from?

dude78GH: I guess this is some kind of ratio. I.e. if line width is less than 85% of defined width per line - there is no point of justifying it coz spaces would be too long.

dude78GH:
Algorithm above needs few modifications I believe. Just compare.
This text uses code above to justify
Screenshot_2019-04-29_14-47-14

This one uses code below:
Screenshot_2019-04-29_14-47-50

```javascript
else if (style.align === 'justified')
{
    var minimumLengthToApplyJustification = 0.85;    // to justify text line its width must be not less than 85% of defined width
    if (textSize.lineWidths[i] / textSize.width >= minimumLengthToApplyJustification)
    {
        var extraSpace = textSize.width - textSize.lineWidths[i];
        var spaceSize = context.measureText(' ').width;
        var trimmedLine = lines[i].trim();
        var array = trimmedLine.split(' ');

        extraSpace += (lines[i].length - trimmedLine.length) * spaceSize;

        var extraSpaceCharacters = Math.floor(extraSpace / spaceSize);
        var idx = 0;
        while (extraSpaceCharacters > 0)
        {
            array[idx] += ' ';
            idx = (idx + 1) % (array.length - 1 || 1);
            --extraSpaceCharacters;
        }

        lines[i] = array.join(' ');
    }
}
```

photonstorm:
Thanks for this PR (and the update to it @dude78GH). I have used the updated code and made the change locally, so although I won't merge this PR, the feature is still now in the master branch and you're both credited in the Change Log.

Issue #3845, opened 2018-07-19
Tile.getCenterX and Tile.getCenterY returns wrong value for scaled layers


oloflarsson:
The following two lines:
* https://github.com/photonstorm/phaser/blob/master/src/tilemaps/Tile.js#L473
* https://github.com/photonstorm/phaser/blob/master/src/tilemaps/Tile.js#L487

Currently look like this:
* `return this.getLeft(camera) + this.width / 2;`
* `return this.getTop(camera) + this.height / 2;`

But they should probably look something like this:
* `return (this.getLeft(camera) + this.getRight(camera)) / 2;` 
* `return (this.getTop(camera) + this.getBottom(camera)) / 2;` 

Or perhaps like this:
* `return (tilemapLayer) ? tilemapLayer.tileToWorldX(this.x + 0.5, camera) : (this.x + 0.5) * this.baseWidth`
(taking inspiration from Tile.getLeft: https://github.com/photonstorm/phaser/blob/master/src/tilemaps/Tile.js#L368)

What currently happens is that `tile.getCenterX()` and `tile.getCenterY` will return correct values when the layer has scale 1, but wrong values for other scales.

oloflarsson:
I propose setting these labels: `Bug`, `Priority: Low`, and `Difficulty: Easy`.

This is a code example of how I work around the bug in the game I'm making:

```javascript
hitCoin (sprite, tile) {
    this.layer.removeTileAt(tile.x, tile.y);
    window.incrementGold();

    const x = (tile.getLeft() + tile.getRight()) / 2;
    const y = (tile.getTop() + tile.getBottom()) / 2;
    const style = PxlColor.GOLD;
    const count = 30;
    const lifespan = 500;
    this.particles.explosion(x, y, style, count, lifespan);
}
```

florianvazelle:
I managed to reproduce the issue, and I'm looking for my first open source contribution, so I would like to work on it.

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

PR #4336, opened 2019-01-30
Fix getCenter methods returning wrong value for scaled layers

florianvazelle:

* Updates the Documentation
* Adds a new feature
* Fixes a bug

Describe the changes below:
Tile class methods getCenterX and getCenterY now returning correct values with scaled layers.
Fix the issue #3845.


photonstorm:
👍

PR #4346, opened 2019-02-06
Added this.scale property to Scene

strangeweekend:
This PR

* Updates the Documentation

Describe the changes below:
Purely a doc fix, noticed in the latest typedefs that this property was unavailable yet logging out this.scale within a scene showed it existed. It's existence is also referenced by Phaser.Scene.Systems#scale description.


photonstorm:
👍

PR #4347, opened 2019-02-06
On mobile, have pointerup and pointerout get triggered together.

rgk:

* Fixes a bug

Describe the changes below:

In 3.16.1 if you trigger the pointerup event and then move to another interactive object, the past object's pointerover event gets triggered. This clears that and stops that from happening by stopping the pointerover event from being triggered if the pointerup event has been.


photonstorm:
👍

PR #4350, opened 2019-02-08
Fix: Type declaration for blendMode

brian-lui:

* Fixes a bug

Describe the changes below:

BlendMode can be either a string or an enum


photonstorm:
👍

Issue #4357, opened 2019-02-11
Phaser not correctly restoring state after using 'ESC' to exit fullscreen.


khutchins:
## Version

* Phaser Version: 3.16.2

* Operating system: Windows 10

* Browser: Chrome 72.0.3626.96, Firebox 65.0 (doesn't seem browser-specific)

## Description

When exiting fullscreen mode using 'ESC', I expect that the Phaser game window will be restored to its previous size. Instead, it appears to be keeping its fullscreen size but in windowed mode.

## Example Test Code

https://codepen.io/anon/pen/XOqMVP

To reproduce:
1. Launch code pen.
2. Click on Phaser window to gain focus.
3. Press 'F' to trigger Phaser to toggle fullscreen.
4. Exit fullscreen using ESC (not the natural Phaser path)
5. Phaser window will not be the same size as it was before entering fullscreen.

photonstorm:
Ahhh nuts, yes, you're right.

codypotter:
Perhaps this could be fixed by changing the canvas to default size when the exit full screen event is captured.
`document.addEventListener('fullscreenchange', exitHandler);
document.addEventListener('webkitfullscreenchange', exitHandler);
document.addEventListener('mozfullscreenchange', exitHandler);
document.addEventListener('MSFullscreenChange', exitHandler);`

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

khutchins:
I'm no longer running into the issue with the version I built from head. Thanks.

PR #4366, opened 2019-02-14
Fix exiting full screen issue

HeyStevenXu:
Fetch current parent size before taking action in updateScale

* Fixes a bug

Describe the changes below:

This PR references issue #4357 . The updateScale function is not always rescaling to the most up-to-date value of the parent element of the canvas. As a result, the canvas is still trying to scale to a fullscreen size, even after exiting via ESC. 

Though there are different ways to fix this bug, simply making sure the parent boundary information is up-to-date seems to be the core of the issue.

Let me know what you think. Thanks!


photonstorm:
Thanks for the PR. I've solved this a very slightly different way and credited you in the change log.

PR #4373, opened 2019-02-16
fix animation yoyo when using repeatDelay

cruzdanilo:
This PR

* Fixes a bug

Describe the changes below:
fix animation yoyo when using repeatDelay

### DIFF
--- a/src/animations/Animation.js
+++ b/src/animations/Animation.js
@@ -589,8 +589,10 @@ var Animation = new Class({
 
         if (component._reverse === !isReverse && component.repeatCounter > 0)
         {
-            component.forward = isReverse;
-
+            if (!component._repeatDelay || component.pendingRepeat)
+            {
+                component.forward = isReverse;
+            }
             this.repeatAnimation(component);
 
             return;


photonstorm:
What is this actually fixing?

cruzdanilo:
when using both yoyo and repeatDelay, the delay was only respected after pairs of loops.

cruzdanilo:
you can try this code and see the error happening:

```javascript
function preload() {
  this.load.atlas('gems', 'assets/tests/columns/gems.png', 'assets/tests/columns/gems.json');
}

function create() {
  this.anims.create({
    key: 'diamond',
    frames: this.anims.generateFrameNames('gems', { prefix: 'diamond_', end: 15, zeroPad: 4 }),
    yoyo: true,
    repeatDelay: 1000,
    repeat: -1,
  });
  this.add.sprite(400, 100, 'gems').play('diamond');
}

const config = {
  type: Phaser.AUTO,
  parent: 'phaser-example',
  width: 800,
  height: 600,
  scene: { preload, create },
};

```

photonstorm:
Sorry, I can't see anything wrong in the way the code above runs. Here is a visually clearer example:

```
    var image = this.add.image(100, 300, 'block');

    var tween = this.tweens.add({
        targets: image,
        x: 700,
        duration: 2000,
        ease: 'Power2',
        yoyo: true,
        repeat: -1,
        repeatDelay: 2000
    });
```

The block moves to y 700, yoyo's back to y 100 immediately and then waits 2000ms before repeating.

If you are expecting it to wait at y 700 for 2000ms before returning again, you should use the `hold` property to do this:

```
    var image = this.add.image(100, 300, 'block');

    var tween = this.tweens.add({
        targets: image,
        x: 700,
        duration: 2000,
        ease: 'Power2',
        yoyo: true,
        repeat: -1,
        hold: 2000
    });
```

Now it will stick at y 700 for 2000ms before returning again. If you wish it to delay at both ends of the tween, you should add in both `hold` and `repeatDelay` to achieve this.

cruzdanilo:
@photonstorm the fix is related to frame based animation, not tweens. please, give my code a little attention.

cruzdanilo:
@photonstorm as i said, after the first loop the character always jumps (back and forth) two times before holding the delay:

PR #4376, opened 2019-02-17
Docs/update types of params for creator factories

G-Rath:
This PR

* Updates the Documentation

## Changes

Adds configuration objects for all `create` GameObject methods.

Renames a couple of `Options` into `<GameObject>Config`, which might break some code if anyone was using those types directly (off the top of my head I think it was just Graphics).

Couple of inconsistencies I found that could be corrected: 
 * ImageCreator: image doesn't take x & y values, but factory does.
 * MeshCreator: mesh doesn't take x & y values, but factory does.
 * TextCreator: if style is null, padding will cause an error.
 * TextCreator: don't know how to type "style".
 * TextCreator: couldn't find definition of "resolution" property.
 * `config` is not optional for all creators.

Let me know if you'd be interested in me doing some "strengthening" of these methods (as in, cleaning them by making sure they're consistent in their defaults, naming, what variables they accept vs their factory, documentation, etc).

### DIFF
--- a/src/gameobjects/bitmaptext/dynamic/DynamicBitmapTextCreator.js
+++ b/src/gameobjects/bitmaptext/dynamic/DynamicBitmapTextCreator.js
@@ -17,7 +17,7 @@ var GetAdvancedValue = require('../../../utils/object/GetAdvancedValue');
  * @method Phaser.GameObjects.GameObjectCreator#dynamicBitmapText
  * @since 3.0.0
  *²
- * @param {Phaser.GameObjects.BitmapText.Types.BitmapTextConfig} config - The configuration object this Game Object will use to create itself.
+ * @param {Phaser.GameObjects.BitmapText.Types.BitmapTextConfig} [config] - The configuration object this Game Object will use to create itself.
  * @param {boolean} [addToScene] - Add this Game Object to the Scene after creating it? If set this argument overrides the `add` property in the config object.
  *
  * @return {Phaser.GameObjects.DynamicBitmapText} The Game Object that was created.
--- a/src/gameobjects/bitmaptext/static/BitmapTextCreator.js
+++ b/src/gameobjects/bitmaptext/static/BitmapTextCreator.js
@@ -18,9 +18,9 @@ var GetValue = require('../../../utils/object/GetValue');
  * @method Phaser.GameObjects.GameObjectCreator#bitmapText
  * @since 3.0.0
  *
- * @param {Phaser.GameObjects.BitmapText.Types.BitmapTextConfig} config - The configuration object this Game Object will use to create itself.
+ * @param {Phaser.GameObjects.BitmapText.Types.BitmapTextConfig} [config] - The configuration object this Game Object will use to create itself.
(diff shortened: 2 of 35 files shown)


photonstorm:
Thanks for this PR and trying to help add types for the creator methods. I hadn't done these yet because it's actually really complicated to accurately describe the config options! If you look at the Creator functions, you'll find that all Game Objects (except Particles and Graphics which are special cases) are passed to the `BuildGameObject` function. This is the core of the creation process and is what allows all of the properties to be set. If you look at that file you'll see there are _lots_ of additional properties that can be set, in lots of different ways. Scale, rotation, position, flip, etc.

Here is a valid set of configs for Sprites:

```
    //  Implicit values
    var config1 = {
        key: 'bunny',
        x: 400,
        y: 300
    };

    //  Implicit values
    var config2 = {
        key: 'bunny',
        x: 400,
        y: 300,
        flipX: true,
        flipY: true
    };

    //  An array means 'pick one element at random'
    var config3 = {
        key: [ 'bunny', 'atari', 'logo' ],
        x: 400,
        y: 300
    };

    //  randInt means pick a random integer between the min and max values given
    var config4 = {
        key: 'atari',
        x: { randInt: [ 0, 800 ] },
        y: 300
    };

    //  randFloat means pick a random float between the min and max values given
    var config5 = {
        key: 'logo',
        x: 400,
        y: 300,
        alpha: { randFloat: [ 0.1, 1 ] }
    };

    //  If the property is a function it will invoke it and use the result as the value
    var config6 = {
        key: 'atari',
        x: 400,
        y: function () {
            return 100 + Math.random() * 500;
        }
    };

    //  An example using randFloat to set the scale of a Sprite
    var config7 = {
        key: 'logo',
        x: 400,
        y: 300,
        scale: { randFloat: [ 0.2, 2 ] }
    };

    //  An example using randFloat to set independent x and y scale values
    var config8 = {
        key: 'logo',
        x: 400,
        y: 300,
        scale: { x: { randFloat: [ 0.2, 2 ] }, y: { randFloat: [ 1.5, 3 ] } }
    };
```

Notice the different formats a single property can take. `y` for example could be a direct number, an array of options to pick from, your own function, a built-in special function like `randFloat` and so on. This was the part I hadn't yet figured out how to do the types for! As this applies to all Game Objects it feels like it needs a base type, that their own special types inherit from.

G-Rath:
That is a lot of different options, but nothing too horrible.

I agree that a base type is likely the best way to go.
It might actually be worth trying to work backwards, by making the definition(s) in either TypeScript or JSONSchema, as that would give you instant validation on if the types are all compatible, and if you've missed something or not. 

It'd be pretty easy to then build the JSDocs from them, once you know the shape.

I'm happy to help with this where I can 😃

photonstorm:
Sure, feel free. It's something I'm going to tackle soon anyway, but I'll check this PR for updates before I do.

G-Rath:
It'll probably be better if you made the first attempt at this, since you know the library far better (and it's documentation structure) far better than me.

For now, I'll focus on the TypeScript Parser, but I'll let you know if that changes.

photonstorm:
Closing this PR off to tidy things up a bit.

Issue #4127, opened 2018-10-22
Phaser.Geom.Intersects functions


RollinSafary:
function `Phaser.Geom.Intersects.LineToLine` gets 3 arguments, and the 3rd argument is optional and it will set intersects point (`Phaser.Geom.Point`) in that. I think that it will be fine, if all intersects functions will have that 3rd argument to return `Phaser.Geom.Points[]` array.

florianvazelle:
Hi ! I'll work on it with @gglaive (It's perfect for our school project).

photonstorm:
Thank you for submitting this feature request. We have implemented this and the feature has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

PR #4384, opened 2019-02-20
Add out argument in intersect functions

florianvazelle:
To implement the feature requested by #4127, we modified the intersects functions in order to have an argument 'out' which allows the retrieval of the points of intersects.

The files `PointTo.`, `GetRectangleIntersection` and `RectangleToValues` didn't need to be modified.

We add in `CircleToCircle` and `LineToCircle` the calculation of equations to get the points of intersects. As for the other functions, we only need to call more basic functions. For example, `RectangleToTriangle` calls `LineToRectangle` which calls `LineToLine`.

When a line is compared with an other identical, the function return false. We kept this standard for `RectangleToRectangle`, `CircleToCircle` and `TriangleToTriangle`.

Note that when an intersects is found on a corner of a polygon, the point while be counted twice.
It's up to the user to filter the array to get rid of the duplicate.


photonstorm:
Thanks for the effort put into this PR, there's a lot of good stuff here. I've two points I'd like to raise, one is a change request, the other more of a question:

1) While, on the surface, this appears to add in _optional_ features (i.e. getting the point of intersection) it actually creates a huge amount of extra processing in each function, even if you don't need the intersects point. Which makes me think this feature might be better served as a set of entirely new functions, i.e. `GetCircleToCircle` which do exactly what they say, then the original functions can be retained for faster straight boolean checks.

2) The new code needs to pass our eslint tests and adhere to our coding styles (the sort that can't be enforced via linting). Glancing at the code I can see it doesn't do this. I can give more guidance on this once we decide what to do with the actual functions.

florianvazelle:
Thanks for the feedback, we are working on the first point while trying to respect the coding styles.

photonstorm:
Thank you, this looks much better. I've two very small change requests:

1) Please add yourself as the `@author` in the docblock at the top of each file.
2) Please remove the `package.lock` file from the commit.

Happy to merge it after this.

rexrainbow:
@florianvazelle 

Hi, I am interesting in these new features, and make a small testing of `GetCircleToCircle` method, it works great when position of two circle is not equal, but case [`y0 === y1`](https://github.com/photonstorm/phaser/blob/master/src/geom/intersects/GetCircleToCircle.js#L39) could not find intersect points.

Could you please check it again?

PR #4387, opened 2019-02-21
Add other lifecycle methods to scene class

josephmbustamante:

**This PR (delete as applicable)**

* Updates the Documentation

It adds the `init`, `preload`, and `create` methods to the documentation for the `Scene` class. It's based off of [this issue on the docs repository.](https://github.com/photonstorm/phaser3-docs/issues/91)

I tried a couple different ways of writing the JSDoc without actually including the empty methods, but none of them actually worked to generate the Typescript defs, so I ended up just adding the methods in. (I didn't do a deep dive into what exactly `tsgen` is looking for, though, so maybe there's a way to do it I missed)

### DIFF
--- a/src/scene/Scene.js
+++ b/src/scene/Scene.js
@@ -270,6 +270,39 @@ var Scene = new Class({
     /**
      * Should be overridden by your own Scenes.
      *
+     * @method Phaser.Scene#init
+     * @override
+     * @since 3.0.0
+     */
+    init: function ()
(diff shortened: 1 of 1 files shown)


photonstorm:
I figured out how to get the jsdocs to work even if there isn't any code associated with it, so this is now fixed properly in the main repo. Thank you for your PR though :) I'll list you in the change log!

PR #4400, opened 2019-02-26
Declared dataset as object for internet explorer 10

SirLink:
* Fixes a bug

Describe the changes below:

Internet explorer 10 crashed when trying to access the dataset property of the audio object.


photonstorm:
Can you please confirm this doesn't accidentally overwrite an existing `dataset` object that may already be present on Audio objects outside of IE10?

SirLink:
image
Tested in Chrome 73

image
IE 10

photonstorm:
Ok, let's risk it then.

PR #4439, opened 2019-03-23
Add force orientation support

m31271n:
> I am not sure that whether all users need this feature. But, in my situation, I surely need this. So, I took the liberty of writing this PR.

Back to Phaser 2, we have a feature call force orientation, which is described at [here](https://photonstorm.github.io/phaser-ce/Phaser.ScaleManager.html#forceOrientation). I think Phaser 3 should have this feature, too.

## Why we need force orientation?

When a game is running in general browser (for example, not in Cordava), and it should be displayed in Landscape Direction. But, when the user locks his / her phone in Portrait Direction, the game will be displayed like this following current policy of ScaleManager:

Above condition isn't what we want, apparently.

If user want to play this game in a better layout, he / she should disable orientation lock of phone, and rotate the phone is physical world, then the game should be displayed like this:

That's what I want to say: **In order to play a game which is in Landscape Direction and display it in better layout, user have to disable orientation lock in Phone's OS and rotate Phone by themselves**.

### When Phone's orientation lock is enabled

For a game should be displayed in Landscape Direction - [online demo](https://dodrio.github.io/phaser3-game-landscape-example/index.html):

For a game should be displayed in Partrait Direction - [online demo](https://dodrio.github.io/phaser3-game-portrait-example/index.html):

## Why not use `ScaleManager#lockOrientation`?

This API is implemented via `Screen.lockOrientation()`. But, as MDN said:

> This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the compatibility table at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.

In most of browser, it just call a depracated API without doing any other useful thing.

### Step 1: add a toggle

Add a toggle for the policy of force oriencation - `GameConfig#forceOrientation`.

### Step 2: implement the policy

Get expected layout of game according `game.config.width` / `game.config.height`:
+ when `game.config.width >= game.config.height`, `ScaleManager#isGameLandscape` is true.
+ when `game.config.width < game.config.height`, `ScaleManager#isGamePortrait` is true.

Get current layout of screen via `ScaleManager#isScreenLandscape` / `ScaleManager#isScreenLandscape`.

> Currently, `ScaleManager#isScreenLandscape` / `ScaleManager#isScreenLandscape` is simple and crude, i will improve it later.

Compare `ScaleManager#isScreenLandscape` / `ScaleManager#isScreenLandscape` with `ScaleManager#isGameLandscape` / `ScaleManager#isGamePortrait` to determine that if Phaser should:
+ rotate the canvas.
+ remap input onto a rotated canvas.

## Current progress

Complete.

> When input works, clicking green point in demo will print 'Oh!' in browser's console.
> And the input is broken when the canvas is rotated. 

> @photonstorm After reading the code of `InputManager`, I still have no idea where is the best place to remap input. Maybe you can give me some tips. ;)

> Update: Found the way to `remap input onto a rotated canvas`, working on it.

> `remap input onto a rotated canvas` is done.

### DIFF
--- a/src/core/Config.js
+++ b/src/core/Config.js
@@ -77,6 +77,11 @@ var Config = new Class({
         this.scaleMode = GetValue(config, 'scaleMode', 0);
 
         /**
+         * @const {boolean} Phaser.Core.Config#forceOrientation - Toggle force orientation for the Scale Manager. The default is false, which means closed.
+         */
+        this.forceOrientation = GetValue(config, 'forceOrientation', false);
+
+        /**
          * @const {boolean} Phaser.Core.Config#expandParent - Is the Scale Manager allowed to adjust the CSS height property of the parent to be 100%?
          */
         this.expandParent = GetValue(config, 'expandParent', true);
--- a/src/input/InputManager.js
+++ b/src/input/InputManager.js
@@ -1634,8 +1634,8 @@ var InputManager = new Class({
         p1.y = p0.y;
 
         //  Translate coordinates
-        var x = this.scaleManager.transformX(pageX);
-        var y = this.scaleManager.transformY(pageY);
+        var x = this.scaleManager.getAxisX(pageX, pageY);
+        var y = this.scaleManager.getAxisY(pageX, pageY);
 
         var a = pointer.smoothFactor;
 
--- a/src/scale/ScaleManager.js
+++ b/src/scale/ScaleManager.js
@@ -189,6 +189,8 @@ var ScaleManager = new Class({
          */
         this.parentSize = new Size();
 
+        this.parentOriginalSize = new Size();
+
         /**
          * The Game Size component.
          * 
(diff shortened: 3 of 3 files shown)


m31271n:
> Lots of people email me about this PR, so I wrote following comment.

After I wrote the PR, I realized the PR is too big, so I closed it and try another method to extend Phaser.

# Patched Phaser
At that time, Phaser's version is 3.16.2. I patch it with external SceneManager support(which controls orientation in Phaser). The code is [in custom branch](https://github.com/2players/phaser/commits/custom) with these two commit:
+ feat: add support for using external ScaleManager
+ fix: deactive Pointer when its identifier is not matched

Patched Phaser will use custom SceneManager when `scaleManagerClass `is set in the configuration.

You can modify current version of Phaser by referencing these two commits.

# Custom SceneManager
After patching Phaser,I created a custom SceneManager provides the orientation lock at [here](https://github.com/2players/phaser3-harmony/tree/master/src/scale).

You can create your own SceneManager by referencing above path.

# Last

At that time, Phaser's WebGL performance isn't well on some Android device. Under the pressure of project schedule，I fallback to Pixi.js, and the idea of extending Phaser is put aside.

Maybe, the performance problem is solved now. You can give it a try.

And if you like, give it back to the community. Good Luck!

Issue #4259, opened 2018-12-26
Color objects should be accepted wherever an API takes a color string


BigZaphod:
Phaser has a color object `Phaser.Display.Color` but many (most?) of the APIs that accept colors assume the color is formatted as a string and you cannot pass a color object instance to them. I first noticed this with `TextStyle`'s fill and stroke color properties, but it appears to be a common pattern throughout the codebase. There's an easy workaround, of course, which is to use the `rgba` property of the color instance in this case, but that feels wrong to have to do all the time. I think the various color APIs should be audited and updated to accept either a string or a Color object.

LoolzRules:
Hi, I'm going to spend my time on this problem

photonstorm:
Thank you for opening this feature request (all those months ago!). As the development of Phaser 4 has now started I am tidying-up the repo and closing off old issues that won't be directly fixed or implemented in Phaser 3 by me. I would still consider community PRs that address this issue, however, and have left the GitHub tags intact so others can track them if they wish to get involved. However, I'm afraid this is no longer a feature that will be added to v3.

PR #4445, opened 2019-03-25
issue #4259 fix

LoolzRules:

* Adds a new feature

Describe the changes below:

Attempt to fix issue described here: https://github.com/photonstorm/phaser/issues/4259


photonstorm:
Thanks for spending the time opening this PR. However, I'm going to close it without merging because I don't feel it fixes the relevant issue in depth. There are _lots_ of places where colors are taken that need updating and it's quite a big and involved task, so it wouldn't be right to mark the issue as fixed when only implemented in a couple of places. There needs to be a more consistent way to define and extract color values.

PR #4448, opened 2019-03-28
fix: Inconsistent return in Tween#pause()

kyranet:
* Fixes a bug

Describe the changes below:

Calling `tween.pause();` returns the instance of `Tween`, however, if it was already paused, it would return `undefined`, causing problems when chaining the methods, e.g. `tween.pause().seek(0.1);` while the tween was paused.


photonstorm:
👍

PR #4449, opened 2019-03-28
fix: Inconsistent return in TweenManager#makeActive()

kyranet:
* Fixes a bug

Describe the changes below:

Calling `TweenManager#makeActive(tween);` returns the TweenManager instance itself, however, if you create a tween externally (since those aren't in the add or active lists) and try to call `makeActive()` with it, this would return `undefined`.


photonstorm:
👍

PR #4455, opened 2019-04-02
Fixed Phaser.Actions.Spread ignoring the min parameter

galman33:

* Fixes a bug

Describe the changes below:
 
* Fixed Phaser.Actions.Spread ignoring the min parameter

### DIFF
--- a/src/actions/Spread.js
+++ b/src/actions/Spread.js
@@ -40,14 +40,14 @@ var Spread = function (items, property, min, max, inc)
     {
         for (i = 0; i < items.length; i++)
         {
-            items[i][property] += i * step;
+            items[i][property] += i * step + min;
         }
     }
     else
     {
         for (i = 0; i < items.length; i++)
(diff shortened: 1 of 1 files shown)


src/actions/Spread.js
```diff
@@ -40,14 +40,14 @@ var Spread = function (items, property, min, max, inc)
     {
         for (i = 0; i < items.length; i++)
         {
-            items[i][property] += i * step;
+            items[i][property] += i * step + min;
```
jorbascrumps: Should `min` be initialized with a default value (perhaps 0?) to prevent `NaN` result when `min` is omitted?

galman33: I believe min isn't an optional parameter (?)

jorbascrumps: My apologies! You are correct.

photonstorm:
👍

PR #4469, opened 2019-04-10
Docs for scene config and optional scene methods

samme:
This PR

* Updates the Documentation

Context: <https://phaser.discourse.group/t/documentation-for-the-scene-object-in-gameconfig/2005>

I added `Phaser.Scenes.Types.CreateSceneFromObjectConfig`. As I understand it, the scene config object is the union of that and the existing `Phaser.Scenes.Types.SettingsConfig`.

I also expanded the docs for the extra scene methods init, preload, create (#4387) but these don't appear after I do `npm run gen`. Maybe they need their own types. That might even make more sense since technically Phaser.Scene doesn't define those methods.

I removed `@override` for all the scene methods since they don't override a parent class.

### DIFF
--- a/src/core/Config.js
+++ b/src/core/Config.js
@@ -17,7 +17,7 @@ var ValueToColor = require('../display/color/ValueToColor');
 
 /**
  * @classdesc
- * The active game configuration settings, parsed from a {@link GameConfig} object.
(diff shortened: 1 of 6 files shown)


photonstorm:
👍

Issue #4456, opened 2019-04-02
Text interactive wrong area when calling setFontSize() after setInteractive()


thanh-taro:
## Version

* Phaser Version: 3.16.2-FB

* Operating system: Mac OS Mojave 10.14.4

* Browser: Chrome v73.0.3683.86 (Official Build) (64-bit)

## Description

Text interactive wrong area when calling setFontSize() after setInteractive()

## Example Test Code

```js
var config = {
    type: Phaser.AUTO,
    parent: 'phaser-example',
    width: 800,
    height: 600,
    scene: { create: create }
};

var game = new Phaser.Game(config);

function create ()
{
    var text = this.add.text(10, 50, 'hover me and watch the cursor change');
    text.setInteractive({ cursor: 'pointer' });
    text.setFontSize(22);
}

```

## Additional Information

I think the reason is the interactive area did not update when changing font size.

thanh-taro:
No one cares?

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

PR #4476, opened 2019-04-18
Update rectangle hit-area when updated text

rexrainbow:

* Fixes a bug

Describe the changes below:

Update rectangle hit-area when updated text. (#4456)
(This code is modified from [Zone game object](https://github.com/photonstorm/phaser/blob/master/src/gameobjects/zone/Zone.js#L168-L172))


photonstorm:
Thanks for the PR. I've fixed this in a slightly different way, because if they set their own custom hitArea this shouldn't touch it. A new property in InteractiveObject allows for this (and removes the need for the instanceof check too).

Issue #4471, opened 2019-04-14
Scene plugins cannot receive configuration data


TJ09:
## Version
* Phaser Version: 3.16.2

## Description

ScenePlugin inherits the [documentation of the `init()` method](https://photonstorm.github.io/phaser3-docs/Phaser.Plugins.ScenePlugin.html#init__anchor) from BasePlugin, which says that init will be called with the data passed to the plugin config.

Similarly, the [`PluginObject`](https://photonstorm.github.io/phaser3-docs/global.html#PluginObject) type says that the scene config is an array of [`PluginObjectItem`](https://photonstorm.github.io/phaser3-docs/global.html#PluginObject), which has a field called `data` that is "*[a]rbitrary data passed to the plugin's init() method*."

Unfortunately, neither of these are true. PluginManager constructs scene plugins but never actually passes data to them:
https://github.com/photonstorm/phaser/blob/974ceb5654edb9db552b1343fc2bb3f7f69d464d/src/plugins/PluginManager.js#L257-L275
unlike the global case where `init` is indeed called:
https://github.com/photonstorm/phaser/blob/974ceb5654edb9db552b1343fc2bb3f7f69d464d/src/plugins/PluginManager.js#L576-L589

## Example Test Code

```js
var TestBasePlugin = new Phaser.Class({
    Extends: Phaser.Plugins.BasePlugin,

    init: function(data) {
        console.log('TestBasePlugin initialized with', data);
    }
});

var TestScenePlugin = new Phaser.Class({
    Extends: Phaser.Plugins.ScenePlugin,

    init: function(data) {
        console.log('TestScenePlugin initialized with', data);
    }
});

var game = new Phaser.Game({
    type: Phaser.HEADLESS,
    parent: 'phaser-example',
    plugins: {
        global: [
            { key: 'TestBasePlugin', plugin: TestBasePlugin, start: true, data: "test" },
        ],
        scene: [
            { key: 'TestScenePlugin', plugin: TestScenePlugin, mapping: 'testScenePlugin', start: true, data: "test" },
        ],
    },
    scene: {
        create: function() {
            console.log('Scene created', this.testScenePlugin);
        },
    }
});
```

### Expected console output
```
TestBasePlugin initialized with test
TestScenePlugin initialized with test
Scene created ...
```

### Actual console output
```
TestBasePlugin initialized with test
Scene created ...
```

samme:
[...]

PR #4488, opened 2019-04-24
Remove BasePlugin#boot; and plugins docs updates

samme:
* Removes BasePlugin#boot
* Docs updates for BasePlugin and ScenePlugin

In particular, only global plugins use `init()`, `start()`, `stop()` and only scene plugins use `boot()`, `scene`, `systems` (#4471).

I removed BasePlugin#boot since Phaser never calls it for global plugins, as far as I can tell. That could be a breaking change if someone has made a scene plugin by extending BasePlugin only and their plugin class doesn’t define its own boot() method. But I expect most people are extending ScenePlugin.

Should I remove BasePlugin#scene and BasePlugin#systems as well?


photonstorm:
👍

PR #4498, opened 2019-04-28
bugfix/camera-pan-destination

digitsensitive:
* Fixes a bug

Describe the changes below:
Camera pan did not work correct for me, moving in the opposite destination direction.
For me it worked after this small change.


photonstorm:
Please post a code sample showing how this fixes the pan movement, as none of the pan examples are broken with the current implementation and all act as expected.

photonstorm:
Closing due to inactivity. Happy to look at this again with an example.

PR #4499, opened 2019-04-28
fix: Account for container rotation when dragging contained object

### LINKED ISSUE #4437: Dragging an object in a rotated container does not work as expected
## Version

* Phaser Version: 3.16.2

* Browser:

## Description

Dragging an object in a rotated container does not work as expected. The `dragX` and `dragY` parameters to the `'drag'` input function correctly account for the `x` and `y` position of the container, but not its `rotation`.  As a result, when dragging from a rotated container using the typical `object.x = dragX; object.y = dragY;`, the object will not follow the cursor. Furthermore, it is not sufficient to perform 2D vector rotation on the inputs, i.e.: 
```
gameObject.x = dragX * Math.cos(rot) + dragY * Math.sin(rot);
gameObject.y = dragY * Math.cos(rot) - dragX * Math.sin(rot);
```
because it will cause the initial position of the object to jump to a predictable location. I imagine the workaround would have to somehow involve the cursor position on `'dragstart'`, which is doable, but not ideal compared to a hypothetical implementation that automatically accounted for the rotation in the `dragX` and `dragY` parameters.

## Example Test Code

```
// Class definitions elided so as not to clutter the example, but I can try to
// provide a complete, self-contained, working example if you think it's necessary.
const obj = new objectThatInheritsFromGameObject(this, {});
const cont = new objectThatInheritsFromContainer(this, {rotation: Math.PI});
this.add.existing(obj);
obj.setInteractive();
cont.add(obj);
this.input.setDraggable(card);
this.input.on('drag', (pointer, gameObject, dragX, dragY) => {
    gameObject.x = dragX;
    gameObject.y = dragY;
});
```
## Additional Information

[Naive drag](https://i.imgur.com/fUb1R4W.gifv)
[Drag with trigonometric transformation](https://i.imgur.com/vGWg3eX.gifv)

aliblong:
This PR:

* Fixes a bug

Describe the changes below:

Fixes #4437. Objects in rotated containers can now be dragged normally.

### DIFF
--- a/src/input/InputPlugin.js
+++ b/src/input/InputPlugin.js
@@ -1099,11 +1099,19 @@ var InputPlugin = new Class({
 
                 input.dragState = 2;
 
-                input.dragX = pointer.x - gameObject.x;
-                input.dragY = pointer.y - gameObject.y;
-
+                // For contained objects, drag global coordinates need to be transformed to local coordinates
                 input.dragStartX = gameObject.x;
                 input.dragStartY = gameObject.y;
+                input.dragStartXGlobal = pointer.x;
+                input.dragStartYGlobal = pointer.y;
+
+                input.dragX = input.dragStartXGlobal - input.dragStartX;
+                input.dragY = input.dragStartYGlobal - input.dragStartY;
+
+                // This will be used to rotate the drag vector for objects in rotated containers
+                var rotation = this.getContainerRotation(gameObject);
+                input.sin = Math.sin(rotation);
+                input.cos = Math.cos(rotation);
 
                 gameObject.emit(Events.GAMEOBJECT_DRAG_START, pointer, input.dragX, input.dragY);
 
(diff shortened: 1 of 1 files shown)


photonstorm:
Thanks for this PR. I appreciate the time you have put into it. _As it stands_, I cannot merge it for the following, easily fixed, reasons:

1) Please do not create new properties on an object (`input.sin` for example) unless essential. If they should be permanent properties, they should have their own entry in the parent object and be typed and documented as such. Looking at it, I can't see a reason to store `sin` and `cos` beyond the function in which they're used. They could easily be temporary vars.

2) `getContainerRotation` has no place in the InputPlugin class. This needs to live somewhere else, idealy in the Container class itself (with a different name)

aliblong:
Thanks for your feedback! Regarding your two points:

1. The reason I store `sin` and `cos` of the container's rotation as a property of the input is that if I compute it in the scope where it's used, i.e. for each `drag` event, the computation is being redone pointlessly hundreds of time. Haven't profiled it or anything, but that seems like premature pessimization. Is there some better alternative for passing the computed values into that scope? Just using a `var` and doing the computation and accesses in separate branches doesn't seem to work (and is maybe a race condition?).

2. How about I put it in the `gameObject` class? That seems more appropriate, since the operational object is a `gameObject`, which needn't even have a container.

photonstorm:
Thanks for this PR. I have taken bits of the code and merged it into the Input Plugin class, so the end result is the same (if a Game Object is in a rotated Container, it'll update the drag values accordingly), which fixes the issue in question and provides the feature. I can't merge this PR as the input system has changed entirely in 3.18, but the principal is there and you are credited in the Change Log.

PR #4509, opened 2019-05-03
Add an alternative option for including typings

atkinchris:
This adds an alternative option for people using Phaser to include Typescript typings into their application, without having to include an external file in their committed source code.


photonstorm:
Thanks for this PR, I'm going to close it because as of 3.17 the defs now live in this repo and are referenced in the package.json, so will work directly now.

PR #4517, opened 2019-05-09
JSDoc fix returned object type

antoine-pous:
This PR

* Updates JSDoc

Describe the changes below:
Fixed JSDoc @return type. With it we should be able to use metrics without having to cast them each time.


photonstorm:
👍

PR #4519, opened 2019-05-11
Fixed separateCircle cause the position set NaN.

hizzd:

* Fixes a bug

Describe the changes below:

https://github.com/photonstorm/phaser/blob/v3.17.0/src/physics/arcade/World.js#L1470-L1471
```javascript
var nx = (body2.position.x - body1.position.x) / d;
var ny = (body2.position.y - body1.position.y) / d;
```
If (body2.position.x - body1.position.x) is equal to 0 and 'd' is equal to 0, then it is 0 / 0 = NaN.

https://codepen.io/hizzd/pen/zQBrbz?editors=0010
In this example, if the positions of the two sprites are the same, the position will be updated to NaN.


photonstorm:
👍

PR #4523, opened 2019-05-13
most editable car

ysraelJMM:
the possibility of having a car ready to use, especially when it is small. It generated problems when the radius of the wheels was less than 20.

https://drive.google.com/file/d/0B3y9mOwx62BSMWJNS0MwUl8zM2VBbmUwbEt4bl96WmsxREJR/view


photonstorm:
Thank you for this PR and spending time opening it. However, this PR changes the Matter.js source code directly, and not Phaser. So I think it would make a lot more sense to submit this PR to matter directly and ask them to merge it. Once they have, we'll get it when they release a new version.

PR #4524, opened 2019-05-14
Use `customHitArea` in condition

rexrainbow:

does not add a new feature, not fix a bug.

It uses new `customHitArea` property (added in 3.17.0) in hitArea checking condition, like text game object.


photonstorm:
Nice idea 👍

PR #4539, opened 2019-05-18
Removes arrow function syntax

### LINKED ISSUE #4530: Phaser 3.17 breaks on IE 11
## Version

* Phaser Version: 3.17.0

* Operating system: Any windows OS

* Browser: IE 11

## Description

the following commit https://github.com/photonstorm/phaser/commit/e0f2b829f87567e3ccfb31bae17f12d802dddaf4#diff-bec4cf8bfff946115edd34836fa564bcR1192 contains es6 arrow functions that make their way into the final dist files. The es6 syntax breaks older browsers (in my case ie11) where phaser 3.16.2 was working.

## Example Test Code

Execute the phaser.js file from the published dist directory in IE11 or any other older browser that does not support es6 and the bug will be reproduced.

## Additional Information
I would suggest processing the src files with either typescript or babel when producing the dist folder and changing the main in the package.json to point to `dist/phaser.js` instead of src/phaser.js` . 

I would be willing to make the change to the build pipeline myself and submit a pr if that is desired.

jorbascrumps:

* Fixes a bug

Describe the changes below:

Removes arrow function syntax, which was breaking in unsupported browsers. Resolves #4530

### DIFF
--- a/src/scale/ScaleManager.js
+++ b/src/scale/ScaleManager.js
@@ -1203,13 +1203,15 @@ var ScaleManager = new Class({
             {
                 if (fullscreen.keyboard)
                 {
-                    //  eslint-disable-next-line es5/no-arrow-functions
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

PR #4545, opened 2019-05-21
Warn when plugin entry is missing `plugin` property

samme:
This PR

* Adds a new feature

Adds a console warning when PluginManager skips installing a plugin (during boot) because the `plugin` value is missing or empty. IMHOP this change should help users catch this mistake and is also consistent with the warnings given by PluginManager.install() and PluginManager.installScenePlugin().


photonstorm:
👍

PR #4552, opened 2019-05-23
Set canvas size equal to text object size

rexrainbow:

* Fixes a bug

Describe the changes below:

Canvas size might less than text object size, when `fixedWidth` and `fixedHeight` is set.


mirzadzindo1:
Ok

PR #4554, opened 2019-05-23
Remove GroupClassTypeConstructor type

samme:
This PR

* Fixes a bug (TypeScript definitions)

The `GroupClassTypeConstructor` type (#4150) would give errors like this:

```
Argument of type '{ classType: typeof Sprite; }' is not assignable to parameter of type 'GameObject[] | GroupConfig | GroupConfig[]'.
  Type '{ classType: typeof Sprite; }' is not assignable to type 'GroupConfig'.
    Types of property 'classType' are incompatible.
      Type 'typeof Sprite' is not assignable to type 'GroupClassTypeConstructor'.
        Type 'typeof Sprite' provides no match for the signature '(scene: Scene, x: number, y: number, texture: string, frame?: string | number): void'.
```

I removed the type and replaced it with `{Function}`.


mirzadzindo1:
Good

PR #4560, opened 2019-05-28
fix addCanvas Memory Leak

atbulbs:

* Fixes a bug

Describe the changes below:
src /  textures/ TextureManager.js

in the addCanvas function, there is a variable `texture` 
```javascript
var texture = null;
```
this variable will cause a reference to CanvasTexture instance, so Memory Leak happened

I test it by chrome dev tool - Memory snapshots

### DIFF
--- a/src/textures/TextureManager.js
+++ b/src/textures/TextureManager.js
@@ -474,7 +474,6 @@ var TextureManager = new Class({
         }
 
         return null;
-        
     },
 
     /**
--- a/src/textures/TextureManager.js
+++ b/src/textures/TextureManager.js
@@ -460,7 +460,6 @@ var TextureManager = new Class({
     {
         if (skipCache === undefined) { skipCache = false; }
 
-
         if (skipCache)
         {
             return new CanvasTexture(this, key, source, source.width, source.height);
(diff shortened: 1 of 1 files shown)


photonstorm:
It's a local variable, it won't persist beyond the end of the function, where it is returned. I cannot see how this causes any memory leaks, it's the same pattern used through-out the whole API. Going to need more evidence to prove this does anything before I merge it, sorry.

atbulbs:
> It's a local variable, it won't persist beyond the end of the function, where it is returned. I cannot see how this causes any memory leaks, it's the same pattern used through-out the whole API. Going to need more evidence to prove this does anything before I merge it, sorry.

I create a demo to describe our team's problem, please read it and wish your replay.

[https://github.com/atbulbs/phaser_memory_leak_test/tree/master](https://github.com/atbulbs/phaser_memory_leak_test/tree/master)

photonstorm:
Thank you for making the test. However, this PR is changing code that isn't even used in that test, and doesn't fix anything. So please, can you open an issue requesting assistance with `Game.destroy`, which is what the actual issue is, and I'll close off this PR.

PR #4577, opened 2019-06-05
Add Group#name

samme:
This PR

* Adds a new feature

Adds a `name` property for Groups. Can be set via GroupCreateConfig.


src/gameobjects/group/typedefs/GroupConfig.js
```diff
@@ -3,6 +3,7 @@
  * @since 3.0.0
  *
  * @property {?Function} [classType=Sprite] - Sets {@link Phaser.GameObjects.Group#classType}.
+ * @property {?string} [name=''] - Sets Sets {@link Phaser.GameObjects.Group#name}.
```
jorbascrumps: Looks like some duplicate verbiage here :)

photonstorm:
Thanks. Have fixed the duplicate text directly.

PR #4579, opened 2019-06-07
Correct emitOnly values for particle emitter angle, lifespan, x, y

samme:
This PR

* Updates the Documentation
* Fixes a bug

I changed these to emitOnly=true because, judging from [Particle#update](https://photonstorm.github.io/phaser3-docs/Phaser.GameObjects.Particles.Particle.html#update), they're not actually updatable. (`alpha`, `rotate`, `scaleX`, `scaleY`, and `tint` are the only updatable emitter properties.)


photonstorm:
👍

Issue #4123, opened 2018-10-19
Definition of size in circle shape and rectangle shape


rexrainbow:
Size in circle shape is [`this.setSize(this.geom.radius, this.geom.radius)`](https://github.com/photonstorm/phaser/blob/master/src/gameobjects/shape/arc/Arc.js#L112), in rectangle is [`this.setSize(width, height)`](https://github.com/photonstorm/phaser/blob/master/src/gameobjects/shape/rectangle/Rectangle.js#L57).
It seems that definition of width and height in circle shape is not equal to rectangle shape, assume that
- Circle with radius 15 -> diameter = 30, and
- Rectangle 30x30
should have the same width and height.

How about using [`GetAABB()` to get width and height](https://github.com/photonstorm/phaser/blob/master/src/gameobjects/shape/polygon/Polygon.js#L69) in all shape game object? because that all shapes object are points(polygon) at last.

PR #4586, opened 2019-06-11
Set (width,height) of arc/circle to (diameter,diameter)

rexrainbow:

* Fixes a bug

Describe the changes below:

Set (width,height) of arc/cricle to (diameter,diameter), not (radius,radius). #4123


photonstorm:
👍

PR #4587, opened 2019-06-11
Fix EmitterOp config with { random: false }

samme:
This PR

* Fixes a bug

The docs imply that an EmitterOp value like

    { start: 0, end: 1, random: false }

should be identical to 

    { start: 0, end: 1 }

(i.e., nonrandom) but in practice it is identical to

    { start: 0, end: 1, random: true }


photonstorm:
👍

PR #4593, opened 2019-06-14
Fix an error when destroying a scene during create()

samme:
This PR

* Fixes a bug

This lets you call `this.scene.remove()` at the end of a scene's create() method without causing an error.


photonstorm:
I don't mind merging this in, because it's good to capture it, but this is such a terrible thing to do I hope no-one ever needs it!

samme:
I've done this in a boot scene. 😄 

I guess you can do it in update() instead, but that looks weird to me.

PR #4608, opened 2019-06-22
Set customHitArea to false if shape is not defined

rexrainbow:

* Fixes a bug

Describe the changes below:

In [setHitArea](https://github.com/photonstorm/phaser/blob/master/src/input/InputPlugin.js#L2087) method, use [setHitAreaFromTexture](https://github.com/photonstorm/phaser/blob/master/src/input/InputPlugin.js#L2129) if shape is undefined. Suppose that it is *not* a case of `customHitArea`


photonstorm:
Agreed 👍

PR #4609, opened 2019-06-22
Fixed zero-length constraint and API inconsistency

olilanz:

* Fixes a bug

Describe the changes below:

The constraint API no longer accepts zero-length constraints in 3.18. This broke my game, where I use zero-length constraints to attach sensors to my bodies.

I opted for fixing the problem by making the length parameter mandatory (as it was prior to 3.18), and dropping the check for zero.

An alternative fix could have been to keep optionality, and alter the check to accept zero.

Comparing the API to other constraint APIs (e.g. worldConstraint, spring, joint) it appears that the length parameter was not optional there either. I opted for API consistency. 

Hope this helps.

### DIFF
--- a/src/physics/matter-js/Factory.js
+++ b/src/physics/matter-js/Factory.js
@@ -454,11 +454,7 @@ var Factory = new Class({
         options.bodyA = (bodyA.type === 'body') ? bodyA : bodyA.body;
         options.bodyB = (bodyB.type === 'body') ? bodyB : bodyB.body;
 
-        if (length)
-        {
-            options.length = length;
-        }
-
+        options.length = length;
         options.stiffness = stiffness;
 
         var constraint = Constraint.create(options);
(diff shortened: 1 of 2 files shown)


photonstorm:
Thanks for your PR and finding this bug. I've opted for a different fix, which is to just check if the given `length` `isNaN` or not and set it if so. This allows you to pass zero lengths, or pass `null` in order to have the length calculated automatically for you, which is arguably more useful for certain constraints.

So, I won't merge this PR, but I've made the change to the code so the fix is in and you've been credited with it in the change log.

Also, in future, please don't edit the TypeScript defs. The defs are automatically generated from the JSDocs, so any manual edits to them are lost every time the new defs files are built. The correct way to fix the defs is to fix the underlying JSDocs causing the error, then we win on all fronts, as both the docs and the defs are fixed in one move :)

Issue #4616, opened 2019-06-25
Loader XHR headers not set


jorbascrumps:
## Version

* Phaser Version: 3.18.1

* Operating system: macOS 10.14.5

* Browser: Chrome 75.0.3770.100 

## Description

Prior to v3.7.1, it was possible to set XHR headers on Loader requests. It seems that this functionality broke during the Loader updates that were performed for that version. Setting a global header value in the game config also doesn't work.

## Example Test Code

```js
this.load.tilemapTiledJSON('map', 'map.json', {
    header: 'Authorization',
    headerValue: 'Bearer t0k3n'
});
```

## Additional Information

I'm unsure if this functionality was meant to be removed as it's not mentioned in the [LoaderConfig](https://github.com/photonstorm/phaser/blob/master/src/core/typedefs/LoaderConfig.js); however, remnants of it remains in the codebase. If fixed, I would love to investigate the possibility of allowing more than one value be set.

jorbascrumps:
My mistake: my server needed to have the `Access-Control-Allow-Headers` header set.

I'm open to attempting the multiple headers feature if there's some interest.

PR #4618, opened 2019-06-25
Adds support for multiple custom Loader headers

jorbascrumps:

* Adds a new feature

Describe the changes below:

Previously, it was only possible to pass a single custom header to a Loader through `header` and `headerValue`. This adds the ability to pass multiple headers via a new `headers` property:

```js
this.load.image('sprite', 'assets/sprite.png', {
    headers: {
        Authorization: 'Bearer t0k3n',
    },
});
```
References my adventures in #4616

### DIFF
--- a/src/loader/XHRLoader.js
+++ b/src/loader/XHRLoader.js
@@ -30,6 +30,14 @@ var XHRLoader = function (file, globalXHRSettings)
     xhr.responseType = file.xhrSettings.responseType;
     xhr.timeout = config.timeout;
 
+    if (config.headers)
+    {
+        for (var key in config.headers)
+        {
+            xhr.setRequestHeader(key, config.headers[key]);
+        }
+    }
+
     if (config.header && config.headerValue)
     {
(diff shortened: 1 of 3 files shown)


photonstorm:
I'm happy with this feature, but do we really need `header` _and_ `headers` ? It feels like they ought to be merged. I.e. detect if an object is passed in and extract if so, otherwise use the single value.

jorbascrumps:
Definitely agree, @photonstorm! Happy to make this change but wasn't sure how well it would be received if I just proposed an API change out of nowhere.

To clarify: would you prefer _removing_ `header` in favour of `headers`, or keep the existing `header`? If the former, `headerValue` will have to be removed as well.

photonstorm:
I would prefer to remove `header` and keep the new `headers` property.

jorbascrumps:
@photonstorm Great! I'll make that adjustment and let you know.

photonstorm:
Have merged it as-is 👍

PR #4623, opened 2019-06-27
New types for the optional scene methods

samme:
This PR

* Updates the Documentation

This makes the CreateSceneFromObjectConfig definitions more accurate and should finally add the extra scene methods to the JSDocs (#4469).


photonstorm:
👍 👍 👍

Issue #4620, opened 2019-06-26
Arc/Circle Position wrong


CipSoft-Components:
## Version

* Phaser Version: v3.18.1-FB

* Operating system: Windows 10

* Browser: Tested in Chrome, FF, Chrome (Android)

## Description

When drawing a circle (or arc) with <Scene>.add.circle(), it is placed wrong. It is placed to far left and top. After some tests I can say it is placed 0.5 times of its radius in negative x and y direction, from the expected position.

## Example Test Code

Placing a circle in the left top corner, to fit exactly, without setting the origin.
Expectation:
`this.add.circle(200, 200, 200, 0xff0000);`
What is needed in this Version:
`this.add.circle(300, 300, 200, 0xff0000);`

## Additional Information

See your own example:
[http://labs.phaser.io/view.html?src=src/game%20objects/shapes/blend%20modes.js&v=3.18.1](http://labs.phaser.io/view.html?src=src/game%20objects/shapes/blend%20modes.js&v=3.18.1)

Btw. it would be nice, if the version could also be set in the editor for the examples.

jorbascrumps:
I believe this is intended behaviour in order to be consistent with other objects (which all default their origins to 0,0). If you need to position your circle from its centre then change the origin.

The editor version can be set, just change the query parameter.

telinc1:
Actually, [see the test case in this pen](https://codepen.io/telinc1/pen/RzpXrV?editors=0010). The circle's origin is not changed and its coordinates are at the center of the canvas, but it is not rendered where you'd expect. Additionally, it only happens with the WebGL renderer. If you set the `type` in the config to `Phaser.CANVAS`, the circle will appear where it's supposed to.

I suspect that the indirect cause of this is commit 382afd6, coupled with [the offset applied in the `updateData` method](https://github.com/photonstorm/phaser/blob/v3.18.1/src/gameobjects/shape/arc/Arc.js#L350) of the Arc, however, I haven't looked much further into this.

adrianolobo:
I can confirm that change of position in circles.
My app before the changes:

![image](https://user-images.githubusercontent.com/13997867/60273821-4f707b00-98cd-11e9-81d6-2b02ffd89558.png)

after 3.18

![image](https://user-images.githubusercontent.com/13997867/60273838-60b98780-98cd-11e9-9f8b-43ba4933227f.png)


Both of them are `setOrigin(0.5)`

Chumper:
[...]

PR #4626, opened 2019-06-28
Fix render offset issue

rexrainbow:

* Fixes a bug

Describe the changes below:

- Fix render offset issue (#4620)
- Fix set size issue when setting radius


photonstorm:
👍

PR #4628, opened 2019-06-29
Fix error destroying a scene in HEADLESS mode

### LINKED ISSUE #4467: Multi scene bug in HEADLESS mode
## Version 
* Phaser Version: 3.16.2
* Operating system: Windows 8

## Description

When running phaser in headless mode and having more than one scene, Phaser throws the error "Cannot read property 'gl' of null" in the prerendering function. It seems to be calling the wrong functions in headless mode when there is more than one scene. 

Headless mode is great for running multiplayer games on the server, and I love to split up the game in multiple scenes, so fixing this bug would be greatly appreciated. 

## Example Test Code

https://codesandbox.io/s/llj8y70l0z

## Additional Information

image

samme:
This PR

* Fixes a bug

Fixes #4467

### DIFF
--- a/src/cameras/2d/Camera.js
+++ b/src/cameras/2d/Camera.js
@@ -42,7 +42,7 @@ var Vector2 = require('../../math/Vector2');
  * @memberof Phaser.Cameras.Scene2D
  * @constructor
  * @since 3.0.0
- * 
+ *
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

PR #4646, opened 2019-07-10
Remove event hook of children when destroyChildren is false

rexrainbow:

* Fixes a bug

Describe the changes below:

Previously, DESTROY event hook of children won't be removed when destroyChildren is false. 
It will fire error message when destroy children after group destroyed, for example,
```javascript
var txt = this.add.text(0, 0, 'abc');
this.add.group().add(txt).destroy(false);
txt.destroy();
```

Now the DESTROY event hook will be removed in all cases.

### DIFF
--- a/src/gameobjects/group/Group.js
+++ b/src/gameobjects/group/Group.js
@@ -1134,22 +1134,7 @@ var Group = new Class({
             return;
         }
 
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

PR #4668, opened 2019-07-21
Fix typings for new Game scene options

posva:

* Fixes a bug

Describe the changes below:

This fixes typings for TS users and also improves autocompletion for JS users. It prevents users the need to manually type their functions:

```ts
const game = new Phaser.Game({
  scene: {
    preload(this: Phaser.Scene) {
    },
    create(this: Phaser.Scene) {
    }
  }
})
```

**By saving the file I removed many extra whitespace, that's why there are so many modifications** (I recommend turning the hide whitespace option to better see what I actually changed)

While editing the file I had a hard time doing searches in the file because it's really big. I also saw many `see {@link ...}` that where broken because the documentation was missing. For example, functions `init`, `create` and `preload` are missing.
I think it would be a good idea to split up the file in a similar folder structure that src, so it's easy to keep it updated


photonstorm:
Hi - Thank you so much for spending time doing this! However, I'm really sorry that I cannot merge your PR. The TypeScript definitions are automatically generated from the JSDocs within the source code, which means if I merge your PR your changes will be lost the next time the defs are automatically created. Please see the README for more details.

posva:
Sorry, I indeed didn't read that part of the readme. I gave this another try by updating the source files but it doesn't seem to generate correct typings. I tried changing the SceneInitCallback by providing a `@this` in the JsDocs

```diff
/**
 * Can be defined on your own Scenes.
 * This method is called by the Scene Manager when the scene starts, before `preload()` and `create()`.
 *
 * @callback Phaser.Types.Scenes.SceneInitCallback
 * @since 3.0.0
 *
+ * @this {Phaser.Scene}
 * @param {object} data - Any data passed via `ScenePlugin.add()` or `ScenePlugin.start()`. Same as Scene.settings.data.
 */
```

but it generates the wrong function type and omits `this` as a parameter. I'm sorry but I don't know where to go from here so I won't be able to submit a pull request, I will put the details in an issue if anybody is able to pick up from there

PR #4676, opened 2019-07-25
Fixed updateList property in phaser.d.ts

chrismarrin:

This PR 

* Fixes a bug

Describe the changes below:

Fixed the property name for updateList in two locations: GameOjbectFactory and GameObjectCreator


photonstorm:

PR #4685, opened 2019-08-01
fix `intensity` parameter type for typescript

jcyuan:
`intensity` can be `number` type or `Vector2` type
for `Shake.js` and `Camera.js`


photonstorm:
👍

Issue #4531, opened 2019-05-15
Particles scroll factor wrong when zooming camera on CANVAS mode


SirLink:
## Version

* Phaser Version:
Phaser v3.17.0 (Canvas | Web Audio)
* Operating system:
Windows 10
* Browser:
Chrome 
## Description

When you change the camera zoom  to any other value than 1 the particle emitter position starts to go out of place and moves faster than the camera. Probably related to the scroll factor.

## Example Test Code

https://codepen.io/anon/pen/gJmgVp

Click on the game to zoom, then change the type to WEBGL and click again. Notice how the particles end in a different spot.

Martin-Antonov:
I am also having an issue with this. A possibly simpler case can be seen from the examples:

**https://labs.phaser.io/edit.html?src=src\game%20objects\particle%20emitter\follow%20sprite.js**

where on line 30 add: `this.cameras.main.zoom = 0.5;` and change the renderer to CANVAS in the config. 
Any help on this will be much appreciated :)

Martin-Antonov:
I was able to fix it by adding "extra" to the x and y positions of the emitter. It seems that under Canvas the particle emitters do not take into consideration the zoom level. 
Here is my solution, maybe it will be helpful in fixing the bug:

`// The positions we want to place the particle emitter
let endX = 800;
let endY = 600;
// Take camera zoom and center
const zoom = this.scene.cameras.main.zoom;
const camCenterX = this.scene.cameras.main.worldView.centerX;
const camCenterY = this.scene.cameras.main.worldView.centerY;
// The extra x and y, which we need to add to endX and endY, so that the final position is indeed 800 and 600.
// We take the distance between endX and the center of the camera and multiply it by a transformation constant which depends on the camera 

const extraX = (endX - cameraCenterX) *((1/zoom)-1);  // another way of writing this is (1-zoom)/zoom
const extraY = (endY-cameraCenterY)*((1/zoom)-1);

endX+=extraX;
endY+=extraY;
`

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

photonstorm:
This issue has been mentioned on **Phaser**. There might be relevant details there:

https://phaser.discourse.group/t/zoom-tilemaps-parallax-global-vs-camera-zoom-broken/11786/1

Issue #4131, opened 2018-10-28
Particle emitter not moving with camera properly with CANVAS rendering and setDepth


jhooper04:
## Version

* Phaser Version:  v3.16.0 Beta 1 (Canvas | Web Audio)
* Operating system:  Windows 7 64bit
* Browser:  Using Chrome

## Description

Particle emitter position is having strange behavior with camera scrolling. If you create the particle manager after setting the camera bounds or set its depth, then the issue occurs. In the example if you move past the particle emitter, it is moving faster than the terrain and is not following the world space. This only happens when using the CANVAS rendering.

## Example Test Code

```js
var config = {
    type: Phaser.CANVAS,
    width: 800,
    height: 600,
    parent: 'phaser-example',
    physics: {
        default: 'impact',
        impact: {
            setBounds: {
                x: 0,
                y: 0,
                width: 3200,
                height: 600,
                thickness: 32
            }
        }
    },
    scene: {
        preload: preload,
        create: create,
        update: update,
        extend: {
            player: null,
            cursors: null,
            createLandscape: createLandscape
        }
    }
};

var game = new Phaser.Game(config);

function preload ()
{
    this.load.image('ship', 'assets/sprites/shmup-ship2.png');
    this.load.atlas('flares', 'assets/particles/flares.png', 'assets/particles/flares.json');
}

function create ()
{
    var particles = this.add.particles('flares').setDepth(1);

    //  The world is 3200 x 600 in size
    this.cameras.main.setBounds(0, 0, 3200, 600);

    this.createLandscape();

    //  Add a player ship

    this.player = this.impact.add.sprite(1600, 200, 'ship');
    this.player.setMaxVelocity(1000).setFriction(400, 200).setPassiveCollision();

    this.cursors = this.input.keyboard.createCursorKeys();

    particles.createEmitter({
        frame: 'blue',
        x: 1600,
        y: 300,
        lifespan: 2000,
        speed: { min: 400, max: 600 },
        angle: 330,
        gravityY: 300,
        scale: { start: 0.4, end: 0 },
        quantity: 2,
        blendMode: 'ADD'
    });
}

function update()
{
    if (this.cursors.left.isDown)
    {
        this.player.setAccelerationX(-800);
        this.player.flipX = true;
    }
    else if (this.cursors.right.isDown)
    {
        this.player.setAccelerationX(800);
        this.player.flipX = false;
    }
    else
    {
        this.player.setAccelerationX(0);
    }

[...]

PR #4687, opened 2019-08-02
Fixxing ParticleEmitterManager rotation rendering bug

Formic:

* Fixes a bug

Describe the changes below:
There was an issue where particles would orbit when the ParticleEmitterManager rotated and the camera was not at the origin.

I didn't create an issue for this, I wasn't sure if it was captured by #4531 or #4131. I didn't look too closely at those, and cannot confirm if it fixed them as well.

I quickly opened up most of the particle examples and verified they were still behaving as expected.


dlawrence2060:
Why was this closed?

Formic:
@dlawrence2060 It wasn't a real fix, initially I thought it worked, but there was some other issue or it created an issue elsewhere. I do not recall at this moment. 

I opened a bug report, and detailed the complexity of the issue highlighting the root cause in the Additional information.
https://github.com/photonstorm/phaser/issues/4691

PR #4690, opened 2019-08-05
Datamanager feature: update(key, object, forceCreate)

UnFlimFlammable:

This PR:
Adds a new feature

Describe the changes below:
This adds an update method that will take an object and update the existing object in the datamanager with the fields in the object passed in. 

Example:

```
this.data.set('test', {x:10, y:20});
this.data.update('test', {x:20});

var newTest = this.data.get('test');
//newTest.x = 20, newTest.y = 20
```

### DIFF
--- a/src/data/DataManager.js
+++ b/src/data/DataManager.js
@@ -379,6 +379,35 @@ var DataManager = new Class({
     },
 
     /**
+     * @method Phaser.Data.DataManager#update
+     * @since 3.19.0
+     * @fires Phaser.Data.Events#CHANGE_DATA
+     * @param {string} key - the key of the object to be updated
+     * @param {*} object - object containing new fields for the key
+     * @param {boolean} forceCreate - if fields in object aren't present, create them
+     * 
+     * @return {Phaser.Data.DataManager} This DataManager object.
+     */
+    update: function(key, object, forceCreate)
+    {
+        var existing = this.get(key);
+
+        //If the object already exists
+        if(existing !== undefined){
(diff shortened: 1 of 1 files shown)


KingCosmic:
I for one would like this to be added 💃

samme:
I think this needs `eslint --fix`.

src/data/DataManager.js
```diff
@@ -378,6 +378,38 @@ var DataManager = new Class({
...
+     */
+    update: function(key, object, forceCreate)
+    {
+        var existing = this.get(key);
+        if(forceCreate === undefined){forceCreate = true}
+        //If the object already exists
+        if(existing !== undefined){
+            for(var key in object){
+                //If the field exists inside the object, update, otherwise if forceCreate is true add field anyway
+                if(existing[key] !== null || forceCreate === true){
```
samme:
```suggestion
                if ((key in existing) || forceCreate === true) {
```

photonstorm:
eslint issues aside, you can already do this:

```
this.data.set('test', {x:10, y:20});

this.data.values.test.x = 20;

var newTest = this.data.get('test');
//newTest.x = 20, newTest.y = 20
```

jcyuan:
datamanager has `frozen` feature, you can't do this by a `update` method.

photonstorm:
Thanks for this PR, however, I'm afraid I'm going to close it off, as it doesn't pass the eslint check and you can already do this via the `values` property of the Data Manager.

PR #4715, opened 2019-08-19
Container now can apply alpha quads

MrcSnm:

* Updates the Documentation
* Adds a new feature
* Fixes a bug

Describe the changes below:
Now we can apply the alpha quad to the container and to its children too


MrcSnm:
Made the fixes from the eslint config

photonstorm:
Much better, thanks! There's still one tiny thing to change, but I'll do it locally so we can get this into master.

Issue #4672, opened 2019-07-24
Arcade physics: friction force scales incorrectly with physics FPS


BenjaminDRichards:
## Version

* Phaser Version: 3.17, 3.18 tested

* Operating system: MacOS Sierra (10.12.6)

* Browser: tested in Chrome 76

## Description

- Update Arcade Physics FPS to 120.
- Create two game objects with physics enabled, in the style of "character on moving platform".
- Object A has gravity downward (the "player").
- Object B is immovable, and moves side to side (the "platform").
- Expected behaviour: object A is carried along by object B at exactly the same speed.
- Actual behaviour: object A moves 50% faster than object B, and slides to the leading edge of B. (A doesn't ever fall, because once it moves off B, it is no longer being moved, and B moves back under it.)

## Example Test Code

Quickest way to reproduce this is to fire up the friction demo at:

http://labs.phaser.io/edit.html?src=src/physics/arcade/friction.js&v=3.18.1

Then change the physics config to read like this, and run:

`physics: {default: 'arcade', arcade: {fps: 120}},`

This will set FPS to double the default of 60. You should see the top platform, which is supposed to have friction 1, actually pushing the lemmings ahead of it.

## Additional Information

I've traced the problem to the state of `prev` in `Phaser.Physics.Arcade.Body`. This is read by `SeparateX`/`Y` when assessing friction. We see code something like `body1.x += (body2.x - body2.prev.x) * body2.friction.x;`, which is fine if `prev` is precisely one physics step ago. But when `prev` is several steps old, every step increases the distance between now and then, and the friction is applied for every step. This results in friction pushing too hard and moving the object too far.

I have solved the problem in my project by setting friction.x for all objects to `2 / 3`. However, this is clumsy, and incompatible with cases such as:
- Changing the physics step FPS to another value
- Setting custom friction on some platforms

It also causes some stuttering in the movement of the player on the platform, because the friction is no longer a 1:1 relationship.

Because `prev` has existing use cases (e.g. users being able to do their own analysis of frame-by-frame movement), I think it should stay as it is. Instead, I would add a new property, perhaps `prevStep`, which is updated with every physics step. This would allow friction to be calculated more accurately, and also allow users to analyse some kinds of step-by-step movement.

[...]

PR #4729, opened 2019-08-28
Arcade physics fps fixes

BenjaminDRichards:

* Fixes a bug

Describe the changes below:

We've encountered several issues in Phaser when using Arcade Physics and the framerate drops below 60. (Technically, any time when multiple physics steps run per frame, so if physics FPS is above 60 this will also occur.)

Issue 1: Friction starts to flip out. Objects on moving platforms get pushed ahead of the platform and "catch" on the leading edge.

Issue 2: Physics objects start to dip into the floor. In the "Before" demo, the camera is locked to the player, so this appears as the entire world starting to shake up and down.

Issue 3: When objects dip into the floor, their "rest velocity" is non-zero. This can affect debug and other logic.

Early sightings of these issues were reported in issue #4672 

I provide a demo of these issues at work, using a barrel of balloons to drag the frame rate down to the point where these effects becomes noticeable. Debug commands are available for tweaking the frame rate killers for different machines.

Demo 1: Before
https://storage.googleapis.com/staging-files.gamefroot.com/users/71078/games/106343/gamefroot-2019-7-27T3-43-18.540Z/index.html

I've developed fixes for these issues, hopefully without significant impact.

Demo 2: After
https://storage.googleapis.com/staging-files.gamefroot.com/users/71078/games/106343/gamefroot-2019-7-27T3-43-18.540Z-new/index.html

The friction fix in more detail:

* Added `prevFrame` vector to Arcade.Body, so we can distinguish between frame-length changes and step-length changes. Several steps may run for every frame, particularly when fps is low.
* Removed `_reset` flag from Arcade.Body, and replaced it with a check of `this.moves`. The flag only turned on when `this.moves` was true, and never turned off.
* Added reset of `prev` in Arcade.Body#step. This fixes the **friction** issue.
* Removed Arcade.Body#postUpdate setting `_dx`, `_dy`, and `prev`. They remain in the state they were at the end of the last physics step. This will affect the delta methods, which are documented to provide step-based data (not frame-based data); they now do so. (However, because several steps may run per frame, you can't interrogate every step unless you're running functions based on physics events like collisions. You'll just see the latest step.) This should partially balance out the extra load of resetting `prev`, above.
* Added a zero-out of `stepsLastFrame` in Arcade.World#postUpdate, which would otherwise never zero out and keep running at least one pass per frame. This should improve performance when frames can be skipped.

And the floor dip fix:

* Removed `blocked` checks from TileCheckX and TileCheckY. Originally, this prevented multiple checks when an object had come to rest on a floor. However, when multiple steps run per frame, the object will accelerate again, the floor won't stop it on steps 2+, and it will end the frame a short distance into the floor. Removing the `blocked` checks will fix the **floor dip** issue and the **rest velocity** issue. Although this opens up multiple checks, this is probably very rare: how many times does an object hit two different floors in a single frame?
  * An alternative fix might be to simply prevent an object from moving in a given direction if it's been blocked already during this frame's steps. But this would cause weird effects when sliding around corners (you'd actually want to have the option of moving in the blocked direction once you pass the corner), so I don't recommend it.

I think this will result in more reliable physics under high step rate conditions (e.g. low fps, or running high physics fps to cope with high velocity game objects), without any significant performance cost. Both demo scenes run around 17fps in Chrome on my desktop.

Note: Instead of `@since`, I documented the `prevFrame` vector with `@author`, because I wasn't sure if/when this fix would be released, and I figured I might as well toot my own horn. Please alter that section as necessary.

### DIFF
--- a/src/physics/arcade/Body.js
+++ b/src/physics/arcade/Body.js
@@ -160,6 +160,15 @@ var Body = new Class({
         this.prev = new Vector2(gameObject.x, gameObject.y);
 
         /**
+         * The position of this Body during the previous frame.
+         * 
+         * @name Phaser.Physics.Arcade.Body#prevFrame
+         * @type {Phaser.Math.Vector2}
+         * @author Benjamin D. Richards <benjamindrichards@gmail.com>
+         */
+        this.prevFrame = new Vector2(gameObject.x, gameObject.y);
+
+        /**
          * Whether this Body's `rotation` is affected by its angular acceleration and angular velocity.
          *
          * @name Phaser.Physics.Arcade.Body#allowRotation
@@ -697,17 +706,6 @@ var Body = new Class({
         this.physicsType = CONST.DYNAMIC_BODY;
 
         /**
-         * Whether the Body's position needs updating from its Game Object.
-         *
-         * @name Phaser.Physics.Arcade.Body#_reset
-         * @type {boolean}
-         * @private
-         * @default true
-         * @since 3.0.0
-         */
-        this._reset = true;
-
-        /**
          * Cached horizontal scale of the Body's Game Object.
          *
          * @name Phaser.Physics.Arcade.Body#_sx
@@ -909,10 +907,12 @@ var Body = new Class({
 
         this.preRotation = this.rotation;
 
-        if (this._reset)
+        if (this.moves)
         {
             this.prev.x = this.position.x;
             this.prev.y = this.position.y;
+            this.prevFrame.x = this.position.x;
+            this.prevFrame.y = this.position.y;
         }
 
         if (willStep)
@@ -936,6 +936,9 @@ var Body = new Class({
      */
     update: function (delta)
     {
+        this.prev.x = this.position.x;
+        this.prev.y = this.position.y;
+
         if (this.moves)
         {
             this.world.updateMotion(this, delta);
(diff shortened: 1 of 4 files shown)


falifm:
Hi @BenjaminDRichards, I downloaded and built your branch and I just wanna let you know that you've just solved this issue: https://github.com/photonstorm/phaser/issues/4732

Thank you so much! This fix needs to be added to next version of Phaser.

falifm:
@BenjaminDRichards, I just want to inform you that your implementation somehow broke arcade body reset() method. If I call _this.body.reset(startPos, startPos)_ on arcade sprite in order to move it to initial position (flow of my sprite is StartPos -> Move to endPos -> Reset to StartPos), the body is positioned on wrong coordinates. Also I have experienced different behaviour when I call _this.body.enable = false_ before reset. 

I am sorry I am not attaching any code, but I am too busy to separate it from my big project right now. The only thing I know is that with official Phaser distribution reset behaves correctly. If I switch to your build, the error is present. Can you take a look on this issue please? Thanks so much!

BenjaminDRichards:
@falifm Good spotting! Let me dig into that and see if I can improve matters before this gets merged.

photonstorm:
@BenjaminDRichards did you get anywhere with the issue @falifm mentioned? just wondering as I'll hold on that before merging into 3.20.

BenjaminDRichards:
@photonstorm I'm currently working on an unrelated, urgent project, so no progress on this. I'll make noise when I've had the time to diagnose and resolve the issue, I'm guessing in about a week or so.

the-simian:
@BenjaminDRichards really looking forward to this one, thank you man, I can't wait!

photonstorm:
This is all now merged into 3.20 and I've added in a fix which should address the `Body.reset` issue as well. Thanks everyone.

PR #4744, opened 2019-09-10
RenderTexture would not update texture size on resize

yhwh:
This PR Fixes a bug

Describe the changes below:
If the texture size is not updated on resize() then the subsequent calls to draw() or drawFrames() will produce a distorted output.


photonstorm:
👍

PR #4752, opened 2019-09-13
FIX #4751 - Physics.Arcade.World.separateCircle - weird effects when …

### LINKED ISSUE #4751: Physics.Arcade.World.separateCircle - weird effects when bounce=0
## Version
* Phaser Version: 3.19
* Operating system: OSX
* Browser: Chrome

## Description

When you have 2 circle bodies collide and they do not have bounce set, then the separation logic will always use an angle of 180 deg for calculating the separating vector.  This can lead to weird effects depending on the actual collision vector.

## Example Test Code

https://codepen.io/funnisimo/pen/LYPrpBX

funnisimo:

Fixes a bug

Describe the changes below:

FIX #4751 - Physics.Arcade.World.separateCircle - weird effects when bounce=0

* Fixes a bug

See the example here:

https://codepen.io/funnisimo/pen/LYPrpBX

Describe the changes below:

Moved the scaling of velocity for bounce to after angle is calculated and separation is done.

Also...
- If both bodies are movable, then they only need to be offset by half of the offset.  Using the full offset for both bodies separates them too much.
- Changed to use the centers of the bodies to calculate the angle, that way any offsets will be included in the calculation.

### DIFF
--- a/src/physics/arcade/World.js
+++ b/src/physics/arcade/World.js
@@ -323,7 +323,7 @@ var World = new Class({
          *
          * If you have a large number of dynamic bodies in your world then it may be best to
          * disable the use of the RTree by setting this property to `false` in the physics config.
-         * 
+         *
          * The number it can cope with depends on browser and device, but a conservative estimate
          * of around 5,000 bodies should be considered the max before disabling it.
          *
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

PR #4754, opened 2019-09-13
Destroying matterjs fails as world is already null

nagyv:
See https://jsfiddle.net/gm750vo4/1/

This PR

* Fixes a bug

Describe the changes below:
added checking in MatterPhysics::shutdown


photonstorm:
Thanks for your PR and fixing this bug. I'm afraid I cannot merge it, as your PR doesn't pass the ESLint check, but I have applied your fix locally and credited you in the Change Log.

PR #4760, opened 2019-09-21
Set dirty flag only when render state of bob is changed

rexrainbow:

    Enhance a little performance

Describe the changes below:

Only set dirty flag when bob becomes invisible (visible === false, or alpha <= 0) from visible state, or from visible state to invisible state.
Use case: when bob.alpha is changed between 1 to 0.0000x, it won't set dirty flag.


photonstorm:
👍

Issue #4510, opened 2019-05-06
SceneManager stop() not accepting and passing through optional data


GetsukenStudios:
## Version

* Phaser Version: 3.16.2

* Operating system: 

* Browser: 

## Description

When I needed to pass extra data to my `shutdown` event handler in a scene, I realized after some debugging that the method `Phaser.Scenes.SceneManager#stop` is missing the parameter for the optional `data` to pass when stopping a scene and it also doesn't pass `data` to `Phaser.Scenes.Systems#shutdown`.

## Example Test Code

```    
    stop: function (key)
    {
        var scene = this.getScene(key);

        if (scene && !scene.sys.isTransitioning())
        {
            scene.sys.shutdown();
        }

        return this;
    }
````

## Additional Information

My simple fix:
```
    stop: function (key, data)
    {
        var scene = this.getScene(key);

        if (scene && !scene.sys.isTransitioning())
        {
            scene.sys.shutdown(data);
        }

        return this;
    }
```

**Comment by photonstorm** [comment]
It's not missing it, it's simply not meant to be there and isn't part of the API.

**Comment by GetsukenStudios** [comment]
I see, thanks. 

I'm curious, why does it pass `data` in `Phaser.Scenes.Systems#shutdown` when emitting the `SHUTDOWN` event then? Because that's why I assumed that `Phaser.Scenes.SceneManager#stop` is supposed to pass the data when it calls `scene.sys.shutdown()`. 
```
shutdown: function (data)
    {
        this.events.off(Events.TRANSITION_INIT);
        this.events.off(Events.TRANSITION_START);
        this.events.off(Events.TRANSITION_COMPLETE);
        this.events.off(Events.TRANSITION_OUT);

        this.settings.status = CONST.SHUTDOWN;

        this.settings.active = false;
        this.settings.visible = false;

        this.events.emit(Events.SHUTDOWN, this, data);
    }
```

Olliebrown:
I'm examining the API just to make sure this simple solution makes sense in the larger context of the Scene subsystem.  Looking at the current lifecycle of a scene, an optional `data` already pops up in some places:

### Scene.Systems.settings.data
- This comes from an object passed to SceneManager's `add()`, `start()`, or `run()` functions
- It is passed into `Scene.init()` by SceneManager's `bootScene()` method (or `bootQueue()`)
- It is passed into `Scene.create()` by the SceneManager
- It is also just generally accessible within the scene as `this.settings.data`

[...]

PR #4783, opened 2019-10-03
Feature/Fix for issue #4510

Olliebrown:
This PR Adds a new feature

### Description of changes:
As described in issue #4510 there was no parameter accepted in the SceneManager.stop() function so that data could be passed through to the shutdown event for the scene. This PR adds that parameter to stop and documents it in the JSDoc.  It is directly passed to the shutdown event and behaves as transient data just for the event not adjusting or falling back to sys.settings.data in any way.


photonstorm:
Thanks for this. It also needs adding to the ScenePlugin, as you should rarely use the SceneManager directly, but I'll do that here.

PR #4785, opened 2019-10-04
Set 'part.angle' in 'Body.update' into matter-js

Frozzy6:

* Fixes a bug

Describe the changes below:
According to [this matterjs pr](https://github.com/liabru/matter-js/pull/298) angle of body as part another one doesn't correctly changes.


photonstorm:
I don't really like modifying Matter directly, but we don't really have much choice!

PR #4787, opened 2019-10-06
Fix incorrect lighting when batching several sprites

gogoprog:
This PR
* Fixes a bug

Hello!

The lighting is not correct when batching several sprites with different rotations:
Each sprite should use its own uInverseRotationMatrix to compute the lighting correctly.
Because of the batching the uniform has the rotation of the last rendered sprite.

This is wrong so this PR add a check and flush the batching if needed.

Let me know if you have any question or if I need to change something!

Recorded issue here, the cursor is the light position:

Peek 2019-10-06 20-48


photonstorm:
👍

PR #4789, opened 2019-10-07
Updated scaleManager.js Docs

arnekeller:
Default mode was descibed as `NO_SCALE` in the docs but is actually `NONE`

This PR

* Updates the Documentation


photonstorm:
👍

PR #4798, opened 2019-10-13
Move `remove-files-webpack-plugin` to devDependencies

noseglid:

* Fixes a bug

Describe the changes below:

Surely, this dependency should be a devDependency.
I'm building my phaser app with [parceljs](https://parceljs.org/) and do not have `webpack` installed at all. As a consequence I'm running into this warning
```
warning "phaser > remove-files-webpack-plugin@1.1.3" has unmet peer dependency "webpack@*".
```

Everything works fine, because obviously I'm not running the build process when I depend on phaser.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "fs-extra": "^6.0.1",
     "jsdoc": "^3.6.1",
     "node-sloc": "^0.1.11",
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

PR #4807, opened 2019-10-16
fix TS issues and fix the condition to check video support.

jcyuan:
This PR

⚪️  Updates the Documentation
⚪️  Fixes a bug

Describe the changes below:

1, check video support when TextureSource initializing.
2, update jsdoc and ts output.

### DIFF
--- a/src/gameobjects/UpdateList.js
+++ b/src/gameobjects/UpdateList.js
@@ -18,6 +18,7 @@ var SceneEvents = require('../scene/events');
  * Some or all of these Game Objects may also be part of the Scene's [Display List]{@link Phaser.GameObjects.DisplayList}, for Rendering.
  *
  * @class UpdateList
+ * @extends Phaser.Structs.ProcessQueue.<Phaser.GameObjects.GameObject>
  * @memberof Phaser.GameObjects
  * @constructor
  * @since 3.0.0
--- a/src/textures/TextureSource.js
+++ b/src/textures/TextureSource.js
@@ -135,7 +135,7 @@ var TextureSource = new Class({
          * @type {boolean}
          * @since 3.0.0
          */
-        this.isCanvas = (source instanceof HTMLCanvasElement);
+        this.isCanvas = (window.hasOwnProperty('HTMLCanvasElement') && source instanceof HTMLCanvasElement);
 
         /**
          * Is the source image a Video Element?
@@ -144,7 +144,7 @@ var TextureSource = new Class({
          * @type {boolean}
          * @since 3.20.0
          */
-        this.isVideo = (source instanceof HTMLVideoElement);
+        this.isVideo = (window.hasOwnProperty('HTMLVideoElement') && source instanceof HTMLVideoElement);
 
         /**
          * Is the source image a Render Texture?
--- a/src/textures/TextureSource.js
+++ b/src/textures/TextureSource.js
@@ -135,7 +135,7 @@ var TextureSource = new Class({
          * @type {boolean}
          * @since 3.0.0
          */
-        this.isCanvas = (window.hasOwnProperty('HTMLCanvasElement') && source instanceof HTMLCanvasElement);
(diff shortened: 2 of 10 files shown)


src/textures/TextureSource.js
```diff
@@ -135,7 +135,7 @@ var TextureSource = new Class({
          * @type {boolean}
          * @since 3.0.0
          */
-        this.isCanvas = (source instanceof HTMLCanvasElement);
+        this.isCanvas = source instanceof HTMLCanvasElement;
```
photonstorm: Please undo the bracket removal.

src/textures/TextureSource.js
```diff
@@ -144,7 +144,7 @@ var TextureSource = new Class({
          * @type {boolean}
          * @since 3.20.0
          */
-        this.isVideo = (source instanceof HTMLVideoElement);
+        this.isVideo = (window.hasOwnProperty('HTMLVideoElement') && source instanceof HTMLVideoElement);
```
photonstorm: I wonder if this should be a `customEnvironment` config check instead. Although, checking `window` may be less code perhaps.

types/phaser.d.ts
```diff
@@ -5601,45 +5601,51 @@ declare namespace Phaser {
             getPoints(divisions?: integer): Phaser.Math.Vector2[];
 
             /**
-             * [description]
+             * Returns a randomly chosen point anywhere on the path. This follows the same rules as `getPoint` in that it may return a point on any Curve inside this path.
```
photonstorm: You cannot edit the defs file directly, as all of your changes will be automatically overwritten the next time the defs file is generated. Please edit the jsdocs instead, then those changes will filter down to the defs.

src/gameobjects/UpdateList.js
```diff
@@ -18,6 +18,7 @@ var SceneEvents = require('../scene/events');
  * Some or all of these Game Objects may also be part of the Scene's [Display List]{@link Phaser.GameObjects.DisplayList}, for Rendering.
  *
  * @class UpdateList
+ * @extends Phaser.Structs.ProcessQueue.<Phaser.GameObjects.GameObject>
```
photonstorm: Please try generating the defs after doing this, to make sure it doesn't break them. You can use the command `npm run tsdev` to check and view the contents of the output.txt it builds.

src/textures/TextureSource.js
jcyuan: do~

src/textures/TextureSource.js
jcyuan: yes, please consider about a solution, anyway for now this works in our env, other codes related to video gameobject are not compiled into the final output, only this line throws error.

src/gameobjects/UpdateList.js
jcyuan: yes I did this and tested, so this is the reason the commit has `phaser.d.ts` changes in it. remove the `phaser.d.ts` from the commit is ok, `cause users can re-gnerate it by npm run command still.

types/phaser.d.ts
jcyuan: I regenerated this by `tsdev` command

jcyuan:
hi @photonstorm, I have changed some files since the last commit again, please review the changes, thank you.

types/phaser.d.ts
photonstorm: That makes sense - please remove the defs file from your PR so I can merge it :)

photonstorm:
Happy to merge this now, but please remove the TS defs file from the PR first, thanks.

jcyuan:
done, but actually you can't remove things from a commit once it's pushed (or I don't know how to do), so I have restored the `phaser.d.ts` with the masters and it generated a new commit.
if you're not ok with this, I can close this PR and create another with clearer content.

photonstorm:
That worked, thanks 👍

Issue #4099, opened 2018-10-06
[Tiled] add support for layer groups


Babeetlebum:
# Problem

I've been trying to organize my layers into groups using Tiled 1.2.0

<img width="173" alt="tiled_groups" src="https://user-images.githubusercontent.com/5504984/46574825-3bea5080-c9dc-11e8-8ee1-cf6414b9e2a7.png">

However these groups are being ignored by the Tiled parser

# What's happening

The parser looks for layers of type "tilelayer" and ignores the rest. My "Platform" element is of type "group". As a result, `this.layers` is populated only with the layer "Ground" in this example

# Details

The JSON structure of the example above :
```json
{ "height":30,
 "infinite":false,
 "layers":[
        {
         "data":[...],
         "height":30,
         "id":1,
         "name":"Ground",
         "opacity":1,
         "type":"tilelayer",
         "visible":true,
         "width":40,
         "x":0,
         "y":0
        }, 
        {
         "id":2,
         "layers":[
                {
                 "data":[...],
                 "height":30,
                 "id":5,
                 "name":"Platform",
                 "opacity":1,
                 "type":"tilelayer",
                 "visible":true,
                 "width":40,
                 "x":0,
                 "y":0
                }, 
                {
                 "data":[...],
                 "height":30,
                 "id":6,
                 "name":"Platform-grass",
                 "opacity":1,
                 "type":"tilelayer",
                 "visible":true,
                 "width":40,
                 "x":0,
                 "y":0
                }],
         "name":"Platform",
         "opacity":1,
         "type":"group",
         "visible":true,
         "x":0,
         "y":0
        }],
 "nextlayerid":7,
 "nextobjectid":1,
 "orientation":"orthogonal",
 "renderorder":"right-down",
 "tiledversion":"1.2.0",
 "tileheight":16,
 "tilesets":[],
 "tilewidth":16,
 "type":"map",
 "version":1.2,
 "width":40
}
```

The relevant line : l30 in /src/tilemaps/parsers/tiled/ParseTileLayers.js :
```javascript
        if (json.layers[i].type !== 'tilelayer')
        {
            continue;
        }
```

# Feature proposal

I think that the group properties have to be passed on to the nested layers in the following way :
- `child_opacity *= parent_opacity`
- `child_visible = child_visible && parent_visible`
- `child_x += parent_x`
- `child_y += parent_y`

Did I forget any property ?

The new code should probably use a recursive loop to go through the different levels of nesting (groups inside groups)

[...]

PR #4820, opened 2019-10-23
Support for Tiled layer groups

Olliebrown:
This PR
* Adds a new feature
* Fixes a bug

Add support for group layers when parsing JSON output from Tiled as requested in issue #4099.  Along the way, I believe I discovered some problems with the way phaser parses infinite Tiled maps and fixed those as well.  Infinite map layers might still have 'offsetx' and 'offsety' properties but these were previously being ignored.  They are now accounted for properly.

I have tested all these changes against the phaser3-examples both visually and with the 'tester' script and everything that worked before the changes works the same as after my changes.  I added several tests to the examples repo in my local working copy that tested groups and an infinite map to ensure those parts work as well.  Those examples are NOT a part of this PR as they were on top of the examples repo and not this one.

Here is a quick summary of changes:
- Added logic to each of the layer type parsers (Tiled, Object, and Image) to account for group layers.
- Added a File with a single function that creates the important state that needs to be inherited from a group.
- Uses an iterative approach to traversing the group hierarchy with a stack rather than recursion.
- All hierarchy is flattened to a single array of layers and the group layers are discarded after they are accounted for.
- Layer names are prepended with any containing group names separated by slashes ('/')
- Updated documentation for the Tilemap GameObject to mention support for layers and the renaming that can occur.
- Updated error reporting when an invalid tile layer name is requested in order to pre-empt any confusion surround the renaming of layers.  Now lists the valid layer names.

### DIFF
--- a/src/tilemaps/Tilemap.js
+++ b/src/tilemaps/Tilemap.js
@@ -55,6 +55,19 @@ var Tileset = require('./Tileset');
  * StaticTilemapLayer or DynamicTilemapLayer may have its own unique tile size that overrides
  * it.
  *
+ * As of Phaser 3.21.0, if your tilemap includes layer groups (a feature of Tiled 1.2.0+) these
+ * will be traversed and the following properties will affect children:
+ * - opacity (blended with parent) and visibility (parent overrides child)
+ * - Vertical and horizontal offset
+ * The grouping hierarchy is not preserved and all layers will be flattened into a single array.
+ * Group layers are parsed during Tilemap construction but are discarded after parsing so dynamic
+ * layers will NOT continue to be affected by a parent.
+ *
+ * To avoid duplicate layer names, a layer that is a child of a group layer will have its parent
+ * group name prepended with a '/'.  For example, consider a group called 'ParentGroup' with a
+ * child called 'Layer 1'. In the Tilemap object, 'Layer 1' will have the name
+ * 'ParentGroup/Layer 1'.
+ *
  * @class Tilemap
  * @memberof Phaser.Tilemaps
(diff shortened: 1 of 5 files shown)


photonstorm:
Thanks for this great PR. I've merged it in, although it does break a few code style rules, but not enough for it to not be merged (I'll just fix them locally)

floriancargoet:
@Olliebrown @photonstorm 
This PR is great for those who work with Tiled groups but I noticed the custom properties of a group are lost, only the default properties (name, opacity, visible, x and y) are "copied".

I think we could put the custom properties in the group state object and merge them with the parent group state properties. That way, a layer would have its own properties and the inherited properties of its ancestors.
Would you consider a PR doing that?

Properties are in arrays so technically we could concatenate  them and have multiple properties with the same name. Since Tiled doesn't allow multiple properties of the same name, I suggest we do the same and the properties of a child layer override the properties of its ancestors.

Olliebrown:
I’d love to see that! I always try and champion the custom properties to my students as a critical part of The flexibility of Tiled for world design.  We’re you offering to make that PR or is it something I should look into?  I’m not sure where Rich is with changes to Phaser 3 these days but if it’s a clean PR and well tested I imagine he’d take it.

floriancargoet:
I am offering the write the PR :-)

floriancargoet:
@Olliebrown @photonstorm 
I'm working on the PR and I have questions (I found some inconsistencies in property handling and a related bug) and I'd like to know where the best place ask them. Here? In a work-in-progress PR? Some other channel?

Issue #4167, opened 2018-11-15
Light2D not properly working for DynamicTilemapLayer from a Tileset


blackjack26:
## Version
* Phaser Version: 3.15.1
* Operating system: Windows 10 Pro 1709 Build 16299.726
* Browser: Chrome v70.0.3538.102 (64-bit)

## Description

When using a normal map with a tileset image, and create a `DynamicTilemapLayer` with that tileset, I should be able to set the layers pipeline as `Light2D`. What occurs is that the console logs the error:

> Uncaught TypeError: Cannot read property 'dataSource' of undefined
    at ForwardDiffuseLightPipeline.batchTexture (VM75183 ForwardDiffuseLightPipeline.js:237)
    at DynamicTilemapLayerWebGLRenderer [as renderWebGL] (VM75270 DynamicTilemapLayerWebGLRenderer.js:89)
    at WebGLRenderer.render (VM75179 WebGLRenderer.js:1752)
    at CameraManager.render (VM74411 CameraManager.js:612)
    at Systems.render (VM74510 Systems.js:375)
    at SceneManager.render (VM74507 SceneManager.js:600)
    at Game.step (VM74482 Game.js:554)
    at TimeStep.step (VM74542 TimeStep.js:560)
    at step (VM74479 RequestAnimationFrame.js:104)

This prevents the scene from loading and rendering anything.

## Example Test Code

This is a subset of the code I am using to generate the tilemap and add the lights.

```es6
preload() {
  this.load.image({
    key: 'my_tiles',
    url: 'assets/my-tileset.png',
    normalMap: 'assets/my-tileset_n.png'
  });
}

create() {
  this.lights.enable().setAmbientColor( 0x555555 );
  this.lights.addLight( 100, 100, 200 );

  const map = this.make.tilemap( { /* ... */ } );
  const tileset = map.addTilesetImage( 'my_tiles', null, 32, 32, 1, 2 );
  this.layer = map.createBlankDynamicLayer( 'Ground', tileset );
  this.layer.setPipeline( 'Light2D' );
}
```

## Additional Information

The problem is in `ForwardDiffuseLightPipeline.js` on [line 237](https://github.com/photonstorm/phaser/blob/master/src/renderer/webgl/pipelines/ForwardDiffuseLightPipeline.js).

[...]

Issue #4079, opened 2018-10-02
Light2D for Dynamic Tilemap Layer Not Working in 3.14.0


kainage:
## Version
* Phaser Version: 3.14.0

## Description

Phaser fails to load when trying to use the Light2D pipeline with dynamic tilemaps with the following error:

```javascript
ForwardDiffuseLightPipeline.js:235 Uncaught TypeError: Cannot read property 'dataSource' of undefined
    at ForwardDiffuseLightPipeline.batchTexture (VM218 dev.js:136518)
    at DynamicTilemapLayerWebGLRenderer [as renderWebGL] (VM218 dev.js:162463)
    at WebGLRenderer.render (VM218 dev.js:135491)
    at CameraManager.render (VM218 dev.js:11608)
    at Systems.render (VM218 dev.js:142452)
    at SceneManager.render (VM218 dev.js:139972)
    at Game.step (VM218 dev.js:6720)
    at TimeStep.step (VM218 dev.js:7535)
    at step (VM218 dev.js:25359)
```

## Example Test Code

Can be observed on the phaser labs example:

http://labs.phaser.io/view.html?src=src\game%20objects\lights\dynamic%20tilemap%20layer.js&v=3.14.0

## Additional Information

Worked fine on 3.13.0

http://labs.phaser.io/view.html?src=src\game%20objects\lights\dynamic%20tilemap%20layer.js&v=3.13.0

samvieten:
Can confirm, got the same Bug, this happens only on dynamicTilelayers.

samvieten:
addendum: It seems like _normalmaps_ do not work anymore too. Can anyone confirm?

photonstorm:
Normal maps and lights are working fine with the exception being dynamic layers.

samvieten:
You are right, it's only dynamic layers.

photonstorm:
I really wish people would stop doing this. If an issue isn't listed in the Change Log, or closed, then it hasn't been fixed yet, no matter how many releases have come out in the meantime. The change log is meticulous and details everything, it wouldn't exclude something like this.

imilo:
Hello!

@photonstorm , can you please raise a priority of this issue? More than half a year lighting a tilemap is not working.

Thank you!

daumann:
Is there any workaround for this by the community? I would go back to 3.13.0 but there the collision is not working...
[...]

PR #4827, opened 2019-10-25
fix light2d using tileset from dynamictilemaplayer (#4167,#4079)

koljakutschera:
This PR

* Fixes 2 bugs (#4167, #4079)

Dynamictilemaplayer is broken when used with light2d. The reason is that since Phaser v3.14 the tileset property is an array and not a single reference. Currently the code is referencing:

`normalTexture = gameObject.tileset.image.dataSource[0];`

the result is an Error breaking the whole application. To fix this we must replace it with:

`normalTexture = gameObject.tileset[0].image.dataSource[0];`

For safety reasons i added a check if gameObject.tileset is really an array with a fallback if not.

Yes, i ran the linter ;)


photonstorm:
👍

PR #4828, opened 2019-10-28
Fixes for drawing offset hitboxes in input.enableDebug (fixes #4722)

### LINKED ISSUE #4722: enableDebug for input on gameobjects does not display correctly if offset
## Version

* Phaser Version: 3.19

* Browser: Chrome

## Description

When using enableDebug to draw the interactive area, if you give a hitArea to the config in setInteractive, the enableDebug will not draw the correct boundaries for the input area. It doesn't take into account the x and y values of the rectangle given. The actual boundaries do work since in the example, when highlighted over the bottom right of the rectangle, it will still show the hand cursor changing. 

## Example Test Code

```
var config = {
	type:   Phaser.WEBGL,
	parent: 'phaser-game',
	scene:  {
		preload: preload,
		create:  create
	}
};

var game = new Phaser.Game( config );

function preload() {
	this.load.image( 'eye', 'assets/pics/lance-overdose-loader-eye.png' );
}

function create() {
	var sprite = this.add.rectangle( 100, 100, 100, 100, 0xff0000 )
		.setInteractive( {
			hitArea:         new Phaser.Geom.Rectangle( 50, 50, 50, 50 ),
			hitAreaCallback: Phaser.Geom.Rectangle.Contains,
			useHandCursor:   true
		} );
	this.input.enableDebug( sprite );
}
```

Olliebrown:
This PR
* Fixes a bug

Describe the changes below:
- Added an offset variable that is accounted for in the pre-update method of the debug shape
- Offset is computed differently for each type of hitbox to ensure it matches actual hitbox
- Tested with rectangles, triangles, circles, ellipses, lines, and polygons and all work

### DIFF
--- a/src/input/InputPlugin.js
+++ b/src/input/InputPlugin.js
@@ -2403,14 +2403,20 @@ var InputPlugin = new Class({
             debug = null;
         }
 
+        var offsetx = 0;
+        var offsety = 0;
         switch (shapeType)
         {
             case GEOM_CONST.CIRCLE:
                 debug = factory.arc(0, 0, shape.radius);
+                offsetx = shape.x - shape.radius;
+                offsety = shape.y - shape.radius;
                 break;
 
             case GEOM_CONST.ELLIPSE:
                 debug = factory.ellipse(0, 0, shape.width, shape.height);
+                offsetx = shape.x - shape.width / 2;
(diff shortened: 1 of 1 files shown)


photonstorm:
Lovely, thanks!

PR #4833, opened 2019-10-30
Fix IE compatibility with use of Number.MIN_SAFE_INTEGER & Number.MAX_SAFE_INTEGER

jronn:
This PR fixes a bug

Bug: 
tweens onStart callback is not invoked in IE (Events.TWEEN_START never dispatched)

Cause:
Number.MIN_SAFE_INTEGER & Number.MAX_SAFE_INTEGER are not supported in IE

See:
https://docs.w3cub.com/javascript/global_objects/number/min_safe_integer/
https://docs.w3cub.com/javascript/global_objects/number/max_safe_integer/

You could probably just set these constants once somewhere if they're not already set, instead of fixing every use of them, but I'm not familiar enough with the code base to know where I would do so

### DIFF
--- a/src/curves/path/Path.js
+++ b/src/curves/path/Path.js
@@ -404,8 +404,8 @@ var Path = new Class({
         out.y = Number.MAX_VALUE;
 
         var bounds = new Rectangle();
-        var maxRight = Number.MIN_SAFE_INTEGER;
-        var maxBottom = Number.MIN_SAFE_INTEGER;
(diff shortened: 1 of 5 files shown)


photonstorm:
Thank you for this! You're right, it makes more sense to use a cached value, so I've added one into the math consts file and used that everywhere instead. I've still credited you with the fix, of course, even though I can't merge this PR now :)

Issue #4737, opened 2019-09-05
Tilemap.convertLayerToStatic throws error when used multiple times


Vegita2:
## Version

* Phaser Version: v3.17.0

* Operating system: Linux Mint 19.1

* Browser: Google Chrome Version 75.0.3770.100 (Official Build) (64-bit)

## Description

When using `Tilemap.convertLayerToStatic` multiple times the function throws an error.
```
Uncaught TypeError: Cannot read property 'tilemapLayer' of undefined
    at Tilemap.convertLayerToStatic (VM591 3.17.0.js:100799)
    at Scene.create (eval at <anonymous> (frame.html:45), <anonymous>:35:6)
    ...
```

## Example Test Code

```
var config = {
    type: Phaser.WEBGL,
    width: 800,
    height: 600,
    backgroundColor: '#ffffff',
    parent: 'phaser-example',
    pixelArt: true,
    scene: {
        preload: preload,
        create: create
    }
};

var game = new Phaser.Game(config);
var controls;

function preload ()
{
    this.load.image('walls_1x2', 'assets/tilemaps/tiles/walls_1x2.png');
}

function create ()
{
    var map = this.make.tilemap({ width: 200, height: 200, tileWidth: 32, tileHeight: 32 });

    var tiles = map.addTilesetImage('walls_1x2', null, 32, 64);

    var layer = map.createBlankDynamicLayer('layer1', tiles);
    layer.randomize(0, 0, map.width, map.height, [ -1, -1, -1, 0]);

    var layer2 = map.createBlankDynamicLayer('layer2', tiles);
    layer2.randomize(0, 0, map.width, map.height, [ -1, -1, -1, 0, 1]);

    map.convertLayerToStatic(layer)
    map.convertLayerToStatic(layer2)
}


```

Olliebrown:
When you call `convertLayerToStatic()` the layer passed in will be destroyed.  Inside of the `destroy()` function, there is a call to `tilemap.removelayer()` which removes the layer from the array of layers inside the tilemap.  The created static layer is NOT added into the tilemap in its place.  The end result is that there is one fewer layers in the array of layers in the tilemap.

All of the layers in this array that come after the removed layer will now have an INCORRECT layerIndex.  Each layer has a property `layerIndex` that is set when it is added to the tilemap and is equal to its index in that array.  It is NOT updated when a layer is removed and that causes the problem you are seeing.  I will see about submitting a PR to fix this but below are a few workarounds you can do.

--------

[...]

PR #4836, opened 2019-10-31
Fix for bug related to Issue #4737

Olliebrown:
This PR
* Fixes a bug

Describe the changes below:
- When removing a layer from a tilemap, layerIndex on existing tilemapLayers can become invalid
- Added loop to fix any existing tilemapLayer.layerIndex values by decrementing
- Root cause of Issue #4737


photonstorm:
👍

PR #4840, opened 2019-11-02
Fix Phaser.Geom.Intersects.RectangleToRectangle

gogoprog:
This PR

* Fixes a bug

Describe the changes below:

Hi,
Without this fix, the following happens:
```
+----------------+                
|                |                
|                |+--------------+
|       a        ||              |
|                ||              |
|                ||      b       |
+----------------+|              |
                  |              |
                  +--------------+

> Phaser.Geom.Intersects.RectangleToRectangle( a, b )
> true
```
The `RectangleToRectangle` returns that 2 rectangles next to each other do intersect.
I think this is wrong.

Because `.right` and `.bottom` are outside of the rectangle I think the condition must be fixed.

What do you think about this?

Also the other intersects functions probably need the same kind of fix.

### DIFF
--- a/src/geom/intersects/RectangleToRectangle.js
+++ b/src/geom/intersects/RectangleToRectangle.js
@@ -24,7 +24,7 @@ var RectangleToRectangle = function (rectA, rectB)
         return false;
     }
 
-    return !(rectA.right < rectB.x || rectA.bottom < rectB.y || rectA.x > rectB.right || rectA.y > rectB.bottom);
+    return !(rectA.right <= rectB.x || rectA.bottom <= rectB.y || rectA.x >= rectB.right || rectA.y >= rectB.bottom);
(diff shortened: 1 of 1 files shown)


photonstorm:
Sorry, but we've used this intersects approach for 7 years now and it's in-line with the way other APIs handle it (such as CreateJS, Flash, etc) - that is, our rectangles are considered solid, so they extend to the full width of their faces:

```js
var a = new Phaser.Geom.Rectangle(0, 0, 100, 100);
var b = new Phaser.Geom.Rectangle(100, 0, 64, 64);
```

Which is why the above is considered as intersecting, as they are sharing the same coordinate. Think of the 'thickness' of each face as being 1, not zero (hence the solid rectangle), and it's easier to visualize.

PR #4851, opened 2019-11-06
Fix input-event bug in nested container

rexrainbow:

* Fixes a bug

Describe the changes below:

[Here](https://codepen.io/rexrainbow/pen/wvvmxqj) is a demo of this issue: there is a (white rectangle) game object added into a nested container. This child and its parent's parent both have pointer-up event.
When pointer-up at this game object
- Pointer-up in top-left side, parent's parent container will fire event (Which is not correct, since child should fire event in this case)
- Others, child will fire event.

[Original thread of this issue](https://phaser.discourse.group/t/why-did-i-click-on-the-top-left-corner-of-the-gobtn-button-but-execute-layers-pointerup-event/4149)

### DIFF
--- a/src/input/InputPlugin.js
+++ b/src/input/InputPlugin.js
@@ -2658,6 +2658,8 @@ var InputPlugin = new Class({
                     return indexB - indexA;
                 }
             }
+
+            return listB.length - listA.length;
         }
 
         //  Technically this shouldn't happen, but ...


photonstorm:

https://phaser.discourse.group/t/why-did-i-click-on-the-top-left-corner-of-the-gobtn-button-but-execute-layers-pointerup-event/4149/2

RollinSafary:
clicking on left top part of white rectangle and getting click on parent container. Is it right behavior @rexrainbow ?

rexrainbow:
@RollinSafary No, child game object should receive event.

RollinSafary:
Anyway I think that parent container must receive event too, and as I can see it works as expected.
Please think about a component like scroll-able container which elements are clickable. How you can implement this feature if parent container won't listen events if child is interactive ?

rexrainbow:
@RollinSafary 
In input plugin (scene.input), touched game objects will be sorted via their depth, child will be considered to be above of its parent (also above parent's parent). After sorting, 
- If `input.topOnly` property is set to `true`, only topmost game object will receive touch event
- Otherwise, all touched game object will receive touch event from top to bottom.

To let parent container and child get touch events, just set `input.topOnly` to `false`.

photonstorm:
Hmm, I wonder if this is actually down to the container size. If you add `setSize` into your example, it fixes it (without needing to change the Input Plugin):

```js
    var parent1 = this.add.container()
        .add(parent0)
        .setSize(100, 100)
        .setInteractive({
            hitArea: new Phaser.Geom.Rectangle(0, 0, 100, 100),
            hitAreaCallback: Phaser.Geom.Rectangle.Contains,
            cursor: 'pointer',
            useHandCursor: false
        })
        .on('pointerup', function () { print.text += 'pointerup: parent1\n' })
```

rexrainbow:
@photonstorm 

Found anther bug of `container.setSize(w, h)`.

[Here](https://codepen.io/rexrainbow/pen/LYYqMdJ) is a testing:
- 'pointup' at top-left, or bottom-right side of hit-area all work fine, if it dose not set size (comment out line 20 )
- 'pointup' at bottom-right side of hit-area won't be fired, if container.setSize (turn on line 20 )

Technologically, `container.setSize(w, h)` won't change the result of touching events, since it dose not use the size as hitarea in this test.

photonstorm:
👍

Issue #4863, opened 2019-11-17
ParticleEmitter: Edge zone with a curve


rexrainbow:
## Version

* Phaser Version: v3.20.1

## Description

In the [description](https://github.com/photonstorm/phaser/blob/master/src/gameobjects/particles/zones/EdgeZone.js#L34) of edge zone source.
```
An object instance with a `getPoints(quantity, stepRate)` method returning an array of points.
```
Geometry objects like rectangle, circle, ... all have  `getPoints(quantity, stepRate)` method, they wok fine in edge zone.
But curve object only has [getPoints(quantity)](https://github.com/photonstorm/phaser/blob/master/src/curves/Curve.js#L302), therefore it could not work in stepRate mode. This method will return an empty point array since `quantity` is `0`.

## Example Test Code

[In official example](https://labs.phaser.io/edit.html?src=src\game%20objects\particle%20emitter\edge%20zone%20from%20curve.js), 

Change line 40: 
```javascript
emitZone: { type: 'edge', source: curve, quantity: 48, yoyo: false }
```
to
```javascript
emitZone: { type: 'edge', source: curve, quantity: 0, stepRate:30, yoyo: false }
```
to use `stepRate` instead of `quantity`

rexrainbow:
PR merged, issue solved.

PR #4864, opened 2019-11-18
Curve improvement

rexrainbow:

* Adds a new feature

Describe the changes below:

1. Add stepRate argument in `getPoints` method. ( #4863 )
1. Override default Curve.arcLengthDivisions to `1`, to improve the performance of [line.getLengths()](https://github.com/photonstorm/phaser/blob/master/src/curves/Curve.js#L250).


photonstorm:
👍

PR #4872, opened 2019-11-19
Fix bad parameter in Timeline callbacks and events

### LINKED ISSUE #4810: Parameter is missing from Timeline callbacks
## Version

* Phaser Version: 3.20.1

## Description

If you use a Timeline callback with extra parameters (e.g. `onStart` and `onStartParams`) the first extra parameter is missing when the callback is called.

## Example Test Code

https://codepen.io/samme/pen/rNNMGyR

Should log

    Timeline, 'one', 'two', 'three'

but instead logs

    Timeline, undefined, 'two', 'three'

## Additional Information

https://github.com/photonstorm/phaser/blob/e762a892231bd79279acd1edd9a1196950fb810b/src/tweens/Timeline.js#L286-L288

samme:
* Fixes a bug

Fixes #4810

### DIFF
--- a/src/tweens/Timeline.js
+++ b/src/tweens/Timeline.js
@@ -279,12 +279,10 @@ var Timeline = new Class({
      */
     dispatchTimelineEvent: function (event, callback)
     {
-        this.emit(event, this, this.targets);
+        this.emit(event, this);
 
         if (callback)
         {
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

PR #4873, opened 2019-11-20
Add `type` for Group, Arcade.Group, Arcade.StaticGroup

samme:
This PR

* Adds a new feature

Adds a [type](https://photonstorm.github.io/phaser3-docs/Phaser.GameObjects.GameObject.html#type) property as in the other Game Objects. Could be useful for debugging or plugins.

The types are `Group`, `PhysicsGroup`, `StaticPhysicsGroup`.


photonstorm:
Sure, why not 👍

PR #4875, opened 2019-11-22
CHANGELOG.md typo.

Rob-bie:
`Arade` to `Arcade`.


photonstorm:
Yeah, thanks for your PR, but this happens sometimes, however, the Change Log shouldn't be edited once the version the typo appears in has been published. So the typo will remain for the rest of time I'm afraid! (I fix them / merge if it's _before_ we publish, but once it's released it becomes sacrosanct)

PR #4877, opened 2019-11-25
Add arguments step and out in curve.getSpacedPoints() method

rexrainbow:

* Adds a new feature

Describe the changes below:

Take an optional array as the 3rd parameter in which to store the points results.

### DIFF
--- a/src/curves/Curve.js
+++ b/src/curves/Curve.js
@@ -353,24 +353,37 @@ var Curve = new Class({
      * @method Phaser.Curves.Curve#getSpacedPoints
      * @since 3.0.0
      *
-     * @param {integer} [divisions] - [description]
+     * @param {integer} divisions - The number of evenly spaced points from the curve to return. If falsy, step param will be used to calculate the number of points.
+     * @param {number} step - Step between points. Used to calculate the number of points to return when divisions is falsy. Ignored if divisions is positive.     
+     * @param {(array|Phaser.Math.Vector2[])} [out] - An optional array to store the points in.
      *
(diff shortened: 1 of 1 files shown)


src/curves/Curve.js
```diff
@@ -353,24 +353,37 @@ var Curve = new Class({
      * @method Phaser.Curves.Curve#getSpacedPoints
      * @since 3.0.0
      *
-     * @param {integer} [divisions] - [description]
+     * @param {integer} divisions - The number of evenly spaced points from the curve to return. If falsy, step param will be used to calculate the number of points.
+     * @param {number} step - Step between points. Used to calculate the number of points to return when divisions is falsy. Ignored if divisions is positive.
```
photonstorm: stepRate - also, optional

src/curves/Curve.js
```diff
@@ -353,24 +353,37 @@ var Curve = new Class({
      * @method Phaser.Curves.Curve#getSpacedPoints
      * @since 3.0.0
      *
-     * @param {integer} [divisions] - [description]
+     * @param {integer} divisions - The number of evenly spaced points from the curve to return. If falsy, step param will be used to calculate the number of points.
```
photonstorm: Optional

src/curves/Curve.js
rexrainbow: Fixed

src/curves/Curve.js
rexrainbow: Fixed

photonstorm:
👍

PR #4879, opened 2019-11-26
Override default Curve.getUtoTmapping method

rexrainbow:

* Adds a new feature

Describe the changes below:

Since `u` is equal to `t` in LineCurve, `LineCurve.getUtoTmapping` could return `u` directly, to reduce calculation.


photonstorm:
👍

PR #4887, opened 2019-11-28
Bitmaptext: Add methods of GetBounds

rexrainbow:

* Adds a new feature

Describe the changes below:

Add methods of GetBounds into Bitmaptext game object.

### DIFF
--- a/src/gameobjects/bitmaptext/static/BitmapText.js
+++ b/src/gameobjects/bitmaptext/static/BitmapText.js
@@ -42,7 +42,9 @@ var Render = require('./BitmapTextRender');
  *
  * @extends Phaser.GameObjects.Components.Alpha
  * @extends Phaser.GameObjects.Components.BlendMode
+ * @extends Phaser.GameObjects.Components.ComputedSize
  * @extends Phaser.GameObjects.Components.Depth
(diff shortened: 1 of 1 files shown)


photonstorm:
I'm in two minds about this. While it makes sense for Static Bitmap Text, it will return incorrect results for Dynamic Bitmap Text (which extends Static, so will gain this component as well)

rexrainbow:
@photonstorm You are right. There is no easy way to get bounds of all characters in Dynamic-Bitmaptext game object. 
- How about return null for all GetBounds method in Dynamic-Bitmaptext game object?
- Or just leave it for user, user could add ComputedSize and GetBounds methods manually.

photonstorm:
The problem is that we won't be able to override the documentation that is generated for Dynamic BitmapText to make it known that `null` is what's being returned :(

rexrainbow:
PR closed, since I don't have good solution either. :(

rexrainbow:
Here is the sample code of injecting `getBounds` method manually.

```javascript
const Components = Phaser.GameObjects.Components;
Phaser.Class.mixin(
    Phaser.GameObjects.BitmapText,
    [
        Components.ComputedSize,
        Components.GetBounds
    ]
);
```

PR #4893, opened 2019-12-02
Support for loading spine files via manifest

FostUK:
This PR
* Fixes a bug

CHANGES:
* Adds loader prefix to textures
* removes 'SP' prefix
* Removes atlas extension via regex rather than string trim
* Loads images from atlas path

This allows spine files to be loaded via a manifest, e.g:
```
    "spine": {
        "prefix": "spine_",
        "files": [
            {
                "type": "spine",
                "key": "owl",
                "jsonURL": "shared/owl/export/owl-pro.json",
                "atlasURL": "shared/owl/export/owl-pma.atlas"
            }
        ]
    }
```


photonstorm:
👍

PR #4900, opened 2019-12-04
Correct docs for Phaser.Math.GetSpeed()

### LINKED ISSUE #4895: Speed calculation is wrong
* Phaser Version: 3.19

## Description
The calculation of speed is incorrect in Phaser.Math.GetSpeed.
The current calculation is (distance / time) / 1000 
It should be (d/t) * 1000

samme:
This PR

* Updates the Documentation

From the [examples](http://labs.phaser.io/edit.html?src=src/pools\bullets.js 'Bullet Pool') I'm guessing that GetSpeed's calculation is correct and the docs are incorrect.

Input `time` is seconds and output is distance per ms. 

Fixes #4895

### DIFF
--- a/src/math/GetSpeed.js
+++ b/src/math/GetSpeed.js
@@ -5,15 +5,19 @@
  */
 
 /**
- * Calculate the speed required to cover a distance in the time given.
+ * Calculate a per-ms speed from a distance and time (given in seconds).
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

PR #4911, opened 2019-12-18
Fixed semicolon in updateList doc

javigaralva:
This PR

* Updates the Documentation

Fixed semicolon in the `updateList` attribute doc in order to generate correctly the TS definition file.


photonstorm:
👍

PR #4915, opened 2019-12-19
Fix alpha for children of Container

laineus:

* Fixes a bug

Describe the changes below:

# Issue https://github.com/photonstorm/phaser/issues/4916

Alpha of Container doesn't work for the children.

Demo:

- 3.19.0 https://codepen.io/laineus/pen/yLLqBOO
- 3.20.1 https://codepen.io/laineus/pen/JjjZgqJ

# Solution

I believe any Graphics can not be applied alpha in quad values.
But Container is setting quad alpha to the children kind of all.

So I checked if the child is Sprite to decide whether to set single value or quad values.

### DIFF
--- a/src/gameobjects/container/ContainerWebGLRenderer.js
+++ b/src/gameobjects/container/ContainerWebGLRenderer.js
@@ -52,6 +52,7 @@ var ContainerWebGLRenderer = function (renderer, container, interpolationPercent
         renderer.setBlendMode(0);
     }
 
+    var alpha = container._alpha;
     var alphaTopLeft = container.alphaTopLeft;
     var alphaTopRight = container.alphaTopRight;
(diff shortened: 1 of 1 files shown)


src/gameobjects/container/ContainerWebGLRenderer.js
```diff
@@ -72,6 +73,8 @@ var ContainerWebGLRenderer = function (renderer, container, interpolationPercent
             continue;
         }
 
+        var quadAlpha = child.constructor.name === 'Sprite';
```
laineus: I'm not sure are there other gameObjects that can be applied alpha in quad.

photonstorm:
Thanks for this PR - as you know from the issue, this has been resolved in a different way (by addressing the underlying problem), but without your PR and issue it would never have been found, so thank you :)

PR #4918, opened 2019-12-21
fix(types): missing matterjs types

### LINKED ISSUE #4199: Missing MatterJS TypeScript type definitions for Phaser 3
I tried Phaser 3.15.1 with TypeScript with definitions from here: https://github.com/photonstorm/phaser3-docs/tree/master/typescript

One missing piece is the type definition for MatterJS which is not existent except some empty classes.

There are MatterJS type definitions: https://www.npmjs.com/package/@types/matter-js

Unfortunately they don't work together. In phaser.d.ts the namespace MatterJS is used in /@types/matter-js it is Matter. Just renaming MatterJS to Matter doesn't work. I guess MatterJS is a wrapper anyway.

Long story short: it would be nice to have TypeScript type definitions for MatterJS in Phaser 3

briwa:
* Fixes a bug (with the types), fixes #4199

First of all, thank you for this great library, been using it since 2014 (on and off tho).

I noticed that the namespace `MatterJS` in `phaser.d.ts` isn't linking to the `Matter` type definitions properly. Also, a few instances from `scene.matter` should use `typeof` instead of referring to it as its class instance, since it's only getting the static methods of the classes.

Since the fix requires linking the MatterJS type definition in `phaser.d.ts` file, I am also suggesting to use the type from the `@types/matter-js` itself instead of having them locally in the repo. That way, it will always be up-to-date. However, if vendoring it is intentional in the first place, let me know. I could easily revert the commit.

I'm not sure if this is the right approach but it works for my use case. Would love if you point me to the right direction (maybe the point was to merge the declarations?).

Thank you.


photonstorm:
Thanks for this. I'd love to merge it, but the moment the new TS defs are generated it will all be lost. We need to find a way to make these changes permanent. Also, we don't use the mainline Matter branch, so the @types defs are not fully accurate for our needs, so I'm more than happy to maintain our own local defs and keep those up to date. The library barely changes as it is, so it won't be much of an effort.

Updating the TS defs parser code to automatically include the matter defs should be easy. Those `typeof` checks less so, as they need to be made in the JSDocs directly (and I'm really not sure how we'll manage that)

photonstorm:
Thanks for the work you did on this PR. I'm going to close it off as I've now created new Matter JS defs that work properly with the 3.22 release. You can find them them in the types folder if you'd like to check them out in advance of release.

briwa:
@photonstorm Hey, thank you for the fix! I haven't got around to take a look at this again. Much thanks.

PR #4919, opened 2019-12-23
`deltaSmoothingMax` should at least larger than 0

jcyuan:
this variable is for accumulating avg FPS, but if user set 'deltaHistory' to 0 in the GameConfig, then in `step` method of `TimeStpe.js` it will get an NaN value.

https://github.com/photonstorm/phaser/blob/64690cf987937662f23aaa1fe536f1b166093b6e/src/core/TimeStep.js#L559-L567


jcyuan:
close this, and will make a clearer PR

PR #4926, opened 2019-12-24
Add Phaser.Math.Distance methods

samme:
This PR

* Adds a new feature

Four new methods:

- BetweenPoints()
- BetweenPointsSquared()
- Chebyshev()
- Snake()

[Chebyshev][1] and [snake][2] are worth having in a game library, IMHO, and they are pretty tiny. 😃 

Fun fact: Chebyshev distance is the effective distance for Arcade Physics bodies moving at [maximum velocity][3].

[1]: https://en.wikipedia.org/wiki/Chebyshev_distance
[2]: https://en.wikipedia.org/wiki/Taxicab_geometry
[3]: https://photonstorm.github.io/phaser3-docs/Phaser.Physics.Arcade.Body.html#maxVelocity__anchor

### DIFF
--- a/src/math/distance/DistanceBetweenPoints.js
+++ b/src/math/distance/DistanceBetweenPoints.js
@@ -1,0 +1,26 @@
+/**
+ * @author       samme
+ * @copyright    2019 Photon Storm Ltd.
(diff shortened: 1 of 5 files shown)


photonstorm:
Lovely - why not :)

PR #4929, opened 2019-12-27
`deltaSmoothingMax` should at least larger than 0

jcyuan:
this variable is for accumulating avg FPS, but if user set 'deltaHistory' to 0 in the GameConfig, then in `step` method of `TimeStpe.js` it will get an NaN value.

https://github.com/photonstorm/phaser/blob/64690cf987937662f23aaa1fe536f1b166093b6e/src/core/TimeStep.js#L559-L567

### DIFF
--- a/src/core/TimeStep.js
+++ b/src/core/TimeStep.js
@@ -313,7 +313,7 @@ var TimeStep = new Class({
          * @default 10
          * @since 3.0.0
          */
-        this.deltaSmoothingMax = GetValue(config, 'deltaHistory', 10);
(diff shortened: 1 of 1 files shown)


photonstorm:
3.22 has a new property `smoothStep` which allows you to turn the delta smoothing off, rather than setting the history size to zero. Might be easier to just detect if they try setting history size to zero, then toggle this property off automatically.

jcyuan:
sounds reasonable!
got it~ will close this.

Issue #3882, opened 2018-07-31
Phaser.Tilemaps.Tile tint property seems to not work properly


plissken2013es:
Hi.

When I set the tint prop from a tilemap tile, it seems to modify the final color displayed on screen.
I've tried to use the same colors on a Graphics layer with MULTIPLY blend mode to "tint" my tiles, and it works fine.

You may see the example with commented code here: https://codesandbox.io/s/04y572xxmw

Thanks for your time.

mapsandapps:
I'm having the same issue, where tiles are not getting tinted the color I would expect.

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

PR #4935, opened 2019-12-30
Changed tile getTint function to use getTintAppendFloatAlphaAndSwap

Aedalus:

* Fixes a bug

Describe the changes below:

Currently tinting tilemap tiles gives unexpected colors when rendered, as referenced in issue 3882 (https://github.com/photonstorm/phaser/issues/3882). If I tried to tint a plain white tile blue for instance, it came out closer to orange. Experimenting a bit, it appeared that the "r" and "b" values of a tint were roughly being swapped at some point.

Diving into the source code, I found that the tile tint function was currently using the `Utils.getTintAppendFloatAlpha` function to calculate the tint value. I noticed there was a similar `Utils.getTintAppendFloatAlphaAndSwap` function. When tried, it appeared to fix the original issue with colors appearing as the original tints.

I'd very much like a second set of eyes on this change however, since I'm not familiar with the full Phaser rendering engine. I'd hate if the true issue was elsewhere in the codebase, and this was just a case of two wrongs making a right!

### DIFF
--- a/src/tilemaps/dynamiclayer/DynamicTilemapLayerWebGLRenderer.js
+++ b/src/tilemaps/dynamiclayer/DynamicTilemapLayerWebGLRenderer.js
@@ -36,7 +36,7 @@ var DynamicTilemapLayerWebGLRenderer = function (renderer, src, interpolationPer
     var gidMap = src.gidMap;
     var pipeline = src.pipeline;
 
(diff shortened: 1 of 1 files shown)


photonstorm:
Have looked into this and you're right, the original tint colors expected them in BGR format, so using the `AndSwap` function, in this case, is the correct approach 👍

Issue #4689, opened 2019-08-05
Phaser full screen with more than one element


ivopc:
I've used the native Phaser ScaleManager to get a full screen game, but the native Phaser ScaleManager only full screen the game canvas element, I need more than one, because I am using others HTML elements to compose the game UI.
I've try to make full screen by myself, but for some reason Phaser is blocking this, here the code:
```
var toggleFullScreen = function () {


    var elem = document.querySelector("#game");

    elem.requestFullscreen = elem.requestFullscreen || elem.mozRequestFullscreen || elem.msRequestFullscreen || elem.webkitRequestFullscreen;

    if (!document.fullscreenElement) {
        elem.requestFullscreen()
            .then({})
            .catch(err => {
                console.log(`Error: ${err.message} (${err.name})`);
            });
    } else {
        document.exitFullscreen();
    };
};
```
When I try to use my own Phaser is going back and forth (bouncing) the full screen (backing to normal view), I think the Phaser is blocking it for some reason, there's no other solution? I tryed to change the fullscreenOptions in Phaser toggleFullscreen API, but there's no complete description for the property.

ivopc:
WOOOW, solution! I called the method this.scale.stopListeners(); before the my own full screen method, like this:

```
this.fullScreen.setInteractive().addInteraction("pointerdown", () => {
    this.scale.stopListeners();
    Elements.toggleFullScreen();
});
```

Issue #4352, opened 2019-02-09
Issue with fullscreen api


ngdevr:
OS: Windows 8.1
Browser: Google Chrome
API version - 3.16.1

I am using following code to enter and leave fullscreen

```
    handleFullScreen() {
        if (this.scale.isFullscreen) {
            this.scale.stopFullscreen();
        } else {
            this.scale.startFullscreen();
        }
    }
```
This function is called on a button click. This works fine and can enter as well as exit fullscreen mode when I use the button but if I use "esc" key to exit fullscreen mode then it no longer works. Debugging it in console gives following error
```

phaser.min.js:1 Uncaught DOMException: Failed to execute 'insertBefore' on 'Node': The new child element contains the parent.
    at initialize.getFullscreenTarget (http://localhost:8080/test/js/phaser.min.js:1:405372)
    at initialize.startFullscreen (http://localhost:8080/test/js/phaser.min.js:1:404881)
    at TheGame.handleFullScreen (http://localhost:8080/test/js/game.js:150:24)
    at initialize.h.emit (http://localhost:8080/test/js/phaser.min.js:1:10074)
    at initialize.processUpEvents (http://localhost:8080/test/js/phaser.min.js:1:627540)
    at initialize.update (http://localhost:8080/test/js/phaser.min.js:1:622087)
    at initialize.updateInputPlugins (http://localhost:8080/test/js/phaser.min.js:1:429636)
    at initialize.queueMouseUp (http://localhost:8080/test/js/phaser.min.js:1:432154)
    at onMouseUp (http://localhost:8080/test/js/phaser.min.js:1:420581)
``` 
Further debugging shows that opening fullscreen mode is inserting a DOM element which is deleted on exit fullscreen mode but if "esc" key is used to exit fullscreen mode, then that element is not getting deleted so next time trying to open fullscreen mode fails.

BigZaphod:
I see this same behavior on macOS / Safari.

rgk:
A fix to this is to set a parent property of the scale object in the game config like this:

config = { scale: { parent: 'parent' ... } ... };

ngdevr:
I already have parent property set in the game config, still this error occurs.

rgk:
The scale part of the config needs its own parent option. I can confirm without it, this error happens and with it, it does not when paired with the correct parent.

ngdevr:
This is what I have in my game config. Please advise what do you think needs a change here

```
    let config = {
        type: Phaser.AUTO,
[...]

PR #4936, opened 2019-12-30
Fixes external Fullscreen API calls being blocked by fullscreenchange listener

AdamXA:

* ~~Updates the Documentation~~
* ~~Adds a new feature~~
* Fixes a bug

Related to: https://github.com/photonstorm/phaser/issues/4689

Fixes external calls to the Fullscreen API using `el.requestFullscreen()` from being blocked by Phaser.

If you have a Phaser game embedded into a web page, as Phaser will not know that requestFullscreen was called, and the `this._requestedFullscreenChange` flag was never set, the onFullScreenChange listener would fire `this.stopFullscreen()` right away.

**Background:**

`this._requestedFullscreenChange` was originally added to fix this issue: https://github.com/photonstorm/phaser/issues/4352
I have replaced this by checking if the browser is in fullscreen through `document.fullscreenElement`.

**How to reproduce the bug:**

You can goto any Phaser game in Chrome (because Chrome does not require it be fired from an input event in the console), and type `document.body.requestFullscreen();`.
The browser will go fullscreen, then it exits fullscreen right away.

### DIFF
--- a/src/scale/ScaleManager.js
+++ b/src/scale/ScaleManager.js
@@ -347,16 +347,6 @@ var ScaleManager = new Class({
         this._createdFullscreenTarget = false;
 
         /**
-         * Internal var that keeps track of the user, or the browser, requesting fullscreen changes.
-         *
-         * @name Phaser.Scale.ScaleManager#_requestedFullscreenChange
-         * @type {boolean}
-         * @private
-         * @since 3.16.2
-         */
-        this._requestedFullscreenChange = false;
-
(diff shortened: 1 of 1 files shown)


FostUK:
@photonstorm - apologies for the pester but any thoughts on this?
We are waiting on the above to implement fullscreen on our games as we need to trigger itexternally.

-thanks.

photonstorm:
👍

PR #4939, opened 2020-01-03
Improve JSDoc typings & improve tsgen with default generic types

### LINKED ISSUE #4801: Phaser.Geom.Line type definition bug
## Version

* Phaser Version: 3.20.0

* Browser: Google Chrome

## Description

Type definition of `getPoints` function is wrong in `phaser.d.ts` file.

current type definition is
`getPoints<O extends Phaser.Geom.Point>(quantity: integer, stepRate?: integer, output?: O): O;`
must be
`getPoints<O extends Phaser.Geom.Point[]>(quantity: integer, stepRate?: integer, output?: O): O;`

Chnapy:

* Adds a new feature
* Fixes a bug

Describe the changes below:

I improved some typings by editing related JSDocs.

I also edited tsgen code to handle generics default types (like `<T extends string = string>`).
It was first for a useless feature: set Phaser.Scene key generic, for adding a constraint on Phaser.ScenePlugin add/start... good in theory but too complicated for nothing.
Anyway, the work is made, I think it would be good to keep it ?
:warning: Note that I had to upgrade dts-dom to 3.3.

fix #4801


photonstorm:
Thanks, happy to merge this all in.

Issue #4759, opened 2019-09-20
Typings for SpinePlugin


Jonchun:
As far as I can tell, typings for the SpinePlugin are not available. 

**[Optional] Do you want to help provide this feature?**
I'd like to contribute this if it's not done. It's possible that I was just not looking in the right place. Assuming it's not already done, do we have any resources on how the phaser.d.ts was generated? I did have a rough WIP of what I'm using for my project currently which I made with https://www.npmjs.com/package/tsd-jsdoc and some manual editing.

[phaser.spineplugin.d.ts.txt](https://github.com/photonstorm/phaser/files/3635379/phaser.spineplugin.d.ts.txt)

photonstorm:
You're right, there are no typings for the plugin itself (although there are for the Spine runtimes). If you'd like to contribute these I'd be more than happy to accept them. We use our own jsdoc parser to generate the current defs, which could be adapted to work on the Spine Plugin too, I guess, but it might be safer to keep it separate.

Jonchun:
I can certainly contribute these sometime. Is the jsdoc parser you guys use in this repo? I noticed that the typings I generated were slightly different in format/style than the ones in `phaser.d.ts`. I don't know if it really matters, but worth mentioning.

For the short-term future, anyone who is like me and is looking to use the spine plugin can probably just use the above attached file as while i haven't tested it extensively, it works mostly for me.

photonstorm:
Our defs generator can be found in the repo in the scripts/tsgen folder.

supertommy:
I've hand-rolled some definition files for the SpinePlugin in this PR: https://github.com/photonstorm/phaser/pull/4948

I can see a case for updating the jsdoc parser to generate them but also a case for keeping it separate as mentioned by @photonstorm 

I'm assuming the matter and existing spine definition files were not auto-generated?

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

photonstorm:
@supertommy Matter defs are hand-rolled. The Spine ones are those provided by Esoteric themselves.

PR #4948, opened 2020-01-09
add SpinePlugin, SpineGameObject, SpineFile d.ts

supertommy:
**This PR**

* Adds SpinePlugin TypeScript definition files
* reference issue #4759 

Adds TypeScript definition files for:

- SpinePlugin
- SpineGameObject
- SpineFile

Added to the `types` folder. This should remove type check errors when using TypeScript and the SpinePlugin for things like `this.add.spine()`, `this.load.spine()`, etc.

I realize all the definition file names are currently lower case and match the corresponding .js files. But the SpinePlugin file names are in Pascal case so I chose that. Happy to change it if lower case or dash separated is preferred.

I've also kept it to 3 separate files to follow the existing `spine.d.ts`, `spine-canvas.d.ts`, and `spine-webgl.d.ts` convention which seems to mirror the existing .js files.

In SpineFile.d.ts I have namespaced `SpineFileConfig` to `Phaser.Loader.FileTypes` to follow the JSDoc (https://github.com/photonstorm/phaser/blob/v3.21.0/plugins/spine/src/SpineFile.js#L15). However, I see it referenced as being in `Phaser.Types.Loader.FileTypes` in other places.

The comments are also not included in the definition files with the thinking that it is better to have a single source for truth in the source files for the comments.

Happy to take guidance on this or any other changes to better match style and conventions.

### DIFF
--- a/types/SpineFile.d.ts
+++ b/types/SpineFile.d.ts
@@ -1,0 +1,18 @@
+declare namespace Phaser.Loader.FileTypes {
+    interface SpineFileConfig {
+        key: string
+        textureURL?: string
+        textureExtension?: string
+        textureXhrSettings?: Phaser.Types.Loader.XHRSettingsObject
+        normalMap?: string
+        atlasURL?: string
+        atlasExtension?: string
+        atlasXhrSettings?: Phaser.Types.Loader.XHRSettingsObject
+    }
+
+    class SpineFile extends Phaser.Loader.MultiFile {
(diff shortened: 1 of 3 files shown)


photonstorm:
Nice! This should be helpful for TS users.

Jonchun:
Thanks! Been busy and haven't been working on my phaser project as much as I wanted. One thing that I did in my hand-rolled version that is not done here is include the mixin components.

(You can see the one I made a while back at #4759)

So for example, with your typings, if I want to access x/y on the spine game object, it won't work:

`TS2339: Property 'x' does not exist on type 'SpineGameObject'.`

supertommy:
@Jonchun  yes, I'll fix that with a new PR.

photonstorm:

https://phaser.discourse.group/t/visual-studio-code-intellisense-with-plugins/1947/4

PR #4949, opened 2020-01-10
Check for disabled body in collideSpriteVsGroup()

samme:
This PR

* Fixes a bug (optimization)

`group.contains()` can be skipped when the colliding body is disabled.


photonstorm:
👍

PR #4951, opened 2020-01-11
Update documentation and one import, to use the consistent name of `Vector2`

bramp:

* Updates the Documentation

Describe the changes below:

Update documentation and one import, to use the consistent name of `Vector2`

There were examples that used Vec2, which wouldn't work. For example `An array of Point or Vector2 objects: [new Phaser.Math.Vec2(x1, y1), ...]` was not copy and pastable.


photonstorm:
👍

PR #4965, opened 2020-01-22
set depth of input debug body

mktcode:

* Fixes a bug

Describe the changes below:

When you enable a game object for input debug, the debug body's depth is not set, so it's 0, which is problematic when dealing with different layers. This sets the depth to the depth of the actual game object. Maybe it should even be set to an arbitrary high number or highest depth found in other game objects.


photonstorm:
👍

Issue #4983, opened 2020-02-04
Inconsistent behavior of Tiled layer "properties" member 


Nightspeller:
## Version
* Phaser Version: 3.22.0
* Operating system: Windows 10
* Browser: firefox

## Description
class LayerData has a member 'properties' which type is specified as 'object', while actual type is inconsistent: in case layer (defined in Tiled) has properties - it is an array of objects, in case it has no properties - an empty object.

Expected behavior: type should be object[] and in case of no properties specified, member should be an empty array.

## Example Test Code

1. Create a Tiled map with 2 layers. On one layer specify some property.
2. On the scene, create a tilemap: `const map = this.make.tilemap({key: 'example}).`
3. Get the layers:  `const layer1 = map.layers[0]; const layer2 = map.layers[1];`
4. Look at the properties member of layer1 and layer2: `console.log(layer1.properties, layer2.properties)`
5. See that one layer properties is an empty object and the other - is an array (typescript also should show you a warning if you try to do anything with that array, like use '.forEach' since properties type is 'object').

## Additional Information
Solution in mind: change type of the 'properties' to object[] in definitions, and in the implementation (**as far as I understand**) change  `properties: GetFastValue(curl, 'properties', {})` to  `properties: GetFastValue(curl, 'properties', [])` in ParseTileLayers function. (again, this is according to my best understanding - dunno if that is all what needed to be done neither)

samme:
That sounds right, you could make a PR. 🙂

Nightspeller:
You know, I actually decided to try it out and create a fix and the PR. Here it is: https://github.com/photonstorm/phaser/pull/4984

photonstorm:
Thanks for opening this issue, and for submitting a PR to fix it. We have merged your PR into the `master` branch and attributed the work to you in the [Change Log](https://github.com/photonstorm/phaser/blob/master/CHANGELOG.md). If you need to tweak the code for whatever reason please submit a new PR.

PR #4984, opened 2020-02-05
Proposed fix for the issue#4983

Nightspeller:
Proposed fix for the issue with inconsistent type of 'properties' member of Tile Layer data.

This PR

* Fixes a bug

Describe the changes below:

See full description in the issue: https://github.com/photonstorm/phaser/issues/4983

Additional information: I did not include recompiled dist files and type definitions, so you will have to generate it yourself. The reason why I did not do it is that when I do it creates files with much more differences than I would expect (my best assumption is that master at the time of merge did not have latest changes compiled into dist either, thus when I do it I include also changes from other people, potentially unwanted.

Also, please note that this fix is only for Tile Layer Data - this issue I faced, fixed and checked - seems to be working. BUT there _might_ be similar issue with src/tilemaps/parsers/tiled/ParseImageLayers.js 'properties' - unfortunately I did not have a chance to confirm and fix it.

Fair warning - this is my first commit to the open source and while I did my best looking into it and testing the result, I can't guarantee that I tested every scenario.

Hope that helps.

### DIFF
--- a/src/tilemaps/mapdata/LayerData.js
+++ b/src/tilemaps/mapdata/LayerData.js
@@ -149,10 +149,10 @@ var LayerData = new Class({
          * Layer specific properties (can be specified in Tiled)
          *
          * @name Phaser.Tilemaps.LayerData#properties
-         * @type {object}
+         * @type {object[]}
          * @since 3.0.0
          */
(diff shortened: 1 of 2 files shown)


photonstorm:
👍

Issue #3618, opened 2018-04-25
"Only orthogonal map types are supported in this version of Phaser"


njt1982:
https://github.com/photonstorm/phaser/blob/28e27647233cfa1fcefa4db1ccc7101a51264cac/src/tilemaps/parsers/tiled/ParseJSONTiled.js#L41

When will non-orthogonal maps be supported? Specifically for my use-case, hexagonal maps?

photonstorm:
Honestly, no idea. It's not on our roadmap yet, but we'd happily take a community contribution for the feature.

njt1982:
@photonstorm ok - thanks for getting back to me. Any tips/guidance/documentation on the best approach to make a start on that?

Best as a third party plugin (any examples on how to make plugins)? Or best to fork Phaser and open a PR to get it into "core"?

rexrainbow:
The different between maps might be transferring position between tileX/tileY and worldX/worldY.

In orthogonal map, tileX could be calculated by worldX, and wordX could be calculated by tileX.
In isometric, staggered, hexagonal map (there are 4 sub-types of hexagonal map), tileX might be calculated by ( worldX, worldY ), and wordX might be calculated by (tileX, tileY)

[Here](https://gist.github.com/rexrainbow/7d7ef4ae83696b4c186ce162b22fc470) are some formulas of transferring between tileX/tileY and worldX/worldY in orthogonal and isometric map.

But adding additional parameter might break current function interface, is there better solution to extend the map supporting?

the-simian:
@njt1982 If you're looking for examples on plugins there's examples here, in particular the global and scene level plugins http://labs.phaser.io/index.html?dir=plugins/&q=

For a scene level plugin, its useful to inherit from `Phaser.Plugins.BasePlugin`. here's a little code sample: 

```
export default class YourPlugin extends Phaser.Plugins.BasePlugin {
  constructor(scene) {
    super('YourPlugin ', scene);
    this.scene = scene;
  }
  
  //reserved method will fire on init
   init(name) {
    console.log(this.scene); //here is your scene.
  }
 
  //arbitrary method
  test() {
    console.log(this.scene); //here is your scene.
  }
}
```

and you register this plugin in your game config like:
```
plugins: {
    scene: [
      {
        key: 'YourPlugin',
        plugin: YourPlugin,
        mapping: 'yourPlugin'
      }
    ]
  },
```
now `yourPlugin` is accessible attached to the scene.

[...]

PR #4992, opened 2020-02-07
Preliminary PR for isometric support

svipal:
Hello ! I've been working on implementing isometric tilemaps for a while now. 

fixes https://github.com/photonstorm/phaser/issues/3618

Summary of the changes and their rationale : 
- Added orientation field to the layer data, to prevent "this.layer.tilemapLayer.orientation" unwieldy calls in the engine code.
- Changed pixelX and pixelY calculations for the tiles in isometric layers in updatePixelXY to display tilemaps correctly.
- Changed WorldToTileXY and TileToWorldXY components to work with isometric tilemaps.
I also changed every component (getTilesAtWorldXY, getTilesWithin...) that used to call both WorldToTileX and WorldToTileY to call the XY alternatives instead, because sadly with isometric coordinates you need to work on both at the same time when converting to/from the world coordinates. 
- Changed the renderers to use the tileset's sizes when displaying tiles (and not the tiles') when the map is isometric.
This may seem weird - but in Tiled the tileset's tiles don't necessarily have the same height as the tiles, it's usually double, to allow for "block/cube" type isometric maps. Check my [test](https://github.com/svipal/svipal.github.io)'s map (in assets/tilemaps) to understand this.
- Adjusted culling for isometric dynamic layers. I felt like the cullpadding didn't really make sense for iso maps so I didn't include it, but I'm not so sure, let me know. As part of this, I added a isoCullDistances attribute to dyn layers in order for the user to decide how strict the culling is.
- Allowed the parser to parse isometric maps.
- Applied the linter ;)
probably more adjustments I forgot.

For staggered and hexagonal maps :
- Changed pixelX and pixelY calculations ,Changed WorldToTileXY and TileToWorldXY components to work with them.
They use the same culling as standard maps with modified values for the vertical axes.
You can do tileToWorldY, worldToTileY separately with them but not the X variants because of the formulas.

Comments : 
Tiled itself [doesn't support proper object layer display](https://discourse.mapeditor.org/t/z-sorting-with-isometric-maps-and-object-layers-in-tiled/2995). Proper being, detecting when objects are behind walls/ when they are not. They are displayed above the map in all cases. 
I might be wrong, but I believe changing it would require a significant rewrite of the renderer, in order to interleave the objects and the tiles. I think it's better to just let users sort it out with Z-indexes.
Thus I decided not to implement it in the engine by default. Is that reasonable ?

Thank you in advance for your feedback !

Edit : 
~Only standard isometric is supported for now, but staggered is just a formality at this point, would come a week later _at most_ after this PR is accepted.~
Staggered and hexagonal map modes are now supported.


svipal:
Hey, since I haven't got any feedback here or on Discord, I went ahead and implemented Hexagonal as well. At this point, I would really like some interaction with the dev team, it's been more than a month and I haven't gotten a single message despite multiple updates. I don't know if you think my implementation is ok, bad, has bad perf, bugs at any points or anything, and if you plan to merge it at any point ?

hexus:
Phaser 3 development is on hold at the moment while @photonstorm is working on Phaser 4. There has been very little commit activity on the repo since 13th Feb for that reason.

This is discussed in a Phaser 4 development report on Patreon:
- https://www.patreon.com/posts/phaser-4-dev-4-34531102

src/tilemaps/ParseToTilemap.js
```diff
@@ -66,6 +66,7 @@ var ParseToTilemap = function (scene, key, tileWidth, tileHeight, width, height,
 
     if (mapData === null)
     {
+        console.log('null mapdata');
```
hexus: This debug statement shouldn't be left behind.

```suggestion
```

src/tilemaps/components/CullTiles.js
```diff
@@ -45,103 +45,243 @@ var CullTiles = function (layer, camera, outputArray, renderOrder)
     var drawTop = 0;
     var drawBottom = mapHeight;
 
+    // we define the isometric culling function as a dummy early on for it to make sense in scope 
+    var inIsoBounds = function () { return true; };
```
hexus: It would be better to define this dummy function elsewhere, instead of inside the function as part of the variable declaration. This will avoid the function definition allocating memory every time `CullTiles` is called.

src/tilemaps/components/TileToWorldX.js
```diff
@@ -33,7 +34,19 @@ var TileToWorldX = function (tileX, camera, layer)
         tileWidth *= tilemapLayer.scaleX;
     }
 
-    return layerWorldX + tileX * tileWidth;
+    if (orientation === 'orthogonal')
+    {
+        return layerWorldX + tileX * tileWidth;
+    }
+    else if (orientation === 'isometric' || orientation === 'staggered' || orientation === 'hexagonal')
```
hexus: I'd suggest defining constants for tilemap layer orientations instead of relying on specific string values here. Take a look at how constants are defined elsewhere in the framework.

hexus:
Some initial feedback for myself that you can work on in the meanwhile:

- Make sure this change doesn't include any unnecessary or inappropriate files, such `.metals/metals.h2.db`
- Make sure there are no changes to the Phaser build files, as these are purposely only built at _release_ time
- Make sure no files are removed unnecessarily, such as `scripts/tsgen/test/bin/game.js`
- Make sure you haven't left behind any debug statements, such as `console.log()` calls

svipal:
Hey ! Thank you, this is perfect. I did a few passes on debug statements, but didn't notice them all.  Will look into moving the iso-culling function inside a component.

src/tilemaps/components/TileToWorldX.js
hexus: It might actually be better to have entirely separate classes, or at least tilemap _components_, to avoid this branching for the sake of performance.

By component, I mean an object or system of some kind that can represent a tilemap orientation and handle their differences, such as coordinate space conversion as demonstrated here.

This means tilemaps can be _composed_ with an orientation, instead of determining them at runtime. Changing an existing tilemap's orientation would then be a case of swapping out the underlying orientation implementation.

svipal: About the constants, it's certainly better than comparing strings.
About splitting the implementation differently, yeah, you're probably right. I don't think it matters _that_ much for performance, (there are tons of other branchings in a lot of often used components ) but it's - really - not clean.

svipal:
Also I want to point out : the new scripts in package.json are scripts I use to test my implementation. I will remove them at the end.

svipal:
Okay, this was very laborious, refactored everything so that the components themselves are no longer exported directly but provide a function depending on the orientation you pass them.
So that there is no switching at runtime, they are stored in the Tilemap itself. (Tilemap.js)

Advantages are that it's a bit faster in Tiled implemented tilemaps and will not slow down existing maps, and that you can use components with an arbitrary orientation at any time if you want to for instance implement your own class with different orientations on different layers , e.g. by doing `TileToWorldXY(orientation)(arguments)`
The tradeback is that in components you have to use the tilemap's implementation instead of directly calling the components themselves. I hope this won't cause trouble - it's one layer of indirection but it's necessary w/o global state - which I did try : 
I had a version that changed the components themselves with a function when the tilemap was initialized. basically changing the module that is exported globally. I didn't then need to change the components that relied on them, and worked very well, However I realized you'd run into conflicts if you tried to use multiple tilemaps with different orientations at the same time, which I guess *could* happen in some cases (games with minimaps, games with mini games or inventories...).
I felt it was too dirty and didn't do it but maybe you would think otherwise. Let me know.

hexus:
I think the choice to avoid a global switch for tilemap orientation was a good one there. Thanks for the refactoring, what you've described sounds great.

Hopefully myself or @photonstorm will have a chance to look over this again soon!

hexus:
image

The size of the PR really needs addressing before it can be reasonably reviewed, by the way. Please try to keep changes as concise as possible without touching _any_ dist or build files, introducing any unnecessary files or removing any files unnecessarily.

svipal:
Sure, will clean up.

svipal:
Clean up done~

hexus:
Much easier to read through. Thanks!

gc:
Hi, just want to say isometric support would be awesome, and thank you for working on this PR

svipal:
hey gc, thanks.
In fact it would help out a lot if you could try out this branch and see if it works for you.

gc:
Any idea how I would try it out? I have no idea where/how I would do that, or if you have example code on how to start using your stuff you added

paaaav:
Well, there is a small demo at svipal.github.io (source: https://github.com/svipal/svipal.github.io) which basically shows how to use it if you read the source. 

But the project does not use the recent isometric phaser branch. The dist directory in svipals phaser branch is also not completely up to date. Best would be to build Phaser by yourself. But it's not too difficult. 

You got to checkout/clone https://github.com/svipal/phaser and then run "npm run-script build". Of course you need npm and whatnot. But that's basically how I did it.

klesun:
@svipal, you rock, man! Will be using your fork in our project till it's merged to main phaser repo ;).

I take it that steps to build are:
```
git clone git@github.com:svipal/phaser.git
cd phaser
npm ci
npm run-script build
```
And resulting bundle file is located in `/build/phaser.js`

I tried it out, and it seems to work. Though I noticed that `layer.setCullPadding(-1, -1)` does nothing with an isometric map, even though `layer.isoCullDistances.x = -1` and `layer.isoCullDistances.y = -1` work ok like on your demo. Not sure if `setCullPadding` is supposed to work on isometric maps, but letting you know, just in case...

svipal:
@klesun
Thanks for the kind words ! I'm really happy you're using the fork. Let me know how it goes.
As for the cull padding, I removed it in the standard isometric map (it works as normal in staggered), and yeah, put the isoCullDistances instead because I thought It was more practical:

> Adjusted culling for isometric dynamic layers. I felt like the cullpadding didn't really make sense for iso maps so I didn't include it, but I'm not so sure, let me know. As part of this, I added a isoCullDistances attribute to dyn layers in order for the user to decide how strict the culling is.

mcfarlanedev:
Hi Svipal,

I love the work that has been involved in this so far - thank you for working on this. I'm currently testing out the isometric feature in your PR for a small project of mine, and the only issue I have with it is that when I click on a tile, my code seems to be logging out information about the adjacent tile - almost as if something is not quite right with where Phaser starts and ends the tile in isometric dimensions. 

I've provided a screenshot below as an example. In the example, when I click on a tile, I just set its alpha to 0,9, just so I can debug the issue. I expect the tile behind my cursor to have its alpha set, but as you can see, the tile adjacent to that has its alpha set instead.

Again, thank you so much, it's good to have a discussion/progression towards native iso support.

image

svipal:
Hello hello ! Thanks for trying out the branch. Could I have more context and if possible take a look at the code to try and debug this ?

KingCosmic:
Svipal the problem McFarland is talking about is evident even in your own demo, mouse selection of tiles is at best. Wonky.

svipal:
Hey KingCosmic, my demo is very old and doesn't actually use the last version of the branch.
Thanks for reminding me. Pushing now.

mcfarlanedev:
@svipal ,

Your demo has the lines I am trying to test commented out, which is ctrlKey console logging of the current tile).

I have copied your phaser.min.js on your demo site to my example demo, and the following occurs on ctrlKeyDown of the same tile. The top of the tile is null on ctrlKeyDown, the middle part of the tile looks correct (x:0), and the bottom third of the tile is reading the tile information of the next tile along (x:1). 

My tile is 104x180.

This is my update function: 

    controls.update(delta);
    var worldPoint = this.input.activePointer.positionToCamera(this.cameras.main);
    var pointerTile = map.worldToTileXY(worldPoint.x, worldPoint.y, true);
    if (ctrlKey.isDown) {
      var selectedTile = map.getTileAt(pointerTile.x, pointerTile.y);
      console.log(selectedTile);
    }

image

mcfarlanedev:
Just thought I'd provide a visual of the problem too:
iso

svipal:
Thanks a lot for the precision. Will work on it tomorrow.

svipal:
Fixed it but forgot to reput the old files for conflicts - will do that a bit later.

mcfarlanedev:
I can confirm that the correct tiles are now being referenced at the expected pointer coordinates. Thanks @svipal ! 

Peek 2020-06-06 00-42

paaaav:
@svipal 
Did you test your branch with a default orthogonal map and colliders? 
```
TypeError: WorldToTileX is not a function
    OrthoWorldToTileXY WorldToTileXY.js:34
    worldToTileXY StaticTilemapLayer.js:1447
    GetTilesWithinWorldXY GetTilesWithinWorldXY.js:33
    getTilesWithinWorldXY StaticTilemapLayer.js:1088
    collideSpriteVsTilemapLayer World.js:2166
    collideHandler World.js:1830
    collideObjects World.js:1732
    update Collider.js:143
    update World.js:962
    emit index.js:203
    step Systems.js:379
    update SceneManager.js:563
    step Game.js:475
    step self-hosted:879
    step TimeStep.js:599
    step self-hosted:875
    step RequestAnimationFrame.js:116
    (Async: FrameRequestCallback)
    step RequestAnimationFrame.js:118
    (Async: FrameRequestCallback)
...
```

Also could you share your thoughts about the drawing issues I get in staggered and isometric I talked about in the referenced issue (https://github.com/photonstorm/phaser/issues/3618)?
> I've seen some – I think – drawing issues. You can see it [here](https://freeimage.host/i/JcJCVn) It's what happens when you draw one or more maps adjacent.

What can I do about it?

svipal:
hello paaav ! Indeed, not with colliders since I reworked the WorldToTile components. Let me fix that real quick.
As for the drawing issues, it seems super weird to me , could I take a look at the code ?

svipal:
@paaav hey ! I've fixed the bug. It actually wasn't affecting the colliders but the orthogonal coords - woops.
Also renewing the request for a code example for the drawing issues :)

paaaav:
@svipal thank you. I will prepare an example but let me first find out if I did something wrong :)
To iterate further: it just happens when you draw multiple tilemaps near each other, but randomly delayed (of course with "scene.make.tilemap()" and creating it's layers simply by createStaticLayer()). Then *sometimes*, **depending on the draw sequence**, and with the tileset "isometric_grass_and_water" (don't know if that plays a role), such artifacts happen, at least with my complex code. I'll let you know in a few days.

svipal:
Sure, let me know. It would be really nice (if you had time !) to make a minimal reproducible example.

svipal:
@photonstorm
I've been monitoring this issue for a long time, have spent a great deal of time on it, multiple users both privately and on this issue have reported to me they used my branch, I've received feedback from one member of the devteam, and have implemented his revisions, have fixed multiple bugs that I've discovered from my own tests and/or that were reported and I still don't know if and when this will be merged. I won't lie, I did this because there was a bounty on it - at the time I was in financial hardship and was hoping to get the money in around 3 months time. It's been instead 5 months,  I'm even worse off, and If you told me when you're planning to merge this, if you are, or if I should just drop it and move on it would really, *really*, **really** help me out.

paaaav:
@photonstorm Hey there :) In the linked issue you stated "Honestly, no idea. It's not on our roadmap yet, but we'd happily take a community contribution for the feature." I assume the declaration still stands, but I do wonder why you did not comment on it so far. I'd really like it to get into Phaser 3. And of course v4, when the time comes, naturally. 
It's been a while, almost one year since I started giving a bounty on it. And I was really really happy that @svipal accepted it 8 months ago. If svipal's contribution cannot be accepted, it's useless to pursue this path further for all participants, a clear statement will save everyone's time.
Whatever the situation is, it is important that we make the best out of it. Thank you!

photonstorm:
I appreciate the effort @svipal has put into this feature, however, I'm not willing to merge it directly into master _as it stands_ for a number of reasons, **but** I will merge it if you move it to its own folder, so it doesn't change anything that currently exists, but instead lives as its own thing in its own folder.

By this, I mean create a new top-level folder `isotilemaps` and put all of your code from your branch in there.

I can then offer a build script that pulls in the iso tilemap system instead of the current one, so devs who want it can use it easily.

The main reasons I won't merge as-is:

1) There are lots of cases where the coding style doesn't adhere to that used in Phaser (and I feel it would be unfair to make you change them all)
2) It has lots of conditionals that will adversely impact 'non-iso' tilemaps, which are far more common
3) Incomplete documentation / parameters / version numbers

If I merge this directly, as is, support and issues for it will land on my shoulders, as devs rightly won't distinguish it from any other standard Phaser API feature. If I offer this as a special custom-build, at least I can flag to devs using it they should be aware of this fact up front.

By moving it to its own folder, it will still technically be 'merged', meaning you'd still claim the bounty.

However, I notice it doesn't work with the changes made to the 3.50 renderer, so it may be worth quickly updating those first.

svipal:
Thanks a lot for the response.

> There are lots of cases where the coding style doesn't adhere to that used in Phaser (and I feel it would be unfair to make you change them all)
Honestly I don't mind changing that. I used the linter, but it's hard with coding style. Don't hesitate to tell me what you don't like.

>Incomplete documentation / parameters / version number

I will double check, but if you already know exactly where I missed parameters please let me know.
As for version numbers, I was planning to change them when I knew what version the code would be in.

>It has lots of conditionals that will adversely impact 'non-iso' tilemaps, which are far more common
I was under the impression that the changes suggested by  @hexus prevented the code from impacting non iso tilemaps. 
Again, will verify this a bit later. 

>If I merge this directly, as is, support and issues for it will land on my shoulders, as devs rightly won't distinguish it from any other standard Phaser API feature. If I offer this as a special custom-build, at least I can flag to devs using it they should be aware of this fact up front.
I understand that, and I don't mind keeping maintaining this feature for a while. I also completely understand you preferring to put this as a special build.

>By moving it to its own folder, it will still technically be 'merged', meaning you'd still claim the bounty.
>However, I notice it doesn't work with the changes made to the 3.50 renderer, so it may be worth quickly updating those first.

Thanks for this, I can take care of this this a bit later this week.

hexus:
> I was under the impression that the changes suggested by @hexus prevented the code from impacting non iso tilemaps.
Again, will verify this a bit later.

Looking at the code as it stands now, there are still conditionals checking for orientation constants. This can have an impact on performance.

I did suggest providing _components_ for tilemap orientation instead, here: https://github.com/photonstorm/phaser/pull/4992#discussion_r396077664

I don't imagine such an approach to be trivial, though I thought that could help regarding branching.

I think it's great that @photonstorm would like to merge it and make it a build option though. I'd just suggest being careful around the duplicated code, and maintaining it alongside any future changes to tilemaps.

photonstorm:
Ideally, it would have been made as a plugin, offering a new type of Tilemap (so it would have been its own unique Game Object), but it's far too late to do that now and unreasonable to request it, so I think the custom build option is the best route for all concerned.

svipal:
Hey hexus, thanks for joining the convo.
> 
> 
> > I was under the impression that the changes suggested by @hexus prevented the code from impacting non iso tilemaps.
> > Again, will verify this a bit later.
> 
> Looking at the code as it stands now, there are still conditionals checking for orientation constants. This can have an impact on performance.

Really ? Where ?


> I did suggest providing _components_ for tilemap orientation instead, here: [#4992 (comment)](https://github.com

You did, and this is why I did what i did. There should no longer be any checks after the tilemap is initialized.

svipal:
So I looked at the code - I see what I missed, the renderer is doing the orientation checking. 
I will implement the same kind of one-time-initialization for the renderer at the same time I fix the code to be compatible with the new renderer.

svipal:
I'm done with this. Let me know @photonstorm if we need anything else to proceed.

photonstorm:
@svipal thanks, this looks fine. Before I can merge it, could you please do what I requested in my previous message? Specifically, this:

"I will merge it if you move it to its own folder, so it doesn't change anything that currently exists, but instead lives as its own thing in its own folder.

By this, I mean create a new top-level folder `isotilemaps` and put all of your code from your branch in there. So the `isotilemaps` folder would contain everything that the current `tilemaps` folder has in it, with all of your changes over the top of that code. If that doesn't make sense, grab me on Discord."

Basically, in order to merge, I need the `tilemaps` folder to be exactly what is currently on the `master` branch, and I need _your_ `tilemaps` folder to be renamed `isotilemaps` please. That way I can create two build scripts, one with, and one without. Yes, it means there will be lots of duplicate files across the two folders, but that's absolutely fine, as there will never be the case where both of them are in a single build together.

photonstorm:
@svipal actually, never mind. I'll do it myself.

PR #4995, opened 2020-02-09
Fix checkCollision none for tile collisions; and optimizations

samme:
Bug Fix
-------

- [Arcade.Body#checkCollision.none](https://photonstorm.github.io/phaser3-docs/Phaser.Physics.Arcade.Body.html#checkCollision) did not prevent collisions with tiles. Now it does.

Optimizations
-------------

- collideSpriteVsGroup() exits early when the sprite has `checkCollision.none`, skipping an unnecessary iteration of the group.
- In collideSpriteVsGroup(), when looping through the tree results, it skips bodies with `checkCollision.none`. This is a bit of a trade-off. In the best case, it saves an unnecessary iteration of the group. In the worst case, it duplicates the same test that would be made later in `separate()`.

### DIFF
--- a/src/physics/arcade/World.js
+++ b/src/physics/arcade/World.js
@@ -2133,7 +2133,7 @@ var World = new Class({
     {
         var body = sprite.body;
 
-        if (!body.enable)
+        if (!body.enable || body.checkCollision.none)
         {
             return false;
         }
--- a/src/physics/arcade/World.js
+++ b/src/physics/arcade/World.js
@@ -1948,9 +1948,9 @@ var World = new Class({
             {
                 bodyB = results[i];
 
-                if (bodyA === bodyB || !bodyB.enable || !group.contains(bodyB.gameObject))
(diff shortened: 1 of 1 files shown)


src/physics/arcade/World.js
```diff
@@ -1920,7 +1920,7 @@ var World = new Class({
     {
         var bodyA = sprite.body;
 
-        if (group.length === 0 || !bodyA || !bodyA.enable)
+        if (group.length === 0 || !bodyA || !bodyA.enable || bodyA.checkCollision.none)
```
hexus: Is it worth adding the same check for `bodyB` inside the loop below, or am I missing something?

samme: I suppose so.

samme: Pushed this change for `bodyB`.

hexus: Looks good!

photonstorm:
👍

PR #4998, opened 2020-02-10
Fix inputDebug shape position when the target is inside a container

scott20145:
The debug shape use gameObject's local position as the reference, but position of the shape may be wrong when the target is inside a container.
[DEMO] (https://codepen.io/scott20145/pen/JjddRZd)


hexus:
This looks great, thanks

photonstorm:
Thanks for your PR. I've merged it and updated it because calls to `getWorldTransformMatrix` are really expensive, so I'm now caching this matrix. It also needs to be used for the scale and rotation, so I've applied it there as well.

PR #5013, opened 2020-02-18
Remove Sound Manager's game event listeners when destroyed

samme:
This PR

* Fixes a bug

Sound Managers were still listening to Game BLUR, FOCUS, and PRE_STEP events after being destroyed. Wouldn't really cause problems unless you destroy a Sound Manager separately from destroying the game.

I extracted into two new, protected methods:

- Phaser.Sound.BaseSoundManager#onGameBlur
- Phaser.Sound.BaseSoundManager#onGameFocus


photonstorm:
Have made these changes locally, thanks.

PR #5041, opened 2020-03-11
Fix undefined state during processing of AudioFile and VideoFile

samme:
This PR

* Fixes a bug

AudioFile and VideoFile had `state` undefined instead of FILE_PROCESSING.


photonstorm:
👍

PR #5047, opened 2020-03-16
added stepRate check for GetPoints methods

jdcook:
This PR
* Fixes a bug

Describe the changes below:
If an emitter is initialized with an emitZone with a quantity and stepRate of 0, GetPoints() inside of the shape class divides quantity by 0 and it becomes infinity. This change will avoid the division in this case and leave quantity as 0.

You shouldn't pass quantity and stepRate as 0 at the same time in the first place, but no particles showing up is preferable to crashing in an infinite loop.


photonstorm:
👍

PR #5051, opened 2020-03-23
Add Tiled point object and change offset in createFromObjects()

samme:
This PR

* Adds a new feature
* Fixes a bug?

### Current Behavior

[Tiled point objects](https://doc.mapeditor.org/en/latest/reference/json-map-format/#point-example) are parsed to

    { rectangle: true, width: 0, height: 0, … }

When creating a sprite from a Tiled object with `{ width: 0, height: 0 }, [createFromObjects()](https://photonstorm.github.io/phaser3-docs/Phaser.Tilemaps.Tilemap.html#createFromObjects) maps the Tiled object origin onto the sprite origin using the sprite's dimensions (which are not zero). This is [probably not what authors want](https://phaser.discourse.group/t/objects-in-group-created-via-createfromobjects-in-class-spawning-at-incorrect-coordinates/4868). 

### New Behavior

Tiled point objects are parsed to

    { point: true, width: 0, height: 0, … }

When creating a sprite from a zero-size object, createFromObjects() doesn't adjust the position at all. The new sprite is given coordinates identical to the object.

I think this is more sensible, but it is a "breaking" change for projects using createFromObjects() with zero-dimensioned Tiled objects.

### Internal Changes

I [removed some redundant assignments](https://github.com/photonstorm/phaser/pull/5051/commits/df1537fa366ea039c23306799ed44c294eea2709) in Tiled.ParseObject().

### DIFF
--- a/src/tilemaps/Tilemap.js
+++ b/src/tilemaps/Tilemap.js
@@ -671,9 +671,10 @@ var Tilemap = new Class({
                 if (obj.height) { sprite.displayHeight = obj.height; }
 
                 // Origin is (0, 1) in Tiled, so find the offset that matches the Sprite's origin.
+                // Do not offset objects with zero dimensions (e.g. points).
                 var offset = {
-                    x: sprite.originX * sprite.displayWidth,
-                    y: (sprite.originY - 1) * sprite.displayHeight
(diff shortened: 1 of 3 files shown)


photonstorm:
👍

Issue #1458, opened 2014-12-15
can't load audio via blob url


aressler38:
Hi, I have an audio file that is loaded into the browser but not by Phaser. Nevertheless, I want to add it to my game using `game.add.audio()`, so I've created a BLOB url using `URL.createObjectURL`, which creates something that looks like `blob:http%3A//{hostname}/209f7670-286f-42e5-9a61-0584bd148728`. Notice that there is no file extension in the blob URL that the browser creates. This type of extensionless url is causing Phaser to hickup when loading the audio file. I've tracked the issue to this loader method in the unminified `phaser.js` file: 

```
50892     /**
50893     * Private method ONLY used by loader.
50894     * 
50895     * @method Phaser.Loader#getAudioURL
50896     * @private
50897     * @param {array|string} urls - Either an array of audio file URLs or a string containing a single URL path.
50898     */
50899     getAudioURL: function (urls) {
50900 
50901       console.log(urls)
50902         var extension;
50903         
50904         if (typeof urls === 'string') { urls = [urls]; }
50905         
50906         for (var i = 0; i < urls.length; i++)
50907         {   
50908             extension = urls[i].toLowerCase();
50909             extension = extension.substr((Math.max(0, extension.lastIndexOf(".")) || Infinity) + 1);
50910             
50911             if (this.game.device.canPlayAudio(extension))
50912             {   
50913                 return urls[i];
50914             }
50915         
50916         }
50917         
50918         return null;
50919     
50920     },
```

Because the blob URL provides no extension, the `getAudioURL` method returns a `null`; I imagine this would also affect server routes like `/audio/track/1`. 
Would it be better to fetch the audio file and look at the response header's content-type field before deciding whether `this.game.device.canPlayAudio`? 

I've been able to successfully use BLOB urls to load images, spritesheets, and sprites through Phaser.

PR #5052, opened 2020-03-23
Fix loading audio and video with blob urls

aucguy:
This PR

- Fixes a bug (loading audio and video with blob urls fails)

Passing a blob url (ie returned by URL.createObjectURL) as the url to `this.load.audio` or `this.load.video` would cause Phaser to load a different url than the one specified. Additionally, `this.load.path` would incorrectly be prepended to blob urls.

`AudioFile.getAudioURL` and `VideoFile.getVideoURL`, are expected to return an object with two properties: url and type. However when passed a blob url they instead return a string. The return value's `url` property is inspected to get the url. Normally this would just result in the url. However, for blob urls, this results in `undefined` since a string was returned. If the url property is `undefined` then a url derived from the asset key is used instead, hence the incorrect url. Now, the the `getXUrl` functions return the correct type. Additionally,  `this.load.path` is not prepended if the url is a blob url.

A live code demo can be found at [https://jsfiddle.net/aucguy/a2kte9nb/6/](https://jsfiddle.net/aucguy/a2kte9nb/6/). Using the assets causes errors and you can see the incorrect url in under network in the inspector. This was tested in chrome.
This PR is related to #1462, #1458 and #2234. This seems to have been an issue in phaser 2.


photonstorm:
👍

AdamXA:
Thanks! Noticed this not working, saw it was fixed already in master.

Saved me a lot of time! 👍

jannis6023:
Why does this fix not apply anymore? I am just loading a blob URL and see this not working anymore

PR #5055, opened 2020-03-25
Clamp audio tag volume in HTML5AudioSound

samme:
This PR

* Fixes a bug

When using HTML5 Audio, setting game volume or sound volume outside [0, 1] would throw an error:

> IndexSizeError: The index is not in the allowed range


photonstorm:
👍

PR #5058, opened 2020-03-27
adding index to spine atlases when loading through asset pack

jdcook:
This PR

* Fixes a bug

Describe the changes below:

Appended the atlas index when loading spine atlases from asset packs, to match how they are loaded when manually calling load.spine.

Without this, the atlas is not loaded when a new SpineGameObject is initialized.


photonstorm:
👍

PR #5066, opened 2020-04-01
Fix scene update before create

### LINKED ISSUE #5065: Scene update is called before create
## Version

* Phaser Version: 3.22.

## Description

If you restart or stop/start a scene and then queue at least one new file in `preload()`, the scene's `update()` function is called before its `create()` function, likely causing an error.

## Example Test Code

https://codepen.io/samme/pen/MWwRYpd

Click at the prompt, then see console. Assertion should pass: scene status is [RUNNING (5)](https://photonstorm.github.io/phaser3-docs/Phaser.Scenes.html#.RUNNING).

## Additional Information

Scenes.Systems#sceneUpdate needs to be reassigned NOOP before starting the loader.

samme:
This PR

* Fixes a bug

Fixes #5065

### DIFF
--- a/src/scene/SceneManager.js
+++ b/src/scene/SceneManager.js
@@ -452,6 +452,8 @@ var SceneManager = new Class({
         var sys = scene.sys;
         var settings = sys.settings;
 
+        sys.sceneUpdate = NOOP;
+
         if (scene.init)
         {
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

Issue #5069, opened 2020-04-02
Lock Orientation breaks in Firefox


123survesh:
## Version

* Phaser Version: 3.21.0-FB

* Operating system: Debian 10

* Browser: Firefox 68.6.0esr (64-bit)

## Description

Get the following error while using game.scale.lockOrientation() in firefox (both in mobile and desktop).
**'TypeError: 'mozLockOrientation' called on an object that does not implement interface Screen.'**

(Works fine in Chrome and Safari)

## Example Test Code

game.scale.lockOrientation(Phaser.Scale.PORTRAIT);

## Additional Information

Checked online about this issue, seems that screen.lockOrientation is deprecated and not supported by latest versions of firefox.
https://developer.mozilla.org/en-US/docs/Web/API/Screen/lockOrientation

123survesh:
I checked **ScreenOrientation.lock**, it does not exist in Firefox nor chrome at this moment.

photonstorm:
Thanks for opening this issue, and for submitting a PR to fix it. We have merged your PR into the `master` branch and attributed the work to you in the [Change Log](https://github.com/photonstorm/phaser/blob/master/CHANGELOG.md). If you need to tweak the code for whatever reason please submit a new PR.

PR #5070, opened 2020-04-02
Fix for issue #5069, where using lockOrientation breaks in firefox

123survesh:

* Fixes a bug #5069 

Describe the changes below:

Using **game.scale.lockOrientation('portrait');** in firefox gave the error **'TypeError: 'mozLockOrientation' called on an object that does not implement interface Screen.'**.

It was being caused as the method **screen.lockOrientation** was being stored in a variable and used. I wrote a patch that uses **screen.lockOrientation** directly if in firefox.

### DIFF
--- a/src/scale/ScaleManager.js
+++ b/src/scale/ScaleManager.js
@@ -697,7 +697,7 @@ var ScaleManager = new Class({
 
         if (lock)
         {
-            return lock(orientation);
+            return lock.call(screen, orientation);
         }
 
         return false;


src/scale/ScaleManager.js
```diff
@@ -693,14 +693,19 @@ var ScaleManager = new Class({
      */
     lockOrientation: function (orientation)
     {
-        var lock = screen.lockOrientation || screen.mozLockOrientation || screen.msLockOrientation;
+        var lock = screen.lockOrientation || screen.msLockOrientation;
+        var locked = false;
 
         if (lock)
         {
-            return lock(orientation);
```
samme, line 700: I think changing this statement only would fix it:

```
return lock.call(screen, orientation);
```

123survesh: Ah ok, makes sense. I will try it and update the PR. Thanks @samme

123survesh: I made the above suggested change and tested it. It works well. I have created a PR with those changes. I didn't know how to squash the two commits without destroying this PR, so there are two commits here :sweat_smile: .

123survesh: I found out how to squash the commits. It didnt destroy the PR/

photonstorm:
👍

PR #5075, opened 2020-04-04
Add Phaser.Math.Angle.Random() and Phaser.Math.Angle.RandomDegrees()

samme:
This PR

* Adds a new feature

### DIFF
--- a/src/math/angle/Random.js
+++ b/src/math/angle/Random.js
@@ -1,0 +1,17 @@
+var FloatBetween = require('../FloatBetween');
+var CONST = require('../const');
+
+/**
+ * Returns a random angle in the range [0, 2pi].
+ *
+ * @function Phaser.Math.Angle.Random
+ * @since 3.23.0
+ *
+ * @return {number} The angle, in radians.
+ */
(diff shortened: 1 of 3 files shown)


KingCosmic:
there's alread a angle generator on Phaser.Math.RandomNumberGenerator.angle()

is there anything this does differently / couldn't just be some options for that function

rexrainbow:
@KingCosmic They are different, RandomNumberGenerator is a [3rd party random number generator](https://github.com/photonstorm/phaser/blob/master/src/math/random-data-generator/RandomDataGenerator.js), with a *seed* parameter. New methods Angle.Random, and Angle.RandomDegrees uses[ built-in *random* method](https://github.com/photonstorm/phaser/blob/master/src/math/FloatBetween.js#L20), that would have better performance ideally. (But built-in random method can't assign seed)

samme:
Should the range be [-π, π] instead of [0, 2π]? Opinions?

photonstorm:
Merged and updated so the range is -180,180, as that is what Phaser uses internally.

PR #5082, opened 2020-04-08
#5078

### LINKED ISSUE #5078: re-using a Scene and passing new Data
I am using transitions to switch between Scenes.
My Scenes are all more or less state-less and I am passing the contextual data in the transition call:

```
this.scene.transition({
      target: 'ExampleScene',
      data: contextualDataForTheScene,
      sleep: true,
      duration: 400
 });
```

This works fine when the Scene is first used, but the Second time it won't get the new Data, since the ScenePlugin calls the `Phaser.Scenes.Systems.wake` method without the Data from the transition:
https://github.com/photonstorm/phaser/blob/b46776a2c0a623bf2078b533cf29d02d24663fc0/src/scene/ScenePlugin.js#L322

Is there a reason for this? Shouldn't I be re-using Scenes that way?

MrMadClown:
This PR fixes #5078 

* Updates the Documentation
* Adds a new feature /  Fixes a bug (not sure)

- passing the data from the transition to the System.wake method
- added missing property to the SceneTransitionConfig definition

### DIFF
--- a/src/scene/ScenePlugin.js
+++ b/src/scene/ScenePlugin.js
@@ -319,7 +319,7 @@ var ScenePlugin = new Class({
 
         if (target.sys.isSleeping())
         {
-            target.sys.wake();
+            target.sys.wake(GetFastValue(config, 'data'));
         }
         else
         {
--- a/src/scene/typedefs/SceneTransitionConfig.js
+++ b/src/scene/typedefs/SceneTransitionConfig.js
@@ -11,5 +11,5 @@
  * @property {boolean} [moveBelow] - Move the target Scene to be below this one before the transition starts.
  * @property {function} [onUpdate] - This callback is invoked every frame for the duration of the transition.
  * @property {any} [onUpdateScope] - The context in which the callback is invoked.
- * @property {any} [data] - An object containing any data you wish to be passed to the target Scenes init / create methods.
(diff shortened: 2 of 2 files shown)


MrMadClown:
I was unsure how to document to "new" behaviour in the [SceneTransitionConfig](https://github.com/photonstorm/phaser/blob/7fbe57c2ddb42d3ad1966de13737cce11b95e7c4/src/scene/typedefs/SceneTransitionConfig.js#L13).
But while looking at it I noticed that the `remove` property was missing.

samme:
This looks fine. `GetFastValue()` might be unnecessary.

samme:
You could describe it this way:

> An object containing any data you wish to be passed to the target scene's init / create methods (if `sleep` is false) or to the target scene's wake event callback (if `sleep` is true).

MrMadClown:
There is no direkt access to the `config` object in the transition method,  it's always through `GetFastValue`, so I figured that's the way to go.
Would you omit the usage of `GetFastValue` if there is no `defaultValue` given?

samme:
I see it's read the same way here, so I think you're right, keep it.

https://github.com/photonstorm/phaser/blob/6e302931424618b55e4d10d60b70764673b78047/src/scene/ScenePlugin.js#L326

MrMadClown:
Nice, so are we good to merge it then?

samme:
We await the merge fairy.

photonstorm:
👍

PR #5093, opened 2020-04-17
Add PathFollower#pathDelta

samme:
This PR

* Adds a new feature

It's useful to know which direction the follower is traveling, but it's a bit of a pain to [calculate yourself](https://codepen.io/samme/pen/YzPEqYd 'Arcade physics path follower with path correction'), so here is a new property.


photonstorm:
👍

PR #5099, opened 2020-04-19
Fix camera's fadeIn/fadeOut effects

TroKEMp:
This PR

* Fixes a bug

If any object is using the Light2D pipeline (or any other I think), it will not be affected by camera effects.
Set the `TextureTintPipeline` pipeline before rendering camera effects.


photonstorm:
👍

PR #5106, opened 2020-04-22
Fix destroying NoAudioSound and add missing methods

samme:
This PR

* Updates the Documentation
* Fixes a bug

Calling NoAudioSound#destroy() would error (Maximum call stack size exceeded).

NoAudioSound was missing 6 setter methods of HTML5AudioSound/WebAudioSound.


photonstorm:
👍

PR #5107, opened 2020-04-23
Fix TypeError when loading HTML5AudioFile

samme:
This PR

* Fixes a bug

HTML5AudioFile#load would error in strict mode.


photonstorm:
👍

PR #5108, opened 2020-04-23
make config.canvasStyle more compatible

littlee:
assigning "style.cssText" instead of "style"


photonstorm:
Can you test if this works in Internet Explorer, please, as MDN is unclear about it: https://developer.mozilla.org/en-US/docs/Web/API/CSSRule/cssText

photonstorm:
Actually, it's unclear about lots of browsers, so I can't risk merging this as-is. Maybe detect if the property exists first, and only if it does, set it? Otherwise fallback to what it does currently. Thanks.

PR #5111, opened 2020-04-27
Fixes to JSON texture parsers

veleek:
This PR

* Adds a new feature
* Fixes a bug

Describe the changes below:

JSONHash doesn't load custom pivot information but JSONArray does.  Duplicating that functionality for JSONHash.
JSONArray documentation states the format should match that returned by Texture Packer, but texture packer outputs a `pivot` property when using JSONHash or JSONArray (available in the free version of Texture Packer), but uses `anchor` with the Phaser texture data object.  This updates JSONHash (and adds to JSONArray) the ability to read from `anchor` or `pivot` if that's not available.


photonstorm:
👍

PR #5113, opened 2020-04-27
Add browser entrypoint for tooling, bundlers, CDNs

FredKSchott:

* Fixes a bug

Describe the changes below:

Follow up from this conversation: https://github.com/sveltejs/svelte-repl/issues/108#issuecomment-620234154

Tooling looks at a package manifest to find a package's best entrypoint. Because you are already shipping a pre-build UMD distribution for browsers, it would be great if you could tell bundlers and other tools about it so that they can use it as needed.


photonstorm:
👍

PR #5127, opened 2020-05-02
Fix TimeStep#wake() when running

samme:
This PR

* Fixes a bug?

Currently, if you call TimeStep#wake() while the loop is running it calls TimeStep#sleep() and then continues waking, which produces an extra loop step.

I changed it so that if you call TimeStep#wake() while the loop is running, nothing happens.


photonstorm:
👍

PR #5129, opened 2020-05-03
Update Container.js

blopa:
Get first children with available getBounds()

Updates the Documentation
Adds a new feature
Fixes a bug
Describe the changes below:

Please excuse my ignorance, but shouldn't the getBounds() on the container be like this? I just faced this problem when my first child was actually a Graphic, which does not have a function getBounds(), so instead of getting the bounds from the first child, I'm trying to find the first one with bounds.


photonstorm:
I can't merge this PR because it uses ES6 and won't build, however, the approach is correct, so I have implemented it myself in the codebase and attributed it to you in the change log.

PR #5131, opened 2020-05-03
Set file states FILE_LOADING and FILE_LOADED

samme:
This PR

* Fixes a bug

[Phaser.Loader.File.html#state](https://photonstorm.github.io/phaser3-docs/Phaser.Loader.File.html#state) never showed states `FILE_LOADING` (while loading) or `FILE_LOADED` (after loading). That didn't cause any problems, but it seemed incorrect, so I added those states.

### DIFF
--- a/src/loader/File.js
+++ b/src/loader/File.js
@@ -272,6 +272,8 @@ var File = new Class({
         }
         else
         {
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

PR #5134, opened 2020-05-05
Arcade Physics collisions optimizations

samme:
This PR

* Adds a new feature

1. When colliding physics groups with the search tree enabled, there was an unnecessary intersection test for each body returned by the search.

2. When doing an overlap collision, there was an unnecessary intersection test for each pair of overlapping bodies.


photonstorm:
👍

PR #5135, opened 2020-05-05
Always use static tree for collision checks

samme:
This PR

* Adds a new feature

Sprite vs. static group collision tests always use the static tree.


photonstorm:
👍

PR #5139, opened 2020-05-08
Print texture key in warning for zero frames

samme:
This PR

* Adds a new feature

You couldn't tell which texture it was warning about.


photonstorm:
👍

PR #5142, opened 2020-05-12
Spine official runtime bug

jcyuan:

* Fixes a bug

Describe the changes below:

---------------------------

see this:
https://github.com/EsotericSoftware/spine-runtimes/pull/1672
I found this bug when I tried to implement the dyanmic skin switching.

because this is not a bug phaser should maintain, so this PR is I don't kow, reminder? 
so you can merge in advance or wait for official new version and close this PR, up to you.   ;)

### DIFF
--- a/plugins/spine/src/runtimes/spine-webgl.js
+++ b/plugins/spine/src/runtimes/spine-webgl.js
@@ -8017,7 +8017,7 @@ var spine;
             worldVertices[offset + 1] = offsetX * c + offsetY * d + y;
         };
         RegionAttachment.prototype.copy = function () {
-            var copy = new RegionAttachment(name);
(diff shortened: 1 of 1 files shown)


FostUK:
It's possibly worth linking to the spine runtimes via their github release:
e.g:
```
"dependencies": {
    "spine-runtimes": "git+ssh://git@github.com:EsotericSoftware/spine-runtimes.git#3.8.95",
  },
```

so it's easier to bump them in the future. At the moment I think they are just copied in and am not sure what version they are on?

Does this sound worthwhile? I could look into a PR.

photonstorm:
The only reason I didn't link to their runtimes directly is that I had to fix several bugs in them before merging with Phaser. However, this may be resolved now already, so I'll look into updating them.

PR #5147, opened 2020-05-13
KeyboardPlugin checkDown function - undefined parameter bugfix

### LINKED ISSUE #5146: KeyboardPlugin.checkDown doesn't work if optional parameter left out
https://github.com/photonstorm/phaser/blob/85873d2958ba70344c9b5efd3c642e6799af9b6f/src/input/keyboard/KeyboardPlugin.js#L629

It has the (TypeScript) signature of: `checkDown(key: Phaser.Input.Keyboard.Key, duration?: number): boolean;` and a statement that if left out, `duration?` is defaulted to 0 `@param duration The duration which must have elapsed before this Key is considered as being down. Default 0.`

However it simply doesn't work, I suspect that `duration` is left undefined and further functions don't work.

Reproduction:

```ts
// in update
    let keyboard = this.input.keyboard;

    if (keyboard.checkDown(keyboard.addKey('A'), 0)) {
      console.log('A was held down for 0ms');
    }

    if (keyboard.checkDown(keyboard.addKey('S'), 100)) {
      console.log('S was held down for 100ms');
    }

    if (keyboard.checkDown(keyboard.addKey('W'))) {
      console.log('W was held down for 0ms but I wont print');
    }
```

lozzajp:
This PR - Fixes a bug

Fixes #5146 - KeyboardPlugin.checkDown doesn't work if optional parameter left out 

Describe the changes below:

The checkDown method in question has an optional parameter of `duration`, that is said to default to 0 but doesn't. Leads to downstream problem with it being undefined and leads to a bug where trying to use the simple method with 1 parameter didn't work.

Below test case (from #5146) will now print the message:

```ts
    if (keyboard.checkDown(keyboard.addKey('W'))) {
      console.log('W was held down for 0ms but I wont print');
    }
```

### DIFF
--- a/src/input/keyboard/KeyboardPlugin.js
+++ b/src/input/keyboard/KeyboardPlugin.js
@@ -626,7 +626,7 @@ var KeyboardPlugin = new Class({
      * 
      * @return {boolean} `true` if the Key is down within the duration specified, otherwise `false`.
      */
-    checkDown: function (key, duration)
+    checkDown: function (key, duration = 0)
     {
         if (this.enabled && key.isDown)
         {
--- a/src/input/keyboard/KeyboardPlugin.js
+++ b/src/input/keyboard/KeyboardPlugin.js
@@ -626,10 +626,15 @@ var KeyboardPlugin = new Class({
      * 
      * @return {boolean} `true` if the Key is down within the duration specified, otherwise `false`.
      */
-    checkDown: function (key, duration = 0)
+    checkDown: function (key, duration)
     {
         if (this.enabled && key.isDown)
         {
+            if (duration === undefined)
+            {
+                duration = 0;
+            }
+
             var t = SnapFloor(this.time - key.timeDown, duration);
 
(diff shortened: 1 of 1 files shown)


lozzajp:
Couldn't see much in the way of code styles that prevents this, but I am quite new to JS and not sure if this default parameter style is acceptable or if it should be guarded against in the method itself :)

photonstorm:
Thanks for the PR. You can't specify default values in functions like this unless you're coding in ES6, and the v3 codebase is all ES5, so it needs to be done in the function directly:

```
checkDown: function (key, duration)
{
    if (duration === undefined) { duration = 0; }
```

lozzajp:
Updated.

lozzajp:
Look better @photonstorm?

photonstorm:
Thanks for the updated PR - it's almost right :) But don't worry, I've fixed it locally and will attribute it to your in the Change Log.

PR #5158, opened 2020-05-21
Update Docs for render config pixelArt to be more specific

mooreInteractive:
This PR Updates the Documentation

Describe the changes below:
For pixel art, we want the aliasing, so anti-aliasing should be off, or false. The current description makes it seem like pixelArt sets antialiasing to true, but I don't think that it does actually.


photonstorm:
👍

PR #5159, opened 2020-05-22
Allow chaining of multible animations

tgroborsch:
This PR

* Adds a new feature

Describe the changes below:
With this commit it is possible to chain more than one animation together for example:
this.mole.anims.play('digging',true).anims.chain('lifting').anims.chain('looking').anims.chain('lowering');

This commit introduces a property nextAnimsQueue. It holds all additional nextAnim-Keys in the order the chain()-method was called. It is handed over to the nextAnim-Property in the stop()-method.

### DIFF
--- a/src/gameobjects/components/Animation.js
+++ b/src/gameobjects/components/Animation.js
@@ -88,15 +88,6 @@ var Animation = new Class({
         this.nextAnim = null;
 
         /**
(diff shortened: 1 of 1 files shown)


photonstorm:
Thanks, I like this feature :) I had to fix some lint errors that prevented it from building, but otherwise, it looks good.

tgroborsch:
Thanks for your positive feedback! Glad I could help support phaser :-)

PR #5161, opened 2020-05-22
Fix error stopping HTML5AudioSound after page blur

samme:
This PR

* Fixes a bug

If you're using HTML5 Audio with `pauseOnBlur` (the default), play a sound, schedule stopping the sound (e.g., timer, tween complete callback), leave the page, and return to the page, sound `stop()` will error.


photonstorm:
👍

PR #5166, opened 2020-05-25
Add Group#setActive(), Group#setName()

samme:
This PR

* Adds a new feature

These are important for API consistency, as I keep trying to use them. 😉


photonstorm:
Fair enough :)

PR #5169, opened 2020-05-27
Refactor createSceneFromInstance()

samme:
Not a bug, but SceneManager#createSceneFromInstance() had a useless condition, so I removed it. Several bytes saved.


photonstorm:
👍

PR #5172, opened 2020-05-29
Add info to the scene manager method headers where optional data goes.

lozzajp:
* Updates the Documentation

Dived in to this code a bit to build up a table of what scene plugin methods do with the optional data param, and requested a table be added to Samme's already [great scene overview](https://gist.github.com/samme/01a33324a427f626254c1a4da7f9b6a3). 

Little change to some of the header docs in some scene manager methods to better reflect where the optional `data` params go. Specifically mentioning that `Scene.create` will also receive them.


photonstorm:
👍

PR #5180, opened 2020-06-02
Fix bug about setting xRadius/yRadius by width/height

rexrainbow:

* Fixes a bug

Describe the changes below:

xRadius/yRadius is half of width/height


photonstorm:
👍

PR #5183, opened 2020-06-05
Add Arcade.Body#updateFromGameObject

samme:
This PR

* Adds a new feature

This extracts the relevant code from [Body#preUpdate()](https://photonstorm.github.io/phaser3-docs/Phaser.Physics.Arcade.Body.html#preUpdate) as a new public method. It also lets [refreshBody()](https://photonstorm.github.io/phaser3-docs/Phaser.Physics.Arcade.Components.Enable.html#refreshBody) work for dynamic bodies (previously it would error).

### Rationale

After changing the body offset or sprite transform, you might want to read the body's new `position` and `center` immediately, before the next physics step.

### DIFF
--- a/src/physics/arcade/Body.js
+++ b/src/physics/arcade/Body.js
@@ -897,6 +897,30 @@ var Body = new Class({
     },
 
     /**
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

PR #5185, opened 2020-06-06
Add tween 'stop' event and 'onStop' callback

### LINKED ISSUE #5030: Tween onstop event missing
Hi, I'm using tweens to animate my UI components, and sometimes there is a need to kill tweens then start another tween, I'm using promises for tweens and I need to resolve promise if tween was killed.
`this.scene.tweens.killTweensOf` calls stop on running tweens, but there are no emitting events to handle tween killing. So if it's possible, please implement `stop` or `kill` event on tween stop.

Thanks in advance.

samme:
This PR

* Adds a new feature

Closes #5030


photonstorm:
👍

PR #5186, opened 2020-06-06
Configure timeline without tweens

samme:
This PR

* Adds a new feature

Previously, if you created a timeline and passed no `tweens` in the config, the timeline would be created but all config properties were ignored.

Now the timeline's own properties (`completeDelay`, `loop`, `loopDelay`, `useFrames`, `onStart`, `onUpdate`, `onLoop`, `onYoyo`, `onComplete`, etc.) are set from the config.


photonstorm:
👍

Issue #3673, opened 2018-05-18
Mask behaviour when there are nested containers. 


kanthi0802:
Hi,

We have a component kind of List. It basically extends Container and creates children which are also of type Containers. Each of these children contain a background image and  some text objects to display data, they are aligned vertically in the parent list container. Now the issue is when we are trying to apply mask to the parent container , the text children in second level containers are not being rendered at all. Only the background image is rendered.

Further more input events seems to be working on the game objects which are masked .

RollinSafary:
even if you have sprite in container mask isn't working

```
var config = {
    type: Phaser.WEBGL,
    parent: 'phaser-example',
    width: 640,
    height: 480,
    scene: {
        preload: preload,
        create: create,
        update: update
    }
};

var loadBar;
var t = 0.0;
var text;
var game = new Phaser.Game(config);

function preload() {

    this.load.image('swirl', 'assets/pics/color-wheel-swirl.png');
    this.load.image('checker', 'assets/pics/checker.png');
    this.load.bitmapFont('gothic', 'assets/fonts/bitmap/gothic.png', 'assets/fonts/bitmap/gothic.xml');
}

function create() {
    var container = this.add.container()
    loadBar = this.add.graphics();

	var checker = this.make.image({
		x: game.config.width / 2,
		y: game.config.height / 2,
		key: 'checker',
		add: true
	});


    var swirl = this.make.sprite({
    	x: game.config.width / 2, 
    	y: game.config.height / 2, 
    	key: 'swirl',
    	add: true
    });
    container.add(swirl)

    loadBar.x = game.config.width / 2;
    loadBar.y = game.config.height / 2;

    swirl.mask = new Phaser.Display.Masks.BitmapMask(this, loadBar);
    text = this.add.dynamicBitmapText(game.config.width / 2 - 20, game.config.height / 2 - 15, 'gothic', '0%', 32);

}

function update()
{
    var step = Math.abs(Math.sin(t)) * 400;

    loadBar.clear();
    loadBar.lineStyle(40, 0, 1);
    loadBar.beginPath();
    loadBar.arc(0, 0, 100, 0, Phaser.Math.DegToRad(-step), false);
    loadBar.strokePath();
    loadBar.closePath();

    t += 0.01;

    text.setText((step / 400 * 100).toFixed(0) + '%')
}

```

photonstorm:
From the docs:

```
 * Containers can have masks set on them and can be used as a mask too. However, Container children cannot be masked.
[...]

PR #5193, opened 2020-06-17
Fix nested mask behaviour for Phaser.CANVAS mode

scott20145:
This PR
* Fixes a bug

Describe the changes below:

in #3673, The nested mask only works properly in `Phaser.WEBGL`, this PR fix it for `Phaser.CANVAS`.


photonstorm:
Had to fix some lint errors, but otherwise looks good, thanks!

PR #5194, opened 2020-06-17
Update TextStyle to have jsdocs for word wrap related properties

mikewesthad:
This PR:
* Updates the Documentation

Describe the changes below:
I updated TextStyle to have jsdoc tags for: wordWrapWidth, wordWrapCallback, wordWrapCallbackScope, wordWrapUseAdvanced. These properties are already there, they just aren't recognized in the docs or in the types. The rationale to expose them in the docs/types is that these are useful to have access to when running your own custom wrapping. E.g. you can read the wordWrapWidth in your wordWrapCallback so that your Text can change width without needing to apply a new callback - all you need to do is change the wordWrapWidth (something I was doing in a project). I tested the jsdocs tags by building the doc repo.

### DIFF
--- a/src/gameobjects/text/TextStyle.js
+++ b/src/gameobjects/text/TextStyle.js
@@ -295,6 +295,57 @@ var TextStyle = new Class({
         this.baselineY;
 
         /**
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

PR #5197, opened 2020-06-23
Fix startFullscreen() failure in Safari

### LINKED ISSUE #5143: phaser 3 `startFullscreen()` not works on safari.
## Version

* Phaser Version:3.22

* Operating system: Macos

* Browser: Safari

## Description
hi @photonstorm 

I am using `game.scale.startFullscreen()` method.  Problem that appears is it doesn't work on safari macos.  of course, it also doesn't work on ios safari. but it works correctly on other browsers.

On safari, it throws `fullscreenfailed` event.....

I've noticed that phaser uses prefixes.  but still, what could be the reason ? 

```js
<div id="video">
    
</div>
<button onclick="enablefullscreen()">Enable</button>
<script>
function enablefullscreen(){
    toggleFullscreen();
}
function toggleFullscreen() {
  let elem = document.getElementById("video");
  elem.webkitRequestFullscreen()
}
</script>
```

I tried this code and this worked on macos safari as a fullscreen pretty well. So I sum up that fullscreen works fine in safari, but why doesn't phaser's `startFullscreen()` work even though it uses prefixes? 

Any idea... ?

samme:
This PR

* Fixes a bug

Fixes #5143

### DIFF
--- a/src/scale/ScaleManager.js
+++ b/src/scale/ScaleManager.js
@@ -1229,28 +1229,13 @@ var ScaleManager = new Class({
         {
             var fsTarget = this.getFullscreenTarget();
 
-            var fsPromise;
-            
             if (fullscreen.keyboard)
             {
-                fsPromise = fsTarget[fullscreen.request](Element.ALLOW_KEYBOARD_INPUT);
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

PR #5206, opened 2020-07-03
Add Phaser.Math#RotateTo

samme:
This PR

* Adds a new feature

> Position a `point` at the given `angle` and `distance` to (`x`, `y`).

This is a common operation. I don't think `RotateTo` is the best name but I can't think of a better one.


photonstorm:
👍

PR #5213, opened 2020-07-08
Fix setCrop() for Matter.Image and Matter.Sprite

### LINKED ISSUE #5211: Phaser.Physics.Matter. Sprite.setCrop throws TypeError
## Version: 3.x - 3.23 (I can't find a version this works on).

* Phaser Version:

* Operating system: Linux/Windows

* Browser: N/A. Electron, Chrome 83

## Description

Expected Behaviour: `Phaser.Physics.Matter. Sprite` gets cropped using the `Phaser.Physics.Matter. Sprite.setCrop` method. Line 19404 of Phaser.js has the property `ui` of the variable class property `crop` expecting a re-assignment of it's current value and assumes a previous value has been set by the class method `this.resetCropObject();` as stipulated in the "constructors" (class methods who share the same name as the class itself).

Actual Behaviour: The `crop` property was never set to a default value and so attempting to access the child property (EG: crop.u0 causes it to throw an undefined property `TypeError`.

## Example Test Code
https://codepen.io/MattRorke/pen/wvMmQWw

## Additional Information
**Fix**: I've managed to fix it by adding `this._crop = this.resetCropObject();` to line 99 of .\phaser-master\src\physics\matter-js\MatterSprite.js
I'm not sure if there is a better way to do this by inheritance, but it at least fixed the issue for me, for now.

samme:
This PR 

* Fixes a bug

Fixes #5211

### DIFF
--- a/src/physics/matter-js/MatterImage.js
+++ b/src/physics/matter-js/MatterImage.js
@@ -86,6 +86,16 @@ var MatterImage = new Class({
     {
         GameObject.call(this, world.scene, 'Image');
 
+        /**
+         * The internal crop data object, as used by `setCrop` and passed to the `Frame.setCropUVs` method.
+         *
+         * @name Phaser.Physics.Matter.Image#_crop
+         * @type {object}
+         * @private
+         * @since 3.24.0
+         */
+        this._crop = this.resetCropObject();
+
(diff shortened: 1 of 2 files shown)


photonstorm:
👍

PR #5214, opened 2020-07-10
Update ContainerFactory.js

SirJosh3917:
Given that Container's constructor allows x and y to be optional, the JSDoc should match that.

* Updates the Documentation

Describe the changes below:

Given that [the constructor for Container allows the x and y to be optional](https://github.com/photonstorm/phaser/blob/v3.22.0/src/gameobjects/container/Container.js#L70), the JSDoc for a GameObjectFactory should match that. This PR fixes this issue.


photonstorm:
y is actually set to x if undefined, but it doesn't matter, this will do for now!

PR #5215, opened 2020-07-10
StaticTilemapLayerWebGLRenderer: apply rotation and parent transform

cruzdanilo:
This PR
* Fixes a bug

Describe the changes below:
`StaticTilemapLayerWebGLRenderer` was not applying layer rotation and parent transform. 
`StaticTilemapLayerCanvasRenderer` was already doing this and i followed the same style.


photonstorm:
👍

photonstorm:
Had to revert this PR as sadly it breaks rendering tilemaps when the camera zoom is any value other than 1.

PR #5216, opened 2020-07-10
Fix Path moveTo optional y param

SanderVanhove:
This PR:

* Fixes the `moveTo` method of the `Path` class.

Describe the changes below:

Make the `y` parameter optional in typescript typing in case the `x` parameter is a `Phaser.Math.Vector2` object.


photonstorm:
👍

Issue #4755, opened 2019-09-14
Drag problem when game object in transformed camera or container.


ganlvtech:
Drag problem when game object in transformed camera or container.

## Version

* Phaser Version: v3.19.0

* Operating system: Windows 10 OS Version 1903 (Build 18362.356)

* Browser: Chrome 64bit 77.0.3865.75

## Description

Drag problem when game object transformed. Scaling or rotating a container, zooming or rotating a camera will cause this bug.

![01](https://user-images.githubusercontent.com/8651285/64908272-71410780-d730-11e9-9502-116c5c304e2b.gif)

Additionally, when camera scrolled during dragging, the game object won't be transformed. [input/dragging/camera move and rotate.js](https://labs.phaser.io/view.html?src=src/input/dragging/camera%20move%20and%20rotate.js)

![02](https://user-images.githubusercontent.com/8651285/64908671-e6fba200-d735-11e9-898f-a44967c7ad30.gif)

## Example Test Code

This code can be run directly at <http://labs.phaser.io/edit.html>

### Camera

```javascript
var config = {
    scene: {
        create: create
    }
};

var game = new Phaser.Game(config);

function create () {
    var rect = new Phaser.GameObjects.Rectangle(this, 100, 100, 100, 100, 0x00ff00);
    this.children.add(rect);
    rect.setInteractive({
        draggable: true
    });
    rect.on('drag', (pointer, dragX, dragY) => {
        rect.x = dragX;
        rect.y = dragY;
    });
    this.cameras.main.setOrigin(0, 0);
    this.cameras.main.setZoom(2);
    // this.cameras.main.setRotation(0.5);
}
```

### Container

```javascript
var config = {
    scene: {
        create: create
    }
};

var game = new Phaser.Game(config);

function create () {
    var rect = new Phaser.GameObjects.Rectangle(this, 100, 100, 100, 100, 0x00ff00);
    rect.setInteractive({
        draggable: true
    });
    rect.on('drag', (pointer, dragX, dragY) => {
        rect.x = dragX;
        rect.y = dragY;
    });
    var container = this.add.container(0, 0);
    container.add(rect);
    container.setScale(2);
    // container.setRotation(0.5);
}
```

## Additional Information

Maybe related to #1542

Related doc: [Phaser.Input.Events.html#event:GAMEOBJECT_DRAG](https://photonstorm.github.io/phaser3-docs/Phaser.Input.Events.html#event:GAMEOBJECT_DRAG__anchor)

[...]

PR #5218, opened 2020-07-13
Fix drag coordinates with camera zoom (issue #4755)

braindx:
This PR

* Fixes a bug

This fix addresses GitHub issue #4755 

The dragX and dragY passed to 'dragStart' and 'drag' events are supported to be in world coordinates (as specified by the docs) but the coordinates used (pointer.x and pointer.y) were not transformed by the camera into world space. The drags now use pointer.worldX and pointer.worldY which are the post-transform coordinates.


photonstorm:
👍

PR #5220, opened 2020-07-14
Set Phaser.Loader.File.url type to string

zaniar:
This PR

* Updates the Documentation
* Fixes a bug

Description:

`Phaser.Loader.File.url` currently (3.24.0) has type of `function|object|string`.
Previously (3.23.0) it was `string`. It breaks my project and I think the previous one is the correct one.


photonstorm:
I can't see why it has `function` included in the tuple, but it definitely should have `object`. A File URL is absolutely not _just_ a string, it has been incorrect up until now.

PR #5222, opened 2020-07-14
fix for GetPoints() not returning evenly spaced points along line

dsabatier:

* Fixes a bug

Describe the changes below:

This change corrects a bug in GetPoints that calculated incorrect spacing along a line:

before:
image

after:
image

codepen demonstrating the problem:
https://codepen.io/snowbillr/pen/gOPdmvN?editors=0010

### DIFF
--- a/src/geom/line/GetPoints.js
+++ b/src/geom/line/GetPoints.js
@@ -45,7 +45,7 @@ var GetPoints = function (line, quantity, stepRate, out)
 
     for (var i = 0; i < quantity; i++)
     {
-        var position = i / quantity;
+        var position = quantity > 1 ? i / (quantity-1) : 0;
 
         var x = x1 + (x2 - x1) * position;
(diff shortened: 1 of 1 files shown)


samme:
I'm not sure it's a bug, because it's consistent with the other GetPoints() methods. I can see why it looks "wrong" though.

dsabatier:
> I'm not sure it's a bug, because it's consistent with the other GetPoints() methods. I can see why it looks "wrong" though.

that's not a bad point, perhaps it is set up the way it is intentionally.  What about an option to evenly space the points rather than that being its default implementation?

photonstorm:
Having looked at this I'm afraid I can't merge it as-is. It now breaks if you don't give a quantity, but give a `stepRate` and it also breaks when Lines are used as part of a chain, i.e. in a Polygon, when you don't want a point at the end of the line. Perhaps this needs fixing via an optional parameter instead?

PR #5228, opened 2020-07-18
[Fix] common misspelling errors

Necmttn:
For reference: https://en.wikipedia.org/wiki/Wikipedia:Lists_of_common_misspellings/For_machines


photonstorm:
Virtually all of these mistakes use correct British English. The only actual spelling mistakes are in internal comments, not the API docs, so I'm not too worried about them. I don't mind merging some of these changes in, but not all of them, sorry.

PR #5235, opened 2020-07-24
[types] allow Text#setPadding to receive an object

mk360:
This PR
* Fixes a bug

Previously, when using TypeScript, `Text#setPadding` only allowed 4 numbers as arguments, despite the docs stating you could also pass an object as a first and only argument.

### DIFF
--- a/src/gameobjects/text/static/Text.js
+++ b/src/gameobjects/text/static/Text.js
@@ -1022,9 +1022,9 @@ var Text = new Class({
      * @since 3.0.0
      *
      * @param {(number|Phaser.Types.GameObjects.Text.TextPadding)} left - The left padding value, or a padding config object.
-     * @param {number} top - The top padding value.
-     * @param {number} right - The right padding value.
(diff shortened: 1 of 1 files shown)


src/gameobjects/text/static/Text.js
```diff
@@ -1022,9 +1022,9 @@ var Text = new Class({
      * @since 3.0.0
      *
      * @param {(number|Phaser.Types.GameObjects.Text.TextPadding)} left - The left padding value, or a padding config object.
-     * @param {number} top - The top padding value.
-     * @param {number} right - The right padding value.
-     * @param {number} bottom - The bottom padding value.
+     * @param {[number]} top - The top padding value.
```
photonstorm, line 1025: It needs to be like this: `@param {number} [top] - The top padding value.` if the parameter is optional. Same for the other values.

photonstorm:
Thanks for your PR! Please update to use the correct jsdoc format and I'll happily merge it.

photonstorm:
👍

PR #5242, opened 2020-07-29
Fix collisions for objects moved by friction

### LINKED ISSUE #5210: Dynamic bodies no longer hit walls when riding an horizontally moving platform --- ok until 3.9, broken from 3.10
## Version
* Phaser Version: ok until 3.9, broken from 3.10 to latest (3.23)
* Operating system: tested on OSX, Win10, iOS and Android
* Browser: Chrome, Safari (both desktop + mobile)

## Description

Consider a 2D platformer game. The characters can ride moving platforms (up/down/left/right). When riding a platform that moves horizontally, the riding body should hit any wall that would be in the way.

As per below Codepen sandboxes, I tested with all minor Phaser 3 versions:
- it works well until 3.9 included
- it is broken from 3.10 to latest 3.23

About the moving platforms:
- they move thanks to velocity (and NOT by being re-positioned every time step).
- they don't allowGravity and are immovable.
- Their frictionX is 1 to fully transfer motion to riding bodies.

## Example Test Code

The following two Codepen workspaces run the same sandbox, only the Phaser version differs:
- [it works as we'd expect until 3.9](https://codepen.io/jemreactor/pen/OJMOwox)
- [it's broken from 3.10 to 3.23](https://codepen.io/jemreactor/pen/VwerBWL)

NB. cursor keys are enabled. Running towards the wall still triggers the collision after 3.10. Only "passive" movement due to transportation is failing.

## Additional Information
- I uploaded a [quick screencast](https://youtu.be/UQQ2Yqa_-5w) to make it easier to discover the problem.
- The topic has been [discussed on the forum](https://phaser.discourse.group/t/bug-regression-riding-a-moving-body-ignore-collisions/6827/7) before  being reported here. It also documents my workaround until a clean solution is applied.
- Thanks for your attention!

samme:
This PR

* Fixes a bug

Fixes #5210

### DIFF
--- a/src/physics/arcade/SeparateX.js
+++ b/src/physics/arcade/SeparateX.js
@@ -64,6 +64,7 @@ var SeparateX = function (body1, body2, overlapOnly, bias)
         if (body2.moves)
         {
             body1.y += (body2.y - body2.prev.y) * body2.friction.y;
+            body1._dy = body1.y - body1.prev.y;
         }
     }
     else
@@ -75,6 +76,7 @@ var SeparateX = function (body1, body2, overlapOnly, bias)
         if (body1.moves)
         {
             body2.y += (body1.y - body1.prev.y) * body1.friction.y;
+            body2._dy = body2.y - body2.prev.y;
         }
     }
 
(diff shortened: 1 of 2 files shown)


Dercetech:
Thanks Samme!

photonstorm:

https://phaser.discourse.group/t/bug-regression-riding-a-moving-body-ignore-collisions/6827/11

photonstorm:
👍

PR #5244, opened 2020-08-01
Fix types for Text, TextFactory and TextCreator

16patsle:

* Updates the Documentation

Describe the changes below:
The TextCreator was lacking types for its config parameter, so I added a TextConfig type. The TextFactory typed its style parameter as 'object', I changed it to TextStyle. Lastly, I changed the type of Text's padding property to TextPadding, as that type is used other places (like in TextStyle, which can be used to set that padding property).

Let me know if I missed something with regards to TextConfig, I tried typing it similar to how it was done with for instance TileSprite and TileSpriteConfig, based on what object properties TextCreator seemed to use.


photonstorm:
Looks good to me 👍

PR #5250, opened 2020-08-03
Rename all setInteractive() arguments, and docs

### LINKED ISSUE #5122: .setInteractive() parameters names are misleading
## Description

the api of .setInteractive(): https://github.com/photonstorm/phaser/blob/d6e860076637c1686adcfcd64afd40c829b5d181/src/gameobjects/GameObject.js#L454
calls the InputConfiguration - shape
and HitAreaCallback - callback
Which is confusing and can cause mistake if you pass an object instead of shape (for example if you want to change the cursor + set hit area)

Because here:
https://github.com/photonstorm/phaser/blob/d6e860076637c1686adcfcd64afd40c829b5d181/src/input/InputPlugin.js#L2145
it seeks for ""hitArea" and "hitAreaCallback" instead.

Maybe the names should be changed or let 'setHitArea' search for shape / callback as well.

## Example Test Code

`
var sprite = scene.add.sprite(x, y, "key");
sprite.setInteractive({ 
          shape: new Phaser.Geom.Rectangle(0, 0, 50, 50), 
         callback: Phaser.Geom.Rectangle.Contains, 
         cursor: "url(assets/cursors/myCursor.cur), pointer" 
 })`

## Additional Information

If you decide you want to change it, i would be happy to add a pull request.

### LINKED ISSUE #5245: Add setInteractive bounds on graphics to documentation
If documentation related issues shouldn't be here, let me know, ill close and move this elsewhere.

The syntax solution in #4194 should be documented somewhere. Possibly in [Input:InputConfiguration](https://photonstorm.github.io/phaser3-docs/Phaser.Types.Input.html#.InputConfiguration) as Graphics links out to it. 

Having shape be documented as an `any` type with no real explanation of what data it wants, is extremely unhelpful.

As well, the idea of the callback function being used to determine whether a a given point is within a geometry, while extremely modular, is not explained very well. My first idea was to simple pass an array of coordinates, as its similar to the way [this example](https://phaser.io/examples/v3/view/input/game-object/change-sprite-hitarea-size) works.

samme:
This PR

* Updates the Documentation

Fixes #5122 

Fixes #5245

### DIFF
--- a/src/gameobjects/GameObject.js
+++ b/src/gameobjects/GameObject.js
@@ -442,18 +442,27 @@ var GameObject = new Class({
      *
      * You can also provide an Input Configuration Object as the only argument to this method.
      *
+     * @example
+     * sprite.setInteractive();
+     *
+     * @example
+     * sprite.setInteractive(new Phaser.Geom.Circle(45, 46, 45), Phaser.Geom.Circle.Contains);
+     *
+     * @example
+     * graphics.setInteractive(new Phaser.Geom.Rectangle(0, 0, 128, 128), Phaser.Geom.Rectangle.Contains);
+     *
      * @method Phaser.GameObjects.GameObject#setInteractive
      * @since 3.0.0
      *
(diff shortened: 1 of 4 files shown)


photonstorm:
👍

PR #5257, opened 2020-08-09
Fix incorrect event names in ProcessQueue.js#205, 222

halilcakar:

* Fixes a bug

Describe the changes below:

ProcessQueue.js#205 and 222, had incorrect event names.

### DIFF
--- a/src/structs/ProcessQueue.js
+++ b/src/structs/ProcessQueue.js
@@ -11,7 +11,7 @@ var Events = require('./events');
 /**
  * @classdesc
  * A Process Queue maintains three internal lists.
- * 
+ *
  * The `pending` list is a selection of items which are due to be made 'active' in the next update.
  * The `active` list is a selection of items which are considered active and should be updated.
  * The `destroy` list is a selection of items that were active and are awaiting being destroyed in the next update.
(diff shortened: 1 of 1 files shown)


halilcakar:
Hi @photonstorm,

I was looking into structs/RTree.js and saw some duplicate functions on [Lines](https://github.com/photonstorm/phaser/blob/master/src/structs/RTree.js#L204-L207) and [Lines](https://github.com/photonstorm/phaser/blob/master/src/structs/RTree.js#L469-L479).

Is there a reason for this or this was just a mistake?

halilcakar:
Hi again,

I also have a proposal about Call.js.

So originally it's just calling function with `context` and `item` so the only argument here is `item`. 
```javascript
var Call = function (items, callback, context)
{
    for (var i = 0; i < items.length; i++)
    {
        var item = items[i];

        callback.call(context, item);
    }

    return items;
};

module.exports = Call;
```

What would you think if we also pass the `index` too?

```javascript
callback.call(context, item, i);
```

Is this possible to add?

photonstorm:
Thanks for the PR. I'm not worried about the RTree code, as it all comes from an external lib anyway, so when (if) they update, it just gets replaced wholesale by their new version.

We could add the index to `Call` I guess, although I suspect most people just use a `forEach` if they need that.

Issue #5019, opened 2020-02-25
GridAlign doesn't work when only height parameter set


halilcakar:
## Version

* Phaser Version: ^3.22.0

* Operating system: Windows 10 Home Single Language | Version 1903

* Browser: Chrome

## Description

`Phaser.Actions.GridAlign` doesn't work when you only set height property as following.
```
Phaser.Actions.GridAlign(group.getChildren(), {
    height: 10,
    cellWidth: 32,
    cellHeight: 32,
    x: 100,
    y: 100
}); // this doesn't work
```
it's working when we only set width property like following:
```
Phaser.Actions.GridAlign(group.getChildren(), {
    width: 10,
    cellWidth: 32,
    cellHeight: 32,
    x: 100,
    y: 100
}); // this does work 
```

I'm not sure this is the expected behaviour.
## Example Test Code
You can test it on labs [this example](https://labs.phaser.io/edit.html?src=src\actions\grid%20align.js&v=3.22.0
). just try to remove width first and left height alone.
When you do the revers way set width and remove height it's working.

## Additional Information
I believe this should work both way. Isn't it? :) 
On the code we are already checking if there are properties (width and height) and storing them as
```
var widthSet = options.hasOwnProperty('width');
var heightSet = options.hasOwnProperty('height');
```
Is it possible to use these flag's as if we are gonna consider setting by width and/or height?

photonstorm:
Thanks for opening this issue, and for submitting a PR to fix it. We have merged your PR into the `master` branch and attributed the work to you in the [Change Log](https://github.com/photonstorm/phaser/blob/master/CHANGELOG.md). If you need to tweak the code for whatever reason please submit a new PR.

PR #5269, opened 2020-08-20
Fix for #5019

halilcakar:

* Fixes a bug

Describe the changes below:

On the issue #5019, I've asked about if we set only height, it's not doing anything. This might be a solution for that problem

```javascript
var group = this.add.group({
    key: 'diamonds',
    frame: [ 0, 1, 2, 3, 4 ],
    frameQuantity: 20
});

Phaser.Actions.GridAlign(group.getChildren(), {
    width: 10,
    height: 5,
    cellWidth: 32,
    cellHeight: 32,
    x: 100,
    y: 100
});
```
Creates the result of 
image

It stops when it hit's the height limit as futured on [line](https://github.com/photonstorm/phaser/blob/master/src/actions/GridAlign.js#L81)

How ever if you only give height as 5 lets say
```javascript
var group = this.add.group({
    key: 'diamonds',
    frame: [ 0, 1, 2, 3, 4 ],
    frameQuantity: 20
});

Phaser.Actions.GridAlign(group.getChildren(), {
    height: 5,
    cellWidth: 32,
    cellHeight: 32,
    x: 100,
    y: 100
});
```
The result shows as 
image


src/actions/GridAlign.js
```diff
@@ -65,6 +65,27 @@ var GridAlign = function (items, options)
             //  We keep laying them out vertically until we've done them all
             tempZone.y += cellHeight;
         }
+        else if (heightSet && !widthSet)
+        {
+            console.log('heighta göre set!');
```
photonstorm, line 70: This line needs removing, please.

halilcakar: I've just removed that console 👍  

Was testing out with that :)

photonstorm:
👍

Issue #5261, opened 2020-08-11
Improve performance of text game object


rexrainbow:
In phaser3, text game object uses [MeasureText method](https://github.com/photonstorm/phaser/blob/master/src/gameobjects/text/MeasureText.js) to get ascent and descent by scanning canvas's [image data](https://github.com/photonstorm/phaser/blob/master/src/gameobjects/text/MeasureText.js#L64).
In phaser4, text game object gets ascent and descent by [`metrics.actualBoundingBoxAscent`, `metrics.actualBoundingBoxDescent`](https://github.com/phaserjs/phaser/blob/master/src/gameobjects/text/Text.ts#L128-L129). This solution might have better performance.

photonstorm:
It has _much_ better performance! But I'm not willing to make this change in v3, sorry. There's enough changing in 3.5 already without rewriting how Text objects fundamentally work :) Happy to leave this as a v4 benefit!

PR #5270, opened 2020-08-21
Get ascent and descent from context.measureText method

rexrainbow:

* Improve performance

Describe the changes below:

Get ascent and descent from context.measureText method instead of scanning image data. ( #5261 )

### DIFF
--- a/src/gameobjects/text/MeasureText.js
+++ b/src/gameobjects/text/MeasureText.js
@@ -9,7 +9,7 @@ var CanvasPool = require('../../display/canvas/CanvasPool');
 /**
  * Calculates the ascent, descent and fontSize of a given font style.
  *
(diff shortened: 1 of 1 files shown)


rexrainbow:
It is fine to reject this PR for v4 benefit.

photonstorm:
On reflection, I'm happy to merge this in, but I can't have it replace the old way of doing it, as the new one doesn't work on any version of IE. So I think it'll need a conditional to check if it's supported first.

rexrainbow:
Original solution added, for IE and Firefox for Android which do not have `actualBoundingBoxAscent` and `actualBoundingBoxDescent` properties. Thanks for this advice.

photonstorm:
Thanks for the update.

PR #5274, opened 2020-08-26
Add ability to pan sounds

### LINKED ISSUE #5273: Add ability to pan the sounds when using WebAudio
I think that the ability to pan the sounds is pretty useful.
WebAudio API has a simple [StereoPannerNode](https://developer.mozilla.org/en-US/docs/Web/API/StereoPannerNode) for this.
It adds just one property `pan` that sits in the range from -1 to 1.

What do you think about sound panning feature?
I can try to implement it.

pi-kei:
This PR

* Adds a new feature

Resolves #5273

### DIFF
--- a/src/sound/BaseSound.js
+++ b/src/sound/BaseSound.js
@@ -128,7 +128,8 @@ var BaseSound = new Class({
             detune: 0,
             seek: 0,
             loop: false,
-            delay: 0
+            delay: 0,
+            pan: 0
 
         };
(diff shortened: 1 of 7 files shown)


pi-kei:
I think I'm done. I'm not sure what to put in `@since` attributes in jsdoc? Or I shouldn't bother about it?

photonstorm:
👍

PR #5275, opened 2020-08-26
Renderer#snapshotGameObject function added

halilcakar:
* Proposal to add a new function to renderers

Describe the changes below:

It add's a function called `snapshotGameObject` to WebGLRenderer.js and CanvasRenderer.js.

```javascript
    /**
     * Schedules a snapshot of given gameObject.
     * */
    snapshotGameObject: function (gameObject, callback, type, encoderOptions)
    {
        var bounds = gameObject.getBounds();

        return this.snapshotArea(bounds.x, bounds.y, bounds.width, bounds.height, callback, type, encoderOptions);
    },
```

It's only a shorthand to get to Image of a certain area actually, within the gameobject.

This is probably not the best way to create a snapshot of a gameObject,
When you think it should be without background but for temporary solution maybe?

(It is quite usefull if you just want to get an area of a Container =)

### DIFF
--- a/src/renderer/canvas/CanvasRenderer.js
+++ b/src/renderer/canvas/CanvasRenderer.js
@@ -553,6 +553,16 @@ var CanvasRenderer = new Class({
     },
 
     /**
+     * Schedules a snapshot of given gameObject.
+     * */
+    snapshotGameObject: function (gameObject, callback, type, encoderOptions)
+    {
+        var bounds = gameObject.getBounds();
+
+        return this.snapshotArea(bounds.x, bounds.y, bounds.width, bounds.height, callback, type, encoderOptions);
+    },
+
+    /**
      * Schedules a snapshot of the given area of the game viewport to be taken after the current frame is rendered.
      *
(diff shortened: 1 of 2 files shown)


photonstorm:
I like the concept, but I think the method name is misleading - as it will capture the area of the renderer, which may not even have your Game Object in it (if, for example, another Game Object has been placed over the top of it)

halilcakar:
Yea after some thinking it's really misleading 😂

Do you think is there a way to snapshot a gameObject or Container with children without background? I would very much like to have some method like this to use it.

At my work we have a concept of Malbuch(it's a drawing book kinda) and I actually made it downloadable by users but its for now downloading all canvas area which also has the colors palette and header and such..

That's actually how I came up with idea of `snapshotGameObject` function but as you noticed it's only for an area.

If you kinda give me guidance to make it happen, I would like to create a PR for this 🙏😊

photonstorm:
Yeah if you just use a `renderTexture` and draw to that, you can then use `snapshot` on it.

PR #5278, opened 2020-08-29
Reset flags when resetting an Arcade Body

samme:
This PR

* Fixes a bug


photonstorm:
Which bug does this fix?

samme:
If you call [Body#reset()](https://photonstorm.github.io/phaser3-docs/Phaser.Physics.Arcade.Body.html#reset),

- `blocked` and `touching` may be incorrect until the first physics step after reset
- `wasTouching` may be incorrect until the second physics step after reset

photonstorm:
👍

PR #5298, opened 2020-09-11
Add default value for Phaser.Core.Config#audio

samme:
This PR

* Adds a new feature?

Changes the default value from `undefined` to `{}` and refactors Phaser's use of it.

I think this is more correct (per the docs) and it simplifies later access to `game.config.audio`.


photonstorm:
Sure, makes sense to me.

PR #5300, opened 2020-09-11
Says webglRender but it is in PipelineManager

EmilSV:
* Updates the Documentation

Describe the changes below:
@name was wrong


photonstorm:
👍

PR #5301, opened 2020-09-11
bug fix: don't add path to url if it is absolute

firesoft:

* Fixes a bug

Describe the changes below:

don't add Loader.path to file url if it is absolute url
now Loader.path is added at the beginning of url even if the url is absolute, result: absolute url addresses are incorrect


photonstorm:
Thanks for spotting this! This check does take place elsewhere, but not with the `path` as you've found. I won't merge your PR because I've used the code we use in `GetURL` to perform the check instead, but thank you for reporting it - I've made sure to credit you in the Change Log.

PR #5311, opened 2020-09-16
Add ability to use custom easing function(s) with particle emitters

vforsh:
This PR

* Updates the Documentation
* Adds a new feature

Describe the changes below:
Internally EmitterOp uses `GetEaseFunction` which can handle eases as strings and as functions, and also can accept easeParams.


photonstorm:
Makes sense! Merged :)

PR #5312, opened 2020-09-16
GameObjectFactory#existing Should always add to updatelist if have preUpdate function

EmilSV:
* Fixes a bug

Describe the changes below:
"Phaser.GameObjects.GameObjectFactory#existing" 
could only add to either the display list or update list not both. Now it can add to both again


rexrainbow:
New displyList.add will fire event to add game object to updateList, thus it dose not add again. 
Only Group game object dose not have to add to disoplayList, it will be added to updateList through else if condition.

EmilSV:
Ok did ses it in the changelog did not think it would break my custom GameObject classes.

PR #5316, opened 2020-09-18
Allow setting alpha for 2D camera flash effect

kainage:
Adds a new feature:

Allow the overriding of `this.alpha` value for the camera flash effect. 

Previously the alpha value was always set to 1 in the update loop. This allows the user to set the `alpha` value on the camera effect before running it.

Ideally this would be a value you can pass in to the `start` function, but if that were the case I feel like it should go after the `blue` argument. I didn't know if this change warrants the breaking of any previous usage of this function, so I left it at this.


photonstorm:
👍

PR #5317, opened 2020-09-18
Docs: correct CursorKeys

samme:
This PR

* Updates the Documentation

Its properties (`left`, `right`, etc.) are never undefined.


photonstorm:
👍

PR #5326, opened 2020-09-25
Remove Arcade.Body#updateCenter() in Arcade.Body#setOffset()

samme:
Not really a bug, but `updateCenter()` there was unnecessary because `position` is still unchanged.


photonstorm:
👍

PR #5330, opened 2020-09-26
Fix drawing Group to RenderTexture

somnolik:

* Fixes a bug

Describe the changes below:

Group children (generic GameObjects) expect a 'camera' argument
in their 'willRender' function; that was missing in RenderTexture.batchGroup().


photonstorm:
👍

PR #5334, opened 2020-09-28
Correct Arcade Body types

### LINKED ISSUE #5328: Body.onWorldBounds is read-only and cannot be set
## Version

* Phaser Version: 3.24.1

* Operating system: Windows 10 Pro

* Browser: Not browser specific

## Description

**Using Typescript**

Cannot set Physics body 'onWorldBounds' property as it is read-only.

I expected to be able to use `something.body.onWorldBounds = true` since that's what is shown in every example I find to enable worldbounds collision event.

## Example Test Code

```typescript
    const bomb = this.physics.add.image(400, 200, 'bomb')
    bomb.setCollideWorldBounds(true)
    bomb.body.onWorldBounds = true // LINT ERROR: Cannot assign to 'onWorldBounds' because it is a read-only property. ts(2540)
    bomb.setBounce(1)
    bomb.setVelocity(200, 20)
```

## Additional Information

image

How come everywhere it is specified to be used to set a flag but on the source code is indeed read-only ?
Should always work wonders under Javascript.

samme:
This PR

* Updates the Documentation

Fixes #5328

### DIFF
--- a/src/physics/arcade/typedefs/GameObjectWithBody.js
+++ b/src/physics/arcade/typedefs/GameObjectWithBody.js
@@ -1,5 +1,6 @@
 /**
- * @typedef {Phaser.GameObjects.GameObject} Phaser.Types.Physics.Arcade.GameObjectWithBody
+ * @typedef {object} Phaser.Types.Physics.Arcade.GameObjectWithBody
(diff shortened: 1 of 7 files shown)


photonstorm:
👍

PR #5340, opened 2020-10-03
Add more specific 'object' types, including GO Configs

16patsle:
This PR

* Updates the Documentation

Describe the changes below:
When I improved the type for TextCreator's config parameter earlier, I noted that some other creators had similar issues, which caused me to look into other places where `object` is used as a type, specifically in parameters and return types.

I'm not entirely done, I plan on doing a bit more, but so far I think I have covered most of the game object creators. Currently I set the config parameter in ImageCreator and BlitterCreator to accept a SpriteConfig object, since they accept the same config as a sprite. Would it be better to have the types separate, or should SpriteConfig be renamed perhaps?

Also, I noticed a couple of GOs like DOMElement and the shape objects don't seem to have a Creator. Was there any specific reason for this, or just that they weren't really needed? I could probably create them if you want.


photonstorm:
Thanks for this - I'm happy to merge it as-is, but could you please fix the merge conflicts first? Looks like some defs for Game Objects no longer part of 3.50 (like Quad).

16patsle:
Sure! I'll have a look at what more I was thinking of doing, and fixing the merge conflicts. Will probably be a week until I have time though, if I don't forget.

16patsle:
I think everything should be in order now. Completed the ParticleEmitterManagerConfig I had left unfinished (and uncommitted), and added a config typedef for Mesh too.

photonstorm:
👍

PR #5202, opened 2020-07-01
separateCircle fixes

samme:
This PR

* Fixes a bug

Fixes #4770 (in part). 

1. Momentum exchange didn't work correctly vs. immovable bodies. The movable body tended to stop.
2. Body mass was decreasing inertia instead of increasing it.
3. The separation vector seemed to be incorrect, causing the slip/slide collisions (#4770). The separation is now correct for circle–circle collisions. It's not correct for circle–rectangle collisions, but may still be good enough when combined with velocity.

PR #5344, opened 2020-10-05
Correct Arcade Body center after separations

samme:
This PR

* Fixes a bug

[Arcade.Body#center](https://photonstorm.github.io/phaser3-docs/Phaser.Physics.Arcade.Body.html#center) values were incorrect after collisions with the world bounds or (for rectangular bodies) after collisions with another body. I fixed that by updating the body center after those separations.

I also removed updating the body center **before** separating circles (#5202) because it should be unnecessary now.

### DIFF
--- a/src/physics/arcade/Body.js
+++ b/src/physics/arcade/Body.js
@@ -1231,6 +1231,7 @@ var Body = new Class({
         if (wasSet)
         {
             this.blocked.none = false;
+            this.updateCenter();
         }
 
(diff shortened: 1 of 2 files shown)


photonstorm:
Does this fix a 3.24 issue? If so I'll add it to the Change Log.

samme:
Issue goes back to v3.0.0, but it's kind of obscure.

PR #5350, opened 2020-10-09
Add delta argument to WORLD_STEP event

samme:
This PR

* Adds a new feature

In Phaser.Physics.Arcade.Events#WORLD_STEP, the `delta` argument is the physics step delta in seconds.


photonstorm:
👍

PR #5374, opened 2020-10-24
Added loadMediaStream method

pirateksh:

* Adds a new feature

Describe the changes below:

* Added a new method **_loadMediaStream_** which takes parameter stream, loadEvent, and noAudio. Here, the **_stream_** is the **_MediaStream_** Object.
* This will allow us to load the video directly from the **_MediaStream_** object **instead of using a URL (like loadURL method)** thus will be helpful to load a video being **streamed from a webcam**.
* As, **_window.URL.createObjectURL()_** has been deprecated, so, instead of using **_video.src = window.URL.createObjectURL(stream)_**, this method uses **_video.srcObject = stream_**.

### DIFF
--- a/src/gameobjects/video/Video.js
+++ b/src/gameobjects/video/Video.js
@@ -857,6 +857,66 @@ var Video = new Class({
 
         return this;
     },
+    
+    /**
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

PR #5388, opened 2020-11-07
Scale damping by the physics step delta

### LINKED ISSUE #5356: Arcade Physics damping is framerate dependent
E.g., when `useDamping` is true, a drag value of `0.99` over 1 second reduces velocity to ~73% at 30fps or ~54% at 60fps.

This is a bit of a weakness since all other Arcade Physics motion is framerate independent.

Ideally I guess the damping factor would be a 1-second value like the others, but that would be a big change. It could be reinterpreted as a 1/60-second value and scaled by the time delta.

samme:
This PR

* Adds a new feature

This redefines [Arcade.Body#drag](https://photonstorm.github.io/phaser3-docs/Phaser.Physics.Arcade.Body.html#drag) when [damping](https://photonstorm.github.io/phaser3-docs/Phaser.Physics.Arcade.Body.html#damping) is used and scales the damping multiplier by the physics step delta. Drag is now velocity retained after 1 second instead of after 1 step, when damping is used.

Purpose is to make damping consistent for different physics step rates and more accurate when `fixedStep` is off (#5356).

Authors will need to change any existing drag values to get the same effects as before. Convert `drag` to `drag ^ 60` or `drag ^ fps` if you use a different step rate.

Closes #5356.

### DIFF
--- a/src/physics/arcade/Body.js
+++ b/src/physics/arcade/Body.js
@@ -335,10 +335,11 @@ var Body = new Class({
         /**
          * When `useDamping` is false (the default), this is absolute loss of velocity due to movement, in pixels per second squared.
          *
-         * When `useDamping` is true, this is 1 minus the damping factor.
+         * When `useDamping` is true, this is a damping multiplier between 0 and 1.
+         * A value of 0 means the Body stops instantly.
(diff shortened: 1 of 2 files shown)


photonstorm:
👍

PR #5391, opened 2020-11-08
Set Arcade.Body#overlapR in separateCircle()

samme:
This PR

* Fixes a bug

[Arcade.Body#overlapR](https://photonstorm.github.io/phaser3-docs/Phaser.Physics.Arcade.Body.html#overlapR) wasn't set during collisions.


photonstorm:
👍

PR #5407, opened 2020-11-23
Fix: TriangleToLine() not checking start / end point of line properly.

wiserim:
This PR

* Fixes a bug

Describe the changes below:
`Phaser.Geom.Intersects.TriangleToLine(triangle, line)` method doesn't check properly if triangle contains start or end point of line.
It's passing line's point objects to `Phaser.Geom.Triangle.Contains(x, y)`.
Method was replaced by `Phaser.Geom.Triangle.ContainsPoint(point)`.

### DIFF
--- a/src/geom/intersects/TriangleToLine.js
+++ b/src/geom/intersects/TriangleToLine.js
@@ -4,7 +4,7 @@
  * @license      {@link https://opensource.org/licenses/MIT|MIT License}
  */
 
-var Contains = require('../triangle/Contains');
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

photonstorm:
Thinking about this, we don't need to create a `Point` at all. It can be just done with `if (triangle.contains(line.x1, line.y1) || triangle.contains(line.x2, line.y2))`.

PR #5450, opened 2020-12-16
Modify arcade body to still separate bodies in the presence of micro teleports, modifications to bodies' game objects' x & y positions.

lackhand:
Addresses issues like https://www.html5gamedevs.com/topic/39840-collision-player-on-movable-platform/ (etc)

Arcade physics already had the notion of `prev`ious position (position last physics step) which was used in place of `velocity` (change in position to be added by the physics) in appropriate places in the arcade physics engine. However, the old logic treated this once-per-frame sync as a miniature reset, wiping out the previous step's position; the new logic treats this once-per-frame sync as a one-step kick in velocity, retaining the previous step's position across the frame boundary.

Bodies that wish to teleport *without* separation can continue to use `body.reset` to do so, which will cancel all `velocity` and `dx` and `dy`; if this collides two bodies, they will be embedded as in the current logic.

This should not affect any existing code (tested locally in several cases, but only verified by eye), because existing game object position + separation logic was unintuitive and unlikely to be relied upon by users.

Please advise on additional testing steps! Thank you!

### DIFF
--- a/src/physics/arcade/Body.js
+++ b/src/physics/arcade/Body.js
@@ -1030,8 +1030,6 @@ var Body = new Class({
 
         if (this.moves)
         {
-            this.prev.x = this.position.x;
-            this.prev.y = this.position.y;
             this.prevFrame.x = this.position.x;
             this.prevFrame.y = this.position.y;
         }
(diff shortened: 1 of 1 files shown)


lackhand:
This partially undoes https://github.com/photonstorm/phaser/pull/4729 -- I'm checking the `friction` examples particularly to make sure this is correct.

lackhand:
This entirely breaks friction ( per http://127.0.0.1:8080/view.html?src=src\physics\arcade\basic%20platform.js&v=dev ).
Request withdrawn, will address this issue.

PR #5451, opened 2020-12-17
Arcade collisions now support modifications to the parent game object (, and friction)

lackhand:
Addresses issues like https://www.html5gamedevs.com/topic/39840-collision-player-on-movable-platform/ , https://www.html5gamedevs.com/topic/28652-why-my-physics-arcade-overlap-function-is-not-working/?tab=comments#comment-164826 , personal annoyance, (etc)

Arcade physics already had the notion of previous position (position last physics step) which was used in place of velocity (change in position to be added by the physics) in appropriate places in the arcade physics engine. However, the old logic treated this once-per-frame sync as a miniature reset, wiping out the previous step's position; the new logic treats this once-per-frame sync as a one-step kick in velocity, retaining the previous step's position across the frame boundary.

Bodies that wish to teleport without separation can continue to use body.reset to do so, which will cancel all velocity and dx and dy; if this collides two bodies, they will be embedded as in the current logic.

This should not affect any existing code (tested locally in several cases, but only verified by eye), because existing game object position + separation logic was unintuitive and unlikely to be relied upon by users.

Please advise on additional testing steps! Thank you!

Second attempt at https://github.com/photonstorm/phaser/pull/5450 .
Related work: #4729  .

Some tests in https://github.com/photonstorm/phaser3-examples/compare/master...browndragon:arcade3?expand=1 (I can make that more formal so it can be checked in if you prefer).

### DIFF
--- a/src/physics/arcade/Body.js
+++ b/src/physics/arcade/Body.js
@@ -1022,24 +1022,29 @@ var Body = new Class({
         if (willStep)
         {
             this.resetFlags();
+
+            if (this.moves)
+            {
+                // Snapshot the last physics step position *before* syncing from the gameObject per `prev`'s contract. This lets physics see any changes from tweens etc to the game object's position. If a step isn't going to be executed, it's not correct to snapshot, since then movement direction might be lost.
+                // You can avoid this by explicitly calling `reset` to "teleport" a gameobject around.
+                this.prev.x = this.position.x;
+                this.prev.y = this.position.y;
+            }
         }
 
         this.updateFromGameObject();
 
         this.rotation = this.transform.rotation;
         this.preRotation = this.rotation;
-
         if (this.moves)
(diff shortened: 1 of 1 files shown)


lackhand:
Came up in discord today: https://discord.com/channels/244245946873937922/416623653741133837/806589278351917127

lackhand:
I still stand by this, but I've personally moved my work over to https://github.com/browndragon/phaser/tree/solidArcade .
I'll append a blog post with an explanation of findings as well.
Once that's appending, I'll close this request.

lackhand:
https://browndragon.tumblr.com/post/650184434757386240/the-problem-with-arcade-physics -- thanks! Let me know if there's interest in forming this into a real bug report, a patch, etc. I'll close this particular PR now!

PR #5467, opened 2020-12-25
Remove a type of Post Pipeline instances, not only a single instance

rexrainbow:

* Adds a new feature/bug fix

Describe the changes below:

A post-fix pipeline class can be added to game object many times, via different instances.

Previously, `removePostPipeline(pipelineClass)` method only removes a single pipeline instance. It won't clear all instances of pipelineClass. This PR will remove all instances of pipelineClass.


photonstorm:
👍

PR #5472, opened 2020-12-27
Check event.type in duplicate event bailout

### LINKED ISSUE #5471: Keyboard key event filtered out incorrectly as a duplicate
## Version

* Phaser Version: 3.50.1

* Operating system: Windows

* Browser: Firefox 84

## Description

With any keyboard input, sometimes the keyup event doesn't take effect, because it is filtered out by the keyboard plugin. This happens when a keydown and keyup event for the same key have the same timestamp. Probably related to repeating keydowns lining up with a keyup.

## Example Test Code

The make your first Phaser 3 game tutorial at https://www.phaser.io/tutorials/making-your-first-phaser-3-game/part7 exhibits the problem.

## Additional Information

The keyboard plugin checks for duplicate key events by checking key code and timestamp, but it doesn't check the type. It should check the type.

cjw6k:
Fixes #5471.

* Fixes a bug

Describe the changes below:
Check that the current key event type is the same type as the previous event event, before bailing out of event processing.

### DIFF
--- a/src/input/keyboard/KeyboardPlugin.js
+++ b/src/input/keyboard/KeyboardPlugin.js
@@ -167,6 +167,16 @@ var KeyboardPlugin = new Class({
          */
         this.prevTime = 0;
 
+        /**
+         * Internal repeat key flag.
+         *
+         * @name Phaser.Input.Keyboard.KeyboardPlugin#prevType
+         * @type {string}
+         * @private
+         * @since 3.50.1
+         */
+        this.prevType = null;
+
         sceneInputPlugin.pluginEvents.once(InputEvents.BOOT, this.boot, this);
         sceneInputPlugin.pluginEvents.on(InputEvents.START, this.start, this);
     },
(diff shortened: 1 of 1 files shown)


src/input/keyboard/KeyboardPlugin.js
```diff
@@ -747,14 +747,15 @@ var KeyboardPlugin = new Class({
...
-            if (code === this.prevCode && event.timeStamp === this.prevTime)
+            if (code === this.prevCode && event.timeStamp === this.prevTime && event.type === this.prevType)
             {
                 //  On some systems, the exact same event will fire multiple times. This prevents it.
                 continue;
             }
 
             this.prevCode = code;
             this.prevTime = event.timeStamp;
+            this.prevType = event.type;
```
samme, line 758: I think you'll have to add `this.prevType` to the constructor as well.

photonstorm:
👍

PR #5487, opened 2021-01-05
added null check for WebAudio pannerNode

jdcook:
This PR

* Fixes a bug

Describe the changes below:
Added a null check for pannerNode in WebAudio on destroy. This was throwing an error in Safari, since pannerNode is null.


photonstorm:
Thanks. I will have to lint fix this, but well spotted.

Issue #5380, opened 2020-10-31
[3.50.0-beta.9] Mesh


MadDogMayCry0:
https://codesandbox.io/s/clever-swanson-7hfw5
https://7hfw5.csb.app/
http://doctoral-arrangemen.000webhostapp.com/mesh/

The mesh is not displayed. I tried many different ways, but the result is always the same - the mesh does not appear on the stage.

The sample code is taken from here
https://www.patreon.com/posts/42666302
![image](https://user-images.githubusercontent.com/2493591/97786299-dc56d300-1baa-11eb-876f-111da8ad3510.png)

photonstorm:
You have to use the latest build from master. Or Beta 10 will work fine, too. Here's a fully working example:

```js
class Example extends Phaser.Scene
{
    constructor ()
    {
        super();
    }

    preload ()
    {
        this.load.image('bg', 'assets/skies/gradient13.png');
        this.load.image('ayu', 'assets/pics/ayu.png');
    }

    create ()
    {
        this.add.image(400, 300, 'bg').setFlip(false, true);

        const vertices = [
            -1, 1,
            1, 1,
            -1, -1,
            1, -1
        ];

        const uvs = [
            0, 0,
            1, 0,
            0, 1,
            1, 1
        ];

        const indicies = [ 0, 2, 1, 2, 3, 1 ];

        const mesh = this.add.mesh(400, 300, 'ayu');

        mesh.addVertices(vertices, uvs, indicies);

        mesh.panZ(7);

        this.debug = this.add.graphics();

        mesh.setDebug(this.debug);

        this.add.text(16, 16, 'Rotate with mouse (+ Shift to pan)\nWheel to zoom\nD to toggle debug');

        this.input.keyboard.on('keydown-D', () => {

            if (mesh.debugCallback)
            {
                mesh.setDebug();
            }
            else
            {
                mesh.setDebug(this.debug);
            }

        });

        const rotateRate = 1;
        const panRate = 1;
        const zoomRate = 4;

        this.input.on('pointermove', pointer => {

            if (!pointer.isDown)
            {
                return;
            }

            if (!pointer.event.shiftKey)
            {
                mesh.modelRotation.y += pointer.velocity.x * (rotateRate / 800);
                mesh.modelRotation.x += pointer.velocity.y * (rotateRate / 600);
            }
            else
            {
                mesh.panX(pointer.velocity.x * (panRate / 800));
[...]

PR #5498, opened 2021-01-12
Update Mesh.js documentation around panZ

lackhand:

* Updates the Documentation

Describe the changes below:

This adds warnings about orthographic and perspective cameras, documents setOrtho & panZ with sane (ish) default values for the default projection matrix, and updates the example code in the file.

These changes could go further; I would love it if the default mechanisms "agreed" (so that feeding it z=0 vertex values "just worked"), but that's probably a difficult backwards facing change so I did not make it here.

Inspired by https://github.com/photonstorm/phaser/issues/5380 ; fyi @MadDogMayCry0


photonstorm:
👍

PR #5499, opened 2021-01-13
Fix typo in GenerateVerts.js

lackhand:

* Fixes a bug

Describe the changes below:

The result contains `["verts"]`, but the caller expects [`["vertices"]`](https://github.com/photonstorm/phaser/blob/d67c93646cbfe8ba5997954f8af29a8f3f9e8e91/src/gameobjects/mesh/Mesh.js#L705). Better to expand the field name (IMO).


photonstorm:
👍

PR #5503, opened 2021-01-14
Transparent background for banner gameURL

kainage:
This PR

* Fixes a bug

Describe the changes below:

Use transparent background for banner gameURL to support different browser color themes.
Currently, when using a dark theme for the dev console, both the background and the text are white.


photonstorm:
👍

Issue #5470, opened 2020-12-27
DOMElement configure pointerEvents independently of transformOnly


hayatae:
Currently, the pointerEvents style property of DOMElements is set on render to be `auto` if the element is not in `transformOnly` mode. This creates a bit of an issue if you'd like to have all the other advantages of having phaser track opactiy, display, zIndex, and blendMode, but do not want the mouse pointer to be intercepted by the element. It is not clear to me why pointerEvents are tied into these display properties.

https://github.com/photonstorm/phaser/blob/416ae58da3e010f051fb03db58389a001536cfbc/src/gameobjects/domelement/DOMElementCSSRenderer.js#L106

It seems like an easy fix would be to expose a separate `pointerEvents` property which could be set to any valid css option for `pointer-events`. This property could be defaulted based on the `trasnformOnly` property to maintain backwards compatibility. This would allow a developer to override the `pointer-events` to anything they need it to be.

I'd be comfortable implementing this feature and creating a PR for it if this sounds like a good approach.

endel:
Hey,  I'm also in the need of this 🙋

I'm wondering if `style.pointerEvents` really needs to be set at all? Maybe the end-user should be responsible for setting this CSS property, like this:

```typescript
scene.add.dom(x, y, "div", `
  color: #000;
  pointer-events: none;
`, "hello world");
```

It seems the `pointer-events` is being overwritten in the current implementation, right?

endel:
The [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events) mentions that `auto` is the default value. So I believe it's safe to remove it. 

> `auto`: The element behaves as it would if the pointer-events property were not specified. 

Made a PR for this here 😅  https://github.com/photonstorm/phaser/issues/5470

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

photonstorm:
This change has sadly broken so many things it's time to revert it all back to how it was previously.

hayatae:
> This change has sadly broken so many things it's time to revert it all back to how it was previously.

I was afraid of that happening. Can we fix this by simply adding a new separate property (as suggested in the initial post) to allow overriding the default behavior in Phaser? That would allow for overriding this property if desired while not affecting any existing code.

photonstorm:
[...]

PR #5504, opened 2021-01-14
allow to customize pointerEvents for dom elements

endel:
Fixes https://github.com/photonstorm/phaser/issues/5470


photonstorm:
👍

Issue #3911, opened 2018-08-10
[Feature Request] I need to skew my image


jcyuan:
is this possible? I think this is a basic transform feature?

jcyuan:
or please provide raw operation on GameObject's matrix, thank you very much.

jcyuan:
@photonstorm thanks for replying, actually I'm integrating DragonBones into Phaser3, and all things done except skew. an object to attach to a slot will be a GameObjects.Image or GameObjects.Container, so which is the best way to implement this? to use mesh I think (although currently mesh does not have indices but still ok for the basic display - DragonBones advanced functionality uses vertex grid which needs indices to do disortion etc).

jcyuan:
a high advantage that I use GameObjects.Image to attach to a slot is that it's very easy to get animation work, just set new frame for it that's enough, If use Mesh I need to figure out how to implemenet this, let me have a review in the source code.

photonstorm:
Thank you for opening this feature request (all those months ago!). As the development of Phaser 4 has now started I am tidying-up the repo and closing off old issues that won't be directly fixed or implemented in Phaser 3 by me. I would still consider community PRs that address this issue, however, and have left the GitHub tags intact so others can track them if they wish to get involved. However, I'm afraid this is no longer a feature that will be added to v3.

PR #5505, opened 2021-01-14
Add skew component to transform & pipeline

lackhand:
Adds skew x & y to the Transform component, TransformMatrix, and MultiPipeline.

Example: https://github.com/photonstorm/phaser3-examples/pull/335

### DIFF
--- a/src/gameobjects/components/GetBounds.js
+++ b/src/gameobjects/components/GetBounds.js
@@ -7,6 +7,7 @@
 var Rectangle = require('../../geom/rectangle/Rectangle');
 var RotateAround = require('../../math/RotateAround');
 var Vector2 = require('../../math/Vector2');
+var TransformMatrix = require('./TransformMatrix');
 
 /**
  * Provides methods used for obtaining the bounds of a Game Object.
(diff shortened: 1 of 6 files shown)


lackhand:
This was requested in #3911 .

lackhand:
[force push was to fix lint]

photonstorm:
Hmm, this is really frustrating because I love the thought of adding this feature! But it definitely breaks input (skewed objects are not correctly hit detected any longer) and breaks all the Bounds functions (and anything that relies on them, like alignment) and it'll certainly cause physics bodies to be misaligned, although that's a more manageable problem I guess.

lackhand:
Input is *easy*, I've updated with a fix for that; tests updated @ https://github.com/photonstorm/phaser3-examples/pull/335 .
Meanwhile, I'll take a stab @ Bounds (, alignment, ?).

I'd appreciate some advice on physics; obviously arcade is going to be screwy (but it doesn't work well with scale or rotation as is, so how worried should I be?). I'm guessing, if anything, this would be an acceptable break? LMK.

lackhand:
`GetBounds` required me to double check my linear algebra, but it also is supported (test updated). I've still done nothing with physics; let me know your preferred next steps!

photonstorm:
I'm really sorry, but I'm going to close this PR without merging it. I love the thought of adding this, and I really do appreciate the work you put in, but I honestly think it's going to cause me so many problems in the future, for edge cases we haven't even considered, that I'm happy for this to remain a Phaser 4 only feature (where it's already implemented).

PR #5512, opened 2021-01-19
Move webpack loaders to dev dependencies.

andrewstart:
* Fixes a bug

Describe the changes below:
Moves webpack loaders (exports-loader & imports-loader) to devDependencies. Being in the runtime dependencies meant that Phaser ended up with webpack as a peer dependency, and that was causing problems for me, at least when using PNPM.
This wasn't an issue before 3.50.0, because the versions of the two loaders in use weren't declaring their peer dependency yet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,9 @@
     "dts-dom": "^3.6.0",
     "eslint": "^7.11.0",
     "eslint-plugin-es5": "^1.5.0",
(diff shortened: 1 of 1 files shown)


photonstorm:
Sure, that's fine. I don't think `path` is needed as a dep either tbh.

PR #5514, opened 2021-01-19
Fix typing of MapData objects

etherealmachine:

* Updates the Documentation

Describe the changes below:

I believe that objects is going to be an array as the result of calling this function: https://github.com/photonstorm/phaser/blob/7834095789fd3ac24ee0146e65eefa096b22b1d6/src/tilemaps/parsers/tiled/ParseObjectLayers.js#L20

I'm not 100% sure on the type of the array elements (Phaser.Types.Tilemaps.ObjectLayerConfig), but that appears to be what ParseObjectLayers returns.


photonstorm:
👍

Issue #5506, opened 2021-01-15
Container Input doesn't work


vforsh:
## Version
Phaser 3.52 (3.51 and 3.50.1 were fine)

## Description
Here is the scene display graph
- Scene
-- Background image
-- Container
-- Image

All 3 game objects on the scene have input enabled (via setInteractive() call) but the container behaves like it has no input enabled. It seems like its hit area is located below the background image even though they render in the correct order.

## Example Test Code
https://codesandbox.io/s/container-input-issue-my7gl
Try commenting line 21 to see the issue

arbassic:
The issue looks more common

Any interactive element placed below or within an interactive container captures input and do not pass it to the container.

Example Test Code
----------

https://codesandbox.io/s/container-input-issue-2-2gqzp

boo-ga-ga:
I think something similar also affects rexUI plugin dialog, in some cases the buttons stop being clickable. 3.51.0 and 3.50.0 were fine as well.

rexrainbow:
@boo-ga-ga 
rexUI dose not use built-in container (`scene.add.container`)

rexrainbow:
@boo-ga-ga 

I saw that clicking bug in dialog, but rexUI dose not use built-in container.

boo-ga-ga:
@rexrainbow I see, thank you very much for the fix and your work!

photonstorm:
This issue has been mentioned on **Phaser**. There might be relevant details there:

https://phaser.discourse.group/t/problem-with-over-out-events-for-two-overlapping-containers/8791/3

photonstorm:
This issue has been mentioned on **Phaser**. There might be relevant details there:

https://phaser.discourse.group/t/problem-with-over-out-events-for-two-overlapping-containers/8791/6

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

Stever1388:
I'm still having issues with this in 3.53 and 3.54, the following code demonstrates the issue:
`this.add.rectangle(0, 0, this.sys.canvas.width, this.sys.canvas.height, 0x351B03, 0.4).setOrigin(0).setInteractive()
[...]

PR #5517, opened 2021-01-21
Fix input bug

rexrainbow:

* Fix bug

Describe the changes below:

Container, and Zone game object can be interactive ( `container.setInteractive()`). But they are not add to render list (`camera.addToRenderList(src)`).

Related issue : #5506

### DIFF
--- a/src/gameobjects/container/ContainerWebGLRenderer.js
+++ b/src/gameobjects/container/ContainerWebGLRenderer.js
@@ -29,6 +29,8 @@ var ContainerWebGLRenderer = function (renderer, container, camera, parentMatrix
         return;
     }
 
+    camera.addToRenderList(src);
+
     var transformMatrix = container.localTransform;
 
     if (parentMatrix)
--- a/src/gameobjects/container/ContainerWebGLRenderer.js
+++ b/src/gameobjects/container/ContainerWebGLRenderer.js
@@ -29,7 +29,7 @@ var ContainerWebGLRenderer = function (renderer, container, camera, parentMatrix
         return;
     }
 
-    camera.addToRenderList(src);
+    camera.addToRenderList(container);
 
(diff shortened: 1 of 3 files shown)


src/gameobjects/container/ContainerWebGLRenderer.js
```diff
@@ -29,6 +29,8 @@ var ContainerWebGLRenderer = function (renderer, container, camera, parentMatrix
         return;
     }
 
+    camera.addToRenderList(src);
```
veleek, line 32: What is src here?  I don't see where this value is supposed to be defined?

Should this be container instead?

rexrainbow: `camera.addToRenderList(src)` is a new method introduced by 3.52.0, other renderer of game object like [Image](https://github.com/photonstorm/phaser/blob/master/src/gameobjects/image/ImageWebGLRenderer.js#L23) also has this command inside.

veleek: Yes, but what is the value of src here.  It appears to be undefined.

If you look at the Image class you linked, the function accepts `src` as a parameter:
```
var ImageWebGLRenderer = function (renderer, src, camera, parentMatrix)
```

ContainerWebGLRenderer has a different parameter name:

```
var ContainerWebGLRenderer = function (renderer, container, camera, parentMatrix)
```

So you should change this to `camera.addToRenderList(src);`

rexrainbow: Thanks, I change `src` to `container` in my next submit

veleek (approved):

photonstorm:

https://phaser.discourse.group/t/phaser-3-rexui-plugins/384/80

photonstorm:
👍

PR #5523, opened 2021-01-23
Load Tiled object properties using names.

veleek:
This PR 

* Fixes a bug

Describe the changes below:

Tiled objects have properties as an array of objects like:

```
[
  { name: ..., type: ..., value: ...},
  { name: ..., type: ..., value: ...},
  ...
]
```

Update the parser to load these using the name/value properties instead
of just treating it as an array and loading properties named '0', '1',
etc.  This is the same as how properties are handled in ParseTilesets.js

### DIFF
--- a/src/tilemaps/Tilemap.js
+++ b/src/tilemaps/Tilemap.js
@@ -831,15 +831,34 @@ var Tilemap = new Class({
                 }
 
                 //  Set properties the class may have, or setData those it doesn't
-                for (var key in obj.properties)
+                if (Array.isArray(obj.properties))
                 {
-                    if (sprite[key] !== undefined)
+                    // Tiled objects custom properties format
+                    obj.properties.forEach(function (propData)
                     {
-                        sprite[key] = obj.properties[key];
-                    }
-                    else
+                        var key = propData['name'];
+                        if (sprite[key] !== undefined)
+                        {
+                            sprite[key] = propData['value'];
+                        }
+                        else
+                        {
+                            sprite.setData(key, propData['value']);
+                        }
+                    });
+                }
+                else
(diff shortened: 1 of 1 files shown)


photonstorm:
Do you know when was this change made to the Tiled JSON format?

It must be quite recent because that's definitely not how properties were handled when this method was rewritten.

veleek:
@photonstorm - Sorry, I do not.  I just recently started using Tiled.  I can try to dig through the source and see if I can identify when.  If it used to be in a different format then I can see about making the code handle both styles.

If you have an example of the original Tiled format that would also be pretty easy for me to change.

photonstorm:
The original format is how the current code deals with it. Any of the tmx files in the phaser3 examples repo use this format. Hence why it must be a recent change. Probably listed in the Tiled change logs somewhere.

veleek:
@photonstorm - Took a closer look and the GitHub diff viewer is really bad so it's way more confusing then the change actually was.

My implementation checks if it's an array and loads it using the new pattern, otherwise it falls back to the old pattern.  I could potentially use a version check but I don't know enough about the tiled source to be able to identify when the change occurred.  Are you okay with the best-effort version implemented?

veleek:
@photonstorm - As per this comment in Tiled https://github.com/mapeditor/tiled/issues/2034#issuecomment-439184633, it looks like this was changed as of Tiled 1.2.  So, same question as before, do you think it's necessary to have a version check or is the Array check as is enough?

photonstorm:
I'm sure the array check is enough, however this change was made over 2 years ago according to that comment - has it since reverted or are you just using a really old version of Tiled?

PR #5533, opened 2021-01-28
Fix IsometricWorldToTileXY result

veleek:
This PR

* Fixes a bug

Describe the changes below:

IsometricWorldToTileXY returns a tile that is offset by (-0.5, 0.5)
tiles from the expected location.  This fixes the function so that
Tile => World => Tile conversion returns the same value.


photonstorm:
👍

PR #5543, opened 2021-02-03
Warn and call onProcessError() for invalid JSON

samme:
This PR

* Adds a new feature
* Fixes a bug

New: when a loaded JSON file fails to parse, there's a console warning with the file key. Without this it was difficult to find which file failed.

Fixed: when a loaded JSON file fails to parse, it's marked [FILE_ERRORED](https://photonstorm.github.io/phaser3-docs/Phaser.Loader.html#.FILE_ERRORED) and the loader continues. Before this change the loader would stall.


photonstorm:
👍

PR #5546, opened 2021-02-07
Fix PlaceOnRectangle JSDoc

xuxucode:
This PR:
* Updates the Documentation

The shift value defaults to 0


photonstorm:
Thanks for this PR. However, the JSDocs are correct but the default value is wrong! Line 30 should be `if (shift === undefined) { shift = 1; }` as that is what's passed to the `RotateLeft` and `RotateRight` functions.

xuxucode:
Thanks for your feedback, I am a newbie in Phaser, I wonder why 1 rather than 0 is default value, could you explain it?

photonstorm:
👍

Issue #5493, opened 2021-01-08
Spine Game object above an ellipse in the displayList makes every sprite blue


EmilSV:
## Version
* Phaser Version: 3.51.0
* Operating system: Windows

## Description

if a spine Game object is above an ellipse game object in the displayList every sprite/image becomes purely blue

## Example Test Code
```
class Example extends Phaser.Scene
{
    constructor ()
    {
        super({
            pack: {
                files: [
                    { type: 'scenePlugin', key: 'SpinePlugin', url: 'plugins/3.8.95/SpinePluginDebug.js', sceneKey: 'spine' }
                ]
            }
        });
    }

    preload ()
    {
        this.load.image('logo', 'assets/sprites/phaser.png');

        this.load.setPath('assets/spine/3.8/demos/');

        this.load.spine('set1', 'demos.json', [ 'atlas1.atlas' ], true);
    }

    create ()
    {
        this.add.image(0, 0, 'logo').setOrigin(0);

        const boy = this.add.spine(460, 600, 'set1.spineboy', 'idle', true);
        const circle = this.add.ellipse(130, 200, 250, 250, 0xffffff);

        this.sys.displayList.bringToTop(boy);
    }
}

const config = {
    type: Phaser.WEBGL,
    parent: 'phaser-example',
    width: 800,
    height: 600,
    backgroundColor: '#2d2d2d',
    scene: Example
};

const game = new Phaser.Game(config);

```
![blue_bug](https://user-images.githubusercontent.com/3351370/104067459-808c7400-5203-11eb-8c87-3c89728049b1.gif)

EmilSV:
maybe this is related to #5449

photonstorm:
This is almost certainly the same as #5449 but I'll leave it open for now.

FloodGames:
Thanks a lot for opening!

A very clear replication demo.

samme:
https://codepen.io/samme/pen/MWbKeEE

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

PR #5548, opened 2021-02-07
Spine render fix

EmilSV:
This PR fixes the bug described in #5493

The issue is when a batch of spine object is done rendering the vertexAttribArray they use are disable and normally they use 3 attributes stored at location 0,1,2.

"renderer.pipelines.rebind()" would then try to restore the vertexAttribArray to their state they had before but as GraphicsPipeline only use 2 attributes only 2 attributes  are restore 0,1 not 2.

to fix this issue the location of the enabled vertexAttribArray  are now cached by the plugin before the batch of spine object begin rendering and restored when it ends.


photonstorm:
Many thanks for spending time opening this PR and figuring out the problem, it saved me lots of time.

I'm not going to merge this PR because the issue actually needed fixing within the Pipeline Manager, not within the Spine Plugin itself. This fix has now been done and I can confirm this bug is resolved as a result (without needing to rebuild the Spine Plugin).

PR #5551, opened 2021-02-09
Fix missing output arg in Phaser.Curves.Line#getTangent

samme:
This PR

* Fixes a bug

With [a Path of Line Curves](http://labs.phaser.io/view.html?src=src/paths/zig%20zag%20path.js "zig zag path"), if you pass an output vector to Path#getTangent() the vector doesn't receive the tangent value:

```js
path.getTangent(0.5, output);
// `output` is still <0, 0>
```

I fixed it by adding optional arguments `t` and `out` to Phaser.Curves.Line#getTangent().


photonstorm:
👍

PR #5559, opened 2021-02-11
Fixes #5558

### LINKED ISSUE #5558: Error when destroying Text game object with headless renderer
## Version

* Phaser Version: 3.52.0

* Operating system: macOS

## Description

The following error is thrown when destroying a Text GameObject while using Phaser.HEADLESS as the renderer

```
Uncaught TypeError: Cannot read property 'type' of null
```

which seems to be pointing to [this line](https://github.com/photonstorm/phaser/blob/master/src/textures/Texture.js#L512). Should be just a matter of adding a check for `renderer && renderer.type`. I can make a PR for this tomorrow, but there might be other places where a similar check is needed.

I remember a similar bug happening in #4976. Is there a way we can prevent similar regressions on this in the future? Perhaps a utility function that safely returns the renderer.type, eliminating the risk of accidentally forgetting to check it safely in the future?

## Example Test Code

[CodeSandbox repro](https://codesandbox.io/s/phaser-headless-bug-v1nzp?file=/src/index.js)

## Additional Information

Only started happening after upgrading from 3.24.1

mattjennings:

* Fixes a bug

Describe the changes below:

checks that renderer exists before accessing renderer.type to prevent error thrown in #5558

### DIFF
--- a/src/display/mask/BitmapMask.js
+++ b/src/display/mask/BitmapMask.js
@@ -146,7 +146,10 @@ var BitmapMask = new Class({
 
         scene.sys.game.events.on(GameEvents.CONTEXT_RESTORED, this.createMask, this);
 
-        renderer.on(RenderEvents.RESIZE, this.createMask, this);
+        if (renderer)
+        {
+            renderer.on(RenderEvents.RESIZE, this.createMask, this);
+        }
     },
 
     /**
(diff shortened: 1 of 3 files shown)


photonstorm:
👍

PR #5571, opened 2021-02-20
Polygons fixes

duberti:

* Fixes a bug

Describe the changes below:
- Fixed Path not being added to display list
- Fix on line to line intersection code

### DIFF
--- a/src/curves/path/Path.js
+++ b/src/curves/path/Path.js
@@ -869,7 +869,7 @@ var Path = new Class({
  */
 GameObjectFactory.register('path', function (x, y)
 {
-    return new Path(x, y);
+    return this.displayList.add(new Path(x, y));
 });
 
 //  When registering a factory function 'this' refers to the GameObjectFactory context.
--- a/src/geom/intersects/GetLineToLine.js
+++ b/src/geom/intersects/GetLineToLine.js
@@ -55,7 +55,7 @@ var GetLineToLine = function (line1, line2, out)
     var T1 = (x3 + dx2 * T2 - x1) / dx1;
 
     //  Intersects?
-    if (T1 < 0 || T2 < 0 || T2 > 1)
+    if (T1 < 0 || T2 < 0 || T1 > 1 || T2 > 1)
     {
         return null;
     }


photonstorm:
`Path` should not be added to the Display List as it does not extend `GameObject` and therefore doesn't provide any of the properties or methods required for passing any of the render checks. What was the reasoning for adding it?

duberti:
Hi, 

Yes sorry, I realized that later, solution seems to be more complex than that...

The thing is that adding a Path to a scene doesn’t seem to work the same way as adding circle or other figures but semantics is the same.

I believe path method doesn’t seem to be ok as it just creates a path instance from parameters and return it without adding it to anything.

Are we missing a GameObject representation for paths, or what would you suggest should be the right approach here? 

Regards.

photonstorm:
Rendering a path is quite unusual, but you can do it by using the Path's ability to draw to a Graphics object. It has methods for that built into it. Anything more complex should be handled directly by your own code.

duberti:
Hi Richard, ok got you. What about the other change? Without validating it’s < 1 you could end up having intersections outside the line segment span.

Regards.

photonstorm:
Can you post a simple example showing how < 1 fails currently, please?

duberti:
Hi Richard,

Here's an example that requires this patch:
var line1 = [{x:100,y:100},{x:200,y:200}]
var line2 = [{x:300,y:100},{x:300,y:300}]

The value for T1 will be calculated as 2, which means it goes outside the
line segment parameterized between [0, 1]

But besides that I've found more errors on the implementation of
GetLineToLine. Another easy to see error is that whenever the line1 is
straight vertical it will not intersect with anything as there's code that
returns null whenever dx1 === 0
[image: image.png]

What are your thoughts regarding collinear lines that overlap? intersection
definition says that result is everything that belongs to both groups but
on this code it returns null, basically if you intersect to equal lines you
will get null as result which is weird...

So I'm working on an improved version of this method as I need a fully
reliable line intersection that solves all these issues and I couldn't find
any other 100% reliable implementation.

photonstorm:
If you want to update the function to deal with coincidental lines, that's fine.

photonstorm:
This was fixed via #5701, so going to close this PR.

PR #5589, opened 2021-03-02
Stop RGBToString from returning CSS strings with decimal places

neil-h:

This PR
* Fixes a bug

Describe the changes below:
Currently if the r g b arguments to RGBToString contain decimal places, then outputting a string with the # prefix may result in decimals places in the CSS string, e.g. '#898989.4cccccd'.
This change slices off anything after the 6 proper digits.


photonstorm:
👍

PR #5601, opened 2021-03-14
Documentation Group#getMatching example fix

masterT:
This PR

* Updates the Documentation

Fixes example for `Phaser.GameObjects.Group#getMatching` which was not using the right function.


photonstorm:
👍

PR #5612, opened 2021-03-18
Remove and destry all Keys in Scene.Shutdown()

### LINKED ISSUE #5611: Shutdown does not disconnect event handles for keys created with this.input.keyboard.addKey
## Version

* Phaser Version: 3.50.1

* Operating system:

* Browser:

## Example Test Code

Calling `Stop()` on a scene does not disconnect event handlers associated with Key objects created by `this.input.keyboard.add(...)`.

```
var MyGame = {};

MyGame.Boot = function ()
{
};

MyGame.Boot.prototype.constructor = MyGame.Boot;

MyGame.Boot.prototype = {

    preload: function ()
    {
        this.load.image('rick', 'assets/pics/rick-and-morty-by-sawuinhaff-da64e7y.png');
    },

    create: function ()
    {
        this.add.image(400, 300, 'rick').setScale(0.7);

        this.input.keyboard.on('keydown-T', function() {
            console.log("Key T - Registered using keyboard.on(...)")
        }, this);

        this.input.keyboard.addKey('Y').on('down', function() {
            console.log("Key Y - Registered using keyboard.addKey(...).on(...)")
        }, this);

        this.input.once('pointerdown', function () {
            console.log("Restarting scene...");
           this.scene.restart();
        }, this);
    }

};

var config = {
    type: Phaser.AUTO,
    parent: 'phaser-example',
    width: 800,
    height: 600
};

game.scene.add('Boot', MyGame.Boot, true);
```

## Additional Information

It appears that the `shutdown` function in `KeyboardPlugin` calls `this.resetKeys()` which resets the state of the key (but does not call `removeAllListeners()`), and then it calls `removeAllListeners` on itself.

So I think that keyboard events created using `this.input.keyboard.on(...)` will be disconnected during Shutdown, but not keys created by `this.input.keyboard.addKey(...)`.  

Does that seem like a correct assessment of the situation?  If so I can pull together a PR to fix it.

Just for reference, there's a bit of conversation on this thread: https://github.com/photonstorm/phaser/issues/5132#issuecomment-799967946.

veleek:

* Fixes #5611

Describe the changes below:
Update KeyboardPlugin.Shutdown() to destroy any Key objects that have been created.

### DIFF
--- a/src/input/keyboard/KeyboardPlugin.js
+++ b/src/input/keyboard/KeyboardPlugin.js
@@ -852,7 +852,7 @@ var KeyboardPlugin = new Class({
     /**
      * Shuts this Keyboard Plugin down. This performs the following tasks:
      *
-     * 1 - Resets all keys created by this Keyboard plugin.
+     * 1 - Removes all keys created by this Keyboard plugin.
      * 2 - Stops and removes the keyboard event listeners.
      * 3 - Clears out any pending requests in the queue, without processing them.
      *
@@ -862,7 +862,8 @@ var KeyboardPlugin = new Class({
      */
     shutdown: function ()
     {
-        this.resetKeys();
+        this.removeAllKeys(true);
+        this.removeAllListeners();
 
         this.sceneInputPlugin.manager.events.off(InputEvents.MANAGER_PROCESS, this.update, this);
 
@@ -870,8 +871,6 @@ var KeyboardPlugin = new Class({
 
         this.scene.sys.events.off(SceneEvents.PAUSE, this.resetKeys, this);
         this.scene.sys.events.off(SceneEvents.SLEEP, this.resetKeys, this);
-
-        this.removeAllListeners();
 
         this.queue = [];
     },


photonstorm:
Honestly, this feels like something that you should handle directly, but I'll merge it for now and we can roll it back if it causes issues down the line.

veleek:
@photonstorm  - I appreciate the merge!

The fact that events registered with `this.input.keyboard.on(...)` are removed as part of shutdown makes this feel consistent.  The key objects are "owned" by the same KeyboardPlugin instance so it doesn't feel any different.  If you think this should be handled directly, do you think that events registered directly should also not be removed as part of shutdown?  Could be worth having a longer conversation just so that how event handlers are deregistered is consistent.

photonstorm:
It's not really about tidying up events. All input events are cleared when a Scene shuts down, but `Key`s are objects that can easily be created both via the Keyboard Plugin and outside of it. Now they will be removed if created via `addKey`, but not if created directly, so it's sort of a split responsibility. Previously you just had to remember to do it yourself, now you _might_ have to :)

veleek:
Ah, I see.  That makes more sense.  Taking a look at the Key code, the constructor does have a parameter:

> * @param {Phaser.Input.Keyboard.KeyboardPlugin} plugin - The Keyboard Plugin instance that owns this Key object.

So since it explicitly states that the plugin owns the key instance, maybe the correct thing to do is to add the created key to the plugin so that it get's cleaned up when the plugin is destroyed which would make it more consistent.

PR #5614, opened 2021-03-21
Do not set pointer-events property on DOM elements container

### LINKED ISSUE #5594: Can't click to focus an input dom element since 3.53
Testable in this official example: [dom input example](https://phaser.io/examples/v3/view/game-objects/dom-element/form-input)
works fine on 3.52

pizkaz:
Fixes #5594.

If I understand correctly, #5504 tried to achieve that the property is left
at its default value. The fact that the DOM container also defines it
was probably overlooked. :-)

### DIFF
--- a/src/dom/CreateDOMContainer.js
+++ b/src/dom/CreateDOMContainer.js
@@ -25,7 +25,6 @@ var CreateDOMContainer = function (game)
         'padding: 0; margin: 0;',
         'position: absolute;',
         'overflow: hidden;',
-        'pointer-events: none;',
         'transform: scale(1);',
         'transform-origin: left top;'
     ].join(' ');


photonstorm:
👍

khasanovbi:
This change breaks my pointermove handling.
I can't customize pointer events of dom container. 

I have a html Controller button that are in top of canvas:
```html
<div style="display: block; width: 1848px; height: 607px; padding: 0px; margin-right: 0px; margin-bottom: 0px; position: absolute; overflow: hidden; transform: scale(1, 1); transform-origin: left top;">
    <div style="height: 70px; line-height: 70px; z-index: 0; display: block; position: absolute; opacity: 1; mix-blend-mode: normal; transform: matrix(1, 0, 0, 1, 0, 0) skew(0rad, 0rad) rotate3d(0, 0, 0, 0deg); transform-origin: 0% 0%;" class="b-game__qr-button js-qr-button">
        Controller
    </div>
</div>
```

It look like dom container doesn't allow to handle pointermove events on canvas.

PR #5624, opened 2021-03-26
Allow to have one row matrix in CheckMatrix

andriibarvynko:

* Fixes a bug

Describe the changes below:

According to current logic:
[[1, 1, 1, 1]] - this is NOT a matrix
[[1], [1], [1], [1]] - but IT IS a matrix

in case we rotate [[1], [1], [1], [1]] twice we will get a null. So I suggest to not limit rows for matrix

Example which demonstrate the issue:
https://jsfiddle.net/andriibarvynko/3Lx5qvym/1/


photonstorm:
Sure, why not. This is such an edge case I can't see it harming anything to change it.

PR #5632, opened 2021-03-29
Add Phaser.Math.MapLinear method

vforsh:
This PR

* Adds a new feature

Describe the changes below:
The method from *phaser-ce*. I don't know why it wasn't added to v3, was it intentional or Rich just forgot about it. But I find this method quite useful, use it often and think it should be the part of Phaser.


photonstorm:
We've already got `Phaser.Math.Linear` and `Phaser.Math.Interpolation.Linear` for interpolating between an array of values. I don't really see what this brings extra.

Issue #5633, opened 2021-03-30
Phaser 3.54 draw offset layer


moJiXiang:
## Version
* Phaser Version: 3.54.0
* Operating system: Windows
* Browser: Chrome

## Description

TileWidth = 62, TileHeight  = 31

White grid layer is the tilemap layer area.

Phaser 3.54 draw isometric map offset of tilemap layer position.

## Example Test Code

![image](https://user-images.githubusercontent.com/5847011/112943944-bd159f00-9164-11eb-881c-d3ca829a3efd.png)

photonstorm:
Thanks for opening this issue, and for submitting a PR to fix it. We have merged your PR into the `master` branch and attributed the work to you in the [Change Log](https://github.com/photonstorm/phaser/blob/master/CHANGELOG.md). If you need to tweak the code for whatever reason please submit a new PR.

danh121097:
Hi @moJiXiang, how to show correct the position of tilemap in Scene canvas?

debone:
@danh121097 I think @photonstorm link to the change log is not yet updated because the change was merged to the to-be-released `3.60` branch. The change log for the latest `3.60` is https://github.com/photonstorm/phaser/releases/tag/v3.60.0-beta.10. You can try it out the beta release at `npm i phaser@3.60.0-beta.10`. When loading tilesets you will be able to use the `tileoffset` parameter.

PR #5637, opened 2021-04-01
fix tilemap draw offset, maybe it is a bug, related issue #5633

moJiXiang:

* Fixes a bug

Describe the changes below:

Maybe issue #5633 will be fix.


photonstorm:
I need to make some changes to this PR to fix eslint issues, but otherwise, it looks good, so merging in, thank you.

PR #5639, opened 2021-04-02
Fix incorrect key when loading a Video with a config object

mattjennings:

* Fixes a bug

Describe the changes below:

Fixes an issue where videos would not get the correct key when loaded with a config object.

i.e

```js
this.load.video({ key: 'my-video', url: '/video.mp4' })
```

the key would end up being `{ key: 'my-video', url: '/video.mp4' }`. This PR would correctly use `my-video` as the key.


photonstorm:
👍

PR #5641, opened 2021-04-03
Fix scene PRE_UPDATE handlers removed at scene shutdown

samme:
This PR

* Fixes a bug

If you added a listener to scene [PRE_UPDATE](https://photonstorm.github.io/phaser3-docs/Phaser.Scenes.Events.html#event:PRE_UPDATE) it would get removed when the scene was shut down (e.g., restarted).

It happened because UpdateList#shutdown was unregistering `undefined` so inadvertently clearing the event.


photonstorm:
👍

PR #5656, opened 2021-04-15
Improve performance of MeasureText

valadaptive:
This PR

* Improves performance

Describe the changes below:

1. The `MeasureText` function previously called `getImageData` twice, once to see if it worked(?) and once to actually use the result.

    `getImageData` is quite slow on Chrome:
image

    By only calling it once, this PR should improve text measurement performance.

2. `metrics.hasOwnProperty('actualBoundingBoxAscent')` would return false even on browsers which supported these new text metrics properties, because `actualBoundingBoxAscent` is defined on its *prototype*. Changing the check to `'actualBoundingBoxAscent' in metrics` allows the fast path to actually be used.

### DIFF
--- a/src/gameobjects/text/MeasureText.js
+++ b/src/gameobjects/text/MeasureText.js
@@ -25,7 +25,7 @@ var MeasureText = function (textStyle)
 
     var metrics = context.measureText(textStyle.testString);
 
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

Issue #5647, opened 2021-04-06
PutTileAt function seem scale the sprite image to fit tile size


moJiXiang:
## Version
* Phaser Version: Phaser 3.54.0
* Operating system: Windows
* Browser: Chrome

## Description
When use putTileAt function to draw a tile in an empty tile,  it seems scale the terrain sprite image to fit tile size.

Sprite image size is width: 64, height: 44, Tile size is width: 62, height: 31.

So, it seems scale image height from 44 to 31.

## Additional Information

![202146-163339](https://user-images.githubusercontent.com/5847011/113682576-e2228880-96f5-11eb-810e-e409ff5bd8df.png)

moJiXiang:
Same problem with #5644

photonstorm:
Duplicate of #5644 - please track in there. Also, there was no code sample provided, so we couldn't have used this issue anyway (please remember this in the future).

PR #5657, opened 2021-04-15
fix PutTileAt tile size with releated tileset width and height, releated bug #5647

moJiXiang:
fix PutTileAt tile size with releated tileset width and height, releated bug #5647

### DIFF
--- a/src/tilemaps/components/PutTileAt.js
+++ b/src/tilemaps/components/PutTileAt.js
@@ -65,6 +65,16 @@ var PutTileAt = function (tile, tileX, tileY, recalculateFaces, layer)
     var newTile = layer.data[tileY][tileX];
     var collides = layer.collideIndexes.indexOf(newTile.index) !== -1;
 
+    // Copy properties from tileset to tiles.
+    var tiles = BuildTilesetIndex(layer.tilemapLayer.tilemap);
+    var index = tile instanceof Tile ? tile.index : tile;
+
+    var sid = tiles[index][2];
+    var set = layer.tilemapLayer.tileset[sid];
+
+    newTile.width = set.tileWidth;
+    newTile.height = set.tileHeight;
+
     SetTileCollision(newTile, collides);
 
     // Recalculate faces only if the colliding flag at (tileX, tileY) has changed
--- a/src/tilemaps/components/PutTileAt.js
+++ b/src/tilemaps/components/PutTileAt.js
@@ -8,6 +8,7 @@ var Tile = require('../Tile');
 var IsInLayerBounds = require('./IsInLayerBounds');
 var CalculateFacesAt = require('./CalculateFacesAt');
 var SetTileCollision = require('./SetTileCollision');
+var BuildTilesetIndex = require('../parsers/tiled/BuildTilesetIndex');
 
 /**
  * Puts a tile at the given tile coordinates in the specified layer. You can pass in either an index


photonstorm:
Hopefully, this doesn't have any side effects, but it's a small change so easy to undo if it does.

roboyeti:
> Hopefully, this doesn't have any side effects, but it's a small change so easy to undo if it does.

Well, it does have the side effect that if you set your tilemap to 32x32 and loaded 64x64 sized asset map into Tileset, it used to scale those assets down to fit the specified tilemap width x height.

I loaded up an old game that is used in a commercial project and we wanted to add some new features to with new things available in the PhaserJS lib, incremented to the first release after 3.55.2 and discovered my maps were a visual disaster (and then the particle system was entirely broken).  I was loading 64x64 assets to allow the game to render differently across devices, allowing more resolution on larger screens.

I poured thru the documentation and found no way with the Tilemap and related objects to force it to down scale the 64x64 to 32x32, as was the default behavior.  Even resizing the height and width of the Tile did absolutely nothing visually.  So far, the only solution appears to be to load a 32x32 Tileset.  Am I missing something?

Obviously this issue is old and people have already worked around it if it was a problem.  Reloading a 64x64 or 32x32 tileset isn't the end of the world for me, but ...

I find both of these changes (this and the particle interface) troubling for the following reasons:

1.> Typically, changing the default behavior of a single method or instantiation in a well used library should be the exception (something is broken), not the rule.  In the case of the Tileset => putTile(s)At calls, it now makes the Tileset the authority over the Tilemap and the Tilelayer... it is barely more than an image with slice dimensions.  A random check-in changed the default behavior and elevated the authority of the Tileset, while providing no ability to make the Tileset obey the rules of the Tilemap and Tilelayer OR a way to tell Tileset, slice by 64x64 cells, but scale to 32x32.

2.> If default behavior is going to be broken, and the default behavior wasn't explicitly an issue for most, then you probably should create a new method or add a parameter to delegate the dimensions from the Tileset to override those of the Tilemap.  Adding one boolean (defaulted to old behavior) to the parameters of putTilesAt()/putTileAt(), for example, would solve it and not break existing code.  The should have been the solution.  New issues (that maybe are trying something that wasn't even originally supported ... as it appears) shouldn't supersede old code bases by default.

3.> The particle emitter changes are concerning as well, for a somewhat minor release.  The new code should sit alongside the old code and a deprecation warning and eventual removal of the old interface (at a major version, say 4.0+) is probably the way to go.  The other option would have been to have the old creation method simply point at the new code for instantiation and return the new object type.

I highly recommend looking at how languages release changes, because they are usually a fine example of providing good releases that don't instantly break a bunch of existing code in the world.  I was using 3.55 and upgrading to 3.60 immediately broke everything, adding a lot of unexpected work (figuring out what happened, what options I had, and then making the changes in several places across many classes that inherited scene in our projects).

Anyway, didn't feel like it was "correct" to open a new issue for any of this, but just a reminder that when you change stuff like these two things, you frustrate people who are actually trying to use your library, more so for commercial projects, which I am.

photonstorm:
@roboyeti  3.60 was a massive release, not a minor release - it represented close to 2 years worth of development work and had a huge comprehensive [Change Log](https://github.com/phaserjs/phaser/blob/master/changelog/3.60/CHANGELOG-v3.60.md) to accompany it, that detailed the changes to particles (and elsewhere) in detail. 

You could argue it should have had a major version number, but that ship sailed a decade ago. We've never followed semver in that way, which is nothing new to anyone familiar with the framework. For us, the 3 is part of the name and the jump from 55 to 60 was a major increment (a jump from 55 to 55.2 is a minor one). It's the same all the way down and all the way up the version numbers.

PR #5683, opened 2021-05-05
GameObjectConfig allow type object

### LINKED ISSUE #5682: GameObjectConfig multiple types should allow `object`
## Version

* Phaser Version: v3.54.0
* Operating system: Any
* Browser: Any

## Description

When using Phaser 3.54.0 with TypeScript, the `GameObjectConfig` only allows `x`, `y`, and `alpha` to be of type number. However, an object is allowed.

Based on the [lab example](http://labs.phaser.io/edit.html?src=src/animation/create%20from%20sprite%20config.js&v=3.54.0), one can create a `SpriteConfig` to generate random `x` and `y` coordinates and `scale` per new sprite created.

## Example Test Code

```js
const config = {
    key: 'gems',
    x: { randInt: [ 0, 800 ] },
    y: { randInt: [ 0, 300 ] },
    scale: { randFloat: [ 0.5, 1.5 ] },
    anims: 'ruby'
};


//  Make 16 sprites using the config above
for (var i = 0; i < 16; i++)
{
    this.make.sprite(config);
}
```
[source](https://github.com/photonstorm/phaser3-examples/blob/969a0923ef17795b0e000d1148cb17fd28953cee/public/src/animation/create%20from%20sprite%20config.js#L29-L41)

While the code runs/executes, in TypeScript, a typedef error is produced.

```
Type '{ randInt: number[]; }' is not assignable to type 'number'
phaser.d.ts(62344, 17): The expected type comes from property 'y' which is declared here on type 'SpriteConfig'
```

## Possible Fix

In the type def of `GameObjectConfig` change `x`, `y`, and `alpha` to accept `object` as well.

karbassi:

* Updates the Documentation
* Fixes a bug

Describe the changes below:

GameObjectConfig x, y, rotation, angle, and alpha should allow type object

Fixes #5682

Thanks @chousiyan for raising the issue

### DIFF
--- a/src/gameobjects/typedefs/GameObjectConfig.js
+++ b/src/gameobjects/typedefs/GameObjectConfig.js
@@ -2,16 +2,16 @@
  * @typedef {object} Phaser.Types.GameObjects.GameObjectConfig
  * @since 3.0.0
  *
- * @property {number} [x=0] - The x position of the Game Object.
- * @property {number} [y=0] - The y position of the Game Object.
+ * @property {(number|object}} [x=0] - The x position of the Game Object.
+ * @property {(number|object}} [y=0] - The y position of the Game Object.
  * @property {number} [depth=0] - The depth of the GameObject.
  * @property {boolean} [flipX=false] - The horizontally flipped state of the Game Object.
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

Issue #5697, opened 2021-05-13
`Tilemap.createFromObjects` unexpected behavior for large images, tile data


lackhand:
## Version

* Phaser Version: HEAD (v3.55.0-dev-FB)

* Operating system: n/a; Mac OSX

* Browser:

## Description

Expected: calling `someTilemap.createFromObjects('someLayer', {gid:someId})` creates approximately the same map as seen in Tiled: ![Screen Shot 2021-05-13 at 4 15 38 PM](https://user-images.githubusercontent.com/207026/118182029-80e18980-b406-11eb-9580-1fc9a836f160.png)
.
Actual: Because it doesn't automatically link tile offsets to texture frames, you get a bunch of missing images: 
![Screen Shot 2021-05-13 at 3 31 27 PM](https://user-images.githubusercontent.com/207026/118182098-99ea3a80-b406-11eb-849d-b93d476dc83e.png)

This can be repaired by the user by calling `setTexture` (or `setFrame`) on the returned objects which is cumbersome...
BUT THEN:
Expected: calling `setTexture` on the objects returned from `createFromObjects` will create approximately the same map as seen in Tiled.
Actual: Object has size ratio with 32 *squared* -- a 4px object is drawn at .5 px, an 8px object at 2 px, 16px @ 8px, 32px @ 1:1, 64px drawn at 128px: 
![Screen Shot 2021-05-13 at 3 36 59 PM](https://user-images.githubusercontent.com/207026/118182187-b1c1be80-b406-11eb-83de-b6fe920f20d9.png)

This is due to Tilemap internally setting the objects to the empty frame (@ 32 pixels), which sets the scale (via `setDisplayWidth` and `Height`; when the image is then later set, the scale is not cleared.

Also:
Expected: `type` of object will be visible to phaser and useful for creating instances (since `type` sounds a lot like `class` to me!)
Actual: `createFromObjects` doesn't support a `type` selector. Even if it did, because `type` can be set either on the `object` itself or on the tile to which the object `gid` refers, though Tiled treats those as equivalent (objects inherit the `type` of their tile in the UI!), since it's not inlined in the exported data, Phaser doesn't refer to it. (you can see this in the tiled UX screenshot where `64` is of type `huge`, contrast with the data in the included zip (see below)...

These are somewhat related; in order to use default image information from tiled, phaser would need to follow the object gid back to the tileset. It doesn't currently.

## Example Test Code

```js
class extends Phaser.Scene {
      constructor() {
        super({
[...]

PR #5698, opened 2021-05-13
Addresses https://github.com/photonstorm/phaser/issues/5697

lackhand:
This diff adds:
* `type` support to `createFromObjects`. See https://doc.mapeditor.org/en/stable/manual/custom-properties/#typed-tiles for the description in Tiled; this is currently not supported as a parameter, and its usage is documented within the diff.
* `createFromObjects` will also examine the tileset. As per the above link, created objects are supposed to inherit properties from the tile with the same gid, but Phaser doesn't do that right now. After this diff it optionally will, by passing an additional parameter to `createFromObjects`.
> NOTE: I'd be very happy to make this a default behavior, but that's somewhat backwards breaking
* `createFromObjects` will attempt to use the same texture and frame as Tiled when creating the object. This will behave poorly if there is no such frame (for instance, if the texture were loaded with `this.load.image` instead of `this.load.spritesheet`, or using different geometry. However, the current behavior draws missingimage, so this is a strong improvement.


photonstorm:
👍

PR #5701, opened 2021-05-14
Line-Line Intersection fix if dx1 === 0

Skel0t:
In the previous version, if x1 == x2 the code was never able to detect an intersection even if there was one.

This limitation was because of the unnecessary division by dx1. With its removal it should work now.

* Fixes a bug


photonstorm:
Thanks, this has come up a few times, so going to merge this and then I'll fix the ESLint issues locally.

Issue #5377, opened 2020-10-27
Video doesn't resume playing after regaining focus


spayton:
## Version

* Phaser Version: from 3.20 -> current dev

* Operating system: win10 x64

* Browser: Desktop chrome 86

## Description

A playing video pauses when swapping to another browser tab but doesn't resume when swapping back

## Example Test Code

To reproduce, simply visit the [labs video example](http://labs.phaser.io/view.html?src=src\game%20objects\video\play%20video.js), then switch to another tab on same browser, then back.  the video will have stopped playing.

## Additional Information

It is noticed that on regaining focus, an attempt is made to resume video playback in globalResume() but is prevented from doing so because _codePaused is set.  The _codePaused property is set true on creation and never cleared,  I'd guess that it should be cleared on playback.

photonstorm:
Thanks for opening this issue, and for submitting a PR to fix it. We have merged your PR into the `master` branch and attributed the work to you in the [Change Log](https://github.com/photonstorm/phaser/blob/master/CHANGELOG.md). If you need to tweak the code for whatever reason please submit a new PR.

PR #5704, opened 2021-05-17
Fixed video not resuming on regaining focus

VektreX:
This PR

* Fixes a bug

Describe the changes below:

As described in issue #5377 the video would not resume on regaining focus unless setting `setPaused(false)` on the object.
Fixed by setting the `_codePaused` attribute back to `false` in the play handler methods.


photonstorm:
👍

PR #5723, opened 2021-05-31
Turn off registered context-restored event

rexrainbow:

* Fixes a bug

Describe the changes below:

CONTEXT_RESTORED event has been registered but won't be turn off when respond game object has been destroyed, thus the game object won't be released fully.

[Bitmask](https://github.com/photonstorm/phaser/blob/master/src/display/mask/BitmapMask.js#L147) also registers CONTEXT_RESTORED event without turn off it. But I did not fix it in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Version 3.56.0 - Miku - in development
+
+### Bug Fixes
+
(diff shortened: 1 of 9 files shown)


rexrainbow:
Oops, I will create another PR for this bug fix

Issue #5685, opened 2021-05-06
Web Audio does not work on iOS Safari


albert-gonzalez:
## Version

* 3.54.0

* iOs

* Safari

## Description

Using Safari in iOS, web audio does not work. When you run phaser it throws an error like this:

```
Error: Audio key "theKey" missing from cache
```

If you disable the audio with enabling the noAudio config, it works. You also can enable disableWebAudio, but this causes other issues.

## Example Test Code

```js
new Phaser.Game({
  type: Phaser.AUTO,
  physics: {
    default: "arcade",
  },
  audio: {
    noAudio: true // if this is true, it works, with audio enabled it fails
  },
  scene: titleSceneConfig,
});
```

## Additional Information

In this thread you can see other people having the same issue: https://phaser.discourse.group/t/audio-works-on-web-but-not-ios/3677

Thank you for this awesome project!

Best regards!

photonstorm:
That error means you've tried to load an audio format that iOS doesn't support.

Load a correct one, like an m4a, and it definitely works.

albert-gonzalez:
I am using ogg, which works in all other browsers. I'll try with m4a and let you know the result.

Thanks for the quick response!

photonstorm:
iOS doesn't support ogg files.

albert-gonzalez:
You will always learn something new 🙂. I'll convert my files to m4a and try again. I close the issue!

Thank you for your support!

photonstorm:
You should provide your audio in 2, ideally 3 formats: m4a, ogg and mp3, then you provide an array of these 3 file types to the Phaser Audio Loader and it'll pick whichever is the first one it can play for that device.

albert-gonzalez:
Great, I'll do that. Thank you for the tip!

photonstorm:
This issue has been mentioned on **Phaser**. There might be relevant details there:

https://phaser.discourse.group/t/audio-works-on-web-but-not-ios/3677/9

Dercetech:
Hi,
I'm currently trying to write a proper "ios + capacitor" audio loader.

@photonstorm , I don't know why, but this is only when the mp3 are loaded from ios AND served by ios (using capacitor to build the package) that "my" issue happens. Using fetch instead of xhr solves this though. Maybe I can book 2h of your time via the support plan so we can work this out together?

Dercetech:
[...]

PR #5725, opened 2021-06-02
Fix Capacitor iOS file load issue

consolenaut:

Fixes a bug
Describe the changes below:

This fix addresses a similar issue to #5685, an issue with loading files when a Phaser project is wrapped in a Capacitor native app shell on iOS.
Capacitor on iOS often serves local files with a capacitor:// custom scheme, so just checking file:// leads to strange behaviour, extending the check to add capacitor:// resolves this issue.

We ran into this when using spine animations, whereby when Phaser tried to get a file URL, rather than getting something similar to `capacitor://myapp/images/file.png` it would get `capacitor://myapp/capacitor://myapp/images/file.png`. This is because Phaser checks to see if the URL is either blob/file/http/https, in which case it uses the raw URL, but if not prepends a baseURL before the URL assuming its a relative path.  This PR adds capacitor:// to the protocol check to prevent malformed double-url issue.

### DIFF
--- a/src/loader/File.js
+++ b/src/loader/File.js
@@ -86,7 +86,7 @@ var File = new Class({
         {
             url = loader.path + loadKey + '.' + GetFastValue(fileConfig, 'extension', '');
         }
-        else if (typeof url === 'string' && !url.match(/^(?:blob:|data:|http:\/\/|https:\/\/|\/\/)/))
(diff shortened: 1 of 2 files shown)


consolenaut:
@photonstorm Any news on this one? Happy to do any re-work thats required, otherwise we'll have to for Phaser to hit our product release deadline

photonstorm:
@consolenaut thank you for the PR. I will merge it into the master branch now, but there is unlikely to be another new Phaser release this month, so you'll need to create your own build from master to use.

PR #5740, opened 2021-06-11
feat: Add support for load image by HTMLImageElement

hanzooo:

* Adds a new feature

Describe the changes below:

Add support for load image by HTMLImageElement.
We use phaser in our project, it is great!  And we find that load image by HTMLImageElement (without a blob process) is a lightweight way, which improve the performance. So I think we can support this feature within the phaser, and add a new config to use it.

### DIFF
--- a/src/core/Config.js
+++ b/src/core/Config.js
@@ -518,6 +518,11 @@ var Config = new Class({
          */
         this.loaderWithCredentials = GetValue(config, 'loader.withCredentials', false);
 
+        /**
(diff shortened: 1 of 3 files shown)


photonstorm:
Sure, this makes a lot of sense. Happy to merge this. I'll need to complete all the JSDocs that were missed, but that's fine, will do it now.

hanzooo:
Amazing, thanks a lot.👍

PR #5742, opened 2021-06-12
Turn off context restore event when game object is destroyed

rexrainbow:

* Fixes a bug

Describe the changes below:

It will cause a memory leakage if context-restore-event is not turned off when Text/TileSprite game object are destroyed. Reference of these game objects will be kept in event system.

[Bitmask](https://github.com/photonstorm/phaser/blob/master/src/display/mask/BitmapMask.js#L147) also registers CONTEXT_RESTORED event without turn off it. But I did not fix it in this PR.

### DIFF
--- a/src/gameobjects/text/Text.js
+++ b/src/gameobjects/text/Text.js
@@ -287,10 +287,7 @@ var Text = new Class({
             this.setLineSpacing(style.lineSpacing);
         }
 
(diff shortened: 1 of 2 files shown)


samme:
Great minds think alike 😄

photonstorm:
👍

photonstorm:
Thanks. I've done the Bitmap Mask one too.

PR #5748, opened 2021-06-16
Use default hitArea if hitArea parameter is not given

rexrainbow:

* Fix a bug

Describe the changes below:

Default hitArea will resize when [`zone.setSize(w, h)`](https://github.com/photonstorm/phaser/blob/master/src/gameobjects/zone/Zone.js#L172-L176).

This PR also can fix issue in [this phaser.io example](https://labs.phaser.io/view.html?src=src\input\zones\resize%20zone.js)


photonstorm:
👍

PR #5750, opened 2021-06-18
assign array by index for skipped chunks

jonnytest1:
loading breaks for layers that skip chunks with no data

* Fixes a bug

Describe the changes below:

infinite maps can omit chunks that dont have data so we need to assing the chunk sub array by index


jonnytest1:
@photonstorm @consolenaut @rexrainbow

photonstorm:
👍

PR #5767, opened 2021-06-30
Fix getBounds for creating MatterTileBody for staggered tiles

### LINKED ISSUE #5764: convertTilemapLayer for isometric tilemaps not setting bodies properly
## Version

* Phaser Version: 3.55.2

* Operating system: macOS

* Browser: 

## Description

I'm having an issue with setting up collisions with a staggered isometric tilemap created in Tiled. I ran across [this discussion](https://phaser.discourse.group/t/setcollisionbyproperty-not-working-the-way-i-think-it-does/2642) outlining the same issue, and I can confirm it's still a problem in 3.55.2. Basically, `convertTilemapLayer` does not set the collision bodies for tiles correctly on isometric maps, neither the default rectangle nor the shape specified in the collision layer.

As suggested in the thread, the issue comes from `getBounds()` on the `Tile` class. Specifically, `this.getLeft()` always returns `undefined` for any of the tiles.

```javascript
    getBounds: function (camera, output)
    {
        if (output === undefined) { output = new Rectangle(); }

        output.x = this.getLeft();
        output.y = this.getTop();
        output.width = this.getRight() - output.x;
        output.height = this.getBottom() - output.y;

        return output;
    }
```

## Example Test Code

```typescript
    const map = this.make.tilemap({ key: 'map' })
    const tileset = map.addTilesetImage('tileset', 'tiles')
    const ground = map.createLayer('Floor', tileset)

    // property "collides" is set on certain tiles within Tiled
    ground.setCollisionByProperty({ collides: true })

    this.matter.world.convertTilemapLayer(ground)
```

## Additional Information

I'm happy to give solving this a shot myself if you could provide some guidance. 🙂

mayacoda:

This PR

* Fixes a bug

Describe the changes below:
- Use `StaggeredTileToWorldXY` to calculate x value for `Tile.GetBounds`
- Fix typo in parameter name
- Fix `StaggeredTileToWorldY` so it properly aligns colliders to the tilemap

Fixes #5764

### DIFF
--- a/src/tilemaps/components/StaggeredTileToWorldY.js
+++ b/src/tilemaps/components/StaggeredTileToWorldY.js
@@ -32,7 +32,7 @@ var StaggeredTileToWorldY = function (tileY, camera, layer)
         tileHeight *= tilemapLayer.scaleY;
     }
 
-    return layerWorldY + tileY * (tileHeight / 2);
+    return layerWorldY + tileY * (tileHeight / 2) + tileHeight;
 };
 
 module.exports = StaggeredTileToWorldY;
--- a/src/tilemaps/Tile.js
+++ b/src/tilemaps/Tile.js
@@ -404,7 +404,12 @@ var Tile = new Class({
     {
         var tilemapLayer = this.tilemapLayer;
 
-        return (tilemapLayer) ? tilemapLayer.tileToWorldX(this.x, camera) : this.x * this.baseWidth;
+        if (tilemapLayer)
+        {
+            var point = tilemapLayer.tileToWorldXY(this.x, this.y, undefined, camera);
+            return point.x;
+        }
+        return this.x * this.baseWidth;
     },
 
     /**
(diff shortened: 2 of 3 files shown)


src/tilemaps/components/StaggeredTileToWorldY.js
```diff
@@ -32,7 +32,7 @@ var StaggeredTileToWorldY = function (tileY, camera, layer)
         tileHeight *= tilemapLayer.scaleY;
     }
 
-    return layerWorldY + tileY * (tileHeight / 2);
+    return layerWorldY + tileY * (tileHeight / 2) + tileHeight;
```
mayacoda, line 35: I'm not entirely sure why this was necessary. Without the change, all colliders were set half a tile height above where they should have been.

With the change (note how the rectangle encompasses the correct tile for the tall grass and the custom collider is at the bottom of the wall tiles)
image

Without the change
image

photonstorm:
Thanks for the PR. I've merged it for now and will do some further testing. It's possible that extra height is required for Arcade Physics, but isn't for Matter.

photonstorm:

https://phaser.discourse.group/t/isometric-collision-matter-converttilemaplayer-not-working/11538/1

PR #5788, opened 2021-07-20
Add Crossorigin support

rmartell:

* Adds a new feature

Describe the changes below:

When using the loadURL method on video, it was not possible to specify a crossorigin request, so it was impossible to load videos from non-same domain locations. Note that this is only when streaming - if you use the loader, crossorigin support was builtin.

I considered using the same methodology as the loader, where it pulls from the sceneconfig on init and allows override, but that I think that would muddle the water in terms of "who has precedence" (loader or video?). Since this is only an issue on loadURL, IMHO made the most sense to add it as a 4th parameter.


photonstorm:
👍

PR #5793, opened 2021-07-28
Update body center after tile separation

samme:
This PR

* Fixes a bug

Like #5344, `Arcade.Body#center` was incorrect in the collision callback after separation with a tile.


photonstorm:
👍

PR #5797, opened 2021-07-30
Fix WEBGL-only selection and throw for unknown renderer type

### LINKED ISSUE #5583: WEBGL still falls back to CANVAS
## Version

* Phaser Version: 3.52.0

## Description

Create a game with type `Phaser.WEBGL`. Run it on a device without WebGL. The game runs in CANVAS mode.

[The docs say](https://photonstorm.github.io/phaser3-docs/Phaser.html#.WEBGL "Phaser.WEBGL"):

> Forces Phaser to use the WebGL Renderer. If the browser does not support it, there is no fallback to Canvas with this setting, so you should trap it and display a suitable message to the user.

But this doesn't happen. In this case WEBGL seems to act just like AUTO.

webgl disabled

## Example Test Code

In Firefox you can set `webGL.disabled` to `false` in `about:config` and then visit any WEBGL example, e.g., <http://phaser.io/examples/v3/view/display/alpha/single-alpha>. The game runs in CANVAS mode with no error.

Or you can do

```js
new Phaser.Game({
    type: Phaser.WEBGL,
    backgroundColor: 0x0074D9,
    callbacks: {
        preBoot: function (game) {
            game.device.features.webGL = false;
        }
    },
    scene: {
        create: function () {
            this.add.text(0, 0, 'Hello');
        }
    }
});
```

samme:
This PR

* Adds a new feature
* Fixes a bug

Fixes #5583

Fixed: When the device does not support WebGL, creating a game with `{ type: Phaser.WEBGL }` will fail with an error. It will not fall back to `CANVAS` as `AUTO` does.

New: Creating a game with an invalid renderer `type` will throw an error. Before it would be treated like `AUTO`.

### DIFF
--- a/src/core/CreateRenderer.js
+++ b/src/core/CreateRenderer.js
@@ -31,22 +31,22 @@ var CreateRenderer = function (game)
     //  Not a custom environment, didn't provide their own canvas and not headless, so determine the renderer:
     if (!config.customEnvironment && !config.canvas && config.renderType !== CONST.HEADLESS)
     {
-        if (config.renderType === CONST.CANVAS || (config.renderType !== CONST.CANVAS && !Features.webGL))
+        if (config.renderType === CONST.AUTO)
         {
-            if (Features.canvas)
-            {
-                //  They requested Canvas and their browser supports it
-                config.renderType = CONST.CANVAS;
-            }
-            else
-            {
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

PR #5819, opened 2021-08-27
Add DOMElement#preRender, remove DOMElement#preUpdate

### LINKED ISSUE #5816: setVisible not working for DOM elements in a container
## Version

* Phaser Version: 3.55.2

* Operating system: Ubuntu

* Browser: Tested on Chrome and Firefox

## Description

It seems like the fix in https://github.com/photonstorm/phaser/issues/4486 is no longer working. When the visibility of the parent container is changed, the DOM node is still visible. I am able to determine that it is because `preUpdate` is not being called on DOM elements, though I am not sure why.

## Example Test Code

```
class MainScene extends Phaser.Scene {
    create() {
        const text = new Phaser.GameObjects.DOMElement(this, 0, 0, "span", { color: "white" }, "Hello");
        const container = this.add.container(50, 50, [text]);
        this.time.delayedCall(1000, () => {
            console.log("Setting container visibility to false");
            container.setVisible(false);
            // If we forcefully call preupdate on the next update
            // then the text correctly disappears
            // this.events.once("update", () => {
            //     text.preUpdate();
            // });
        });
    }
}
```

See a demo: https://codepen.io/prakol16/pen/rNwVyvv

## Additional Information

In addition, perhaps the `preUpdate` code should be moved to a `preStep` event on the game. The reason for this is because if a scene is paused, the visibility of various objects can still change, but `preUpdate` will never be called. Similarly, if the container/its children are changed to be inactive with `setActive(false)`, we'd expect the visibility to still be controlled separately by `setVisible`, but instead, the visibility of the DOM element will no longer change.

samme:
This PR

* Fixes a bug

Removes `DOMElement#preUpdate()` and replaces it with `DOMElement #preRender()`, for checking parent visibility. And DOMElement is no longer added to the scene update list.

Purpose is to fix #5816 and improve DOMElement's visibility behavior. This way DOMElement will respond to its parent's visibility even if the scene is paused or DOMElement has `active = false`.

Fixes #5816

### DIFF
--- a/src/gameobjects/domelement/DOMElement.js
+++ b/src/gameobjects/domelement/DOMElement.js
@@ -303,18 +303,7 @@ var DOMElement = new Class({
 
         scene.sys.events.on(SCENE_EVENTS.SLEEP, this.handleSceneEvent, this);
         scene.sys.events.on(SCENE_EVENTS.WAKE, this.handleSceneEvent, this);
-    },
-
-    //  Overrides Game Object method
-    addedToScene: function ()
-    {
-        this.scene.sys.updateList.add(this);
-    },
-
-    //  Overrides Game Object method
-    removedFromScene: function ()
-    {
-        this.scene.sys.updateList.remove(this);
+        scene.sys.events.on(SCENE_EVENTS.PRE_RENDER, this.preRender, this);
     },
 
     /**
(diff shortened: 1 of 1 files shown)


photonstorm:
Sounds good to me. DOM Elements, the double-edge swords that never stop slicing.

PR #5825, opened 2021-09-04
Fix #5811, log error on invalid scene.start(key)

### LINKED ISSUE #5811: An error should be thrown when an undefined scene is started
## Version

* Phaser Version: v3.55.2

* Operating system: MacOS

* Browser: Chrome

## Description

When starting a scene with an invalid key, the command just kills the current scene with no error shown.

https://github.com/photonstorm/phaser/blob/5c8ecbcf999e6f328d21884e877c9e5935d2d350/src/scene/SceneManager.js#L1154

Seems there is no handling of the negative case of that `if`.

## Example Test Code

When no such scene was registered in the GameConfig:

`this.scene.start("game-over");`

## Additional Information

Would a PR to console log/warn be considered?

ubershmekel:

* Adds a new feature
* Fixes a bug

Describe the changes below:

When starting a scene with an invalid key, the user doesn't know that it failed to change scenes. This change calls `console.warn` to let the user know that such a problem occurred. Also, I switch the logic of the `if` statement to reduce the indentation level of the code.

Fixes #5811

### DIFF
--- a/src/scene/SceneManager.js
+++ b/src/scene/SceneManager.js
@@ -1151,53 +1151,56 @@ var SceneManager = new Class({
 
         var scene = this.getScene(key);
 
-        if (scene)
+        if (!scene)
         {
-            var sys = scene.sys;
-
-            //  If the Scene is already running (perhaps they called start from a launched sub-Scene?)
-            //  then we close it down before starting it again.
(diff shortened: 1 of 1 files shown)


photonstorm:
Sure, why not,  makes sense!

PR #5835, opened 2021-09-14
Docs: correct types in Animation#calculateDuration

### LINKED ISSUE #5833: Typescript optional / null arguments
## Description

Some functions handle null input but the typing does not reflect that.

## Example

[the documented function used](https://newdocs.phaser.io/docs/3.55.2/Phaser.Animations.Animation#calculateDuration)

function signature: `calculateDuration(target: Phaser.Animations.Animation, totalFrames: number, duration: number, frameRate: number): void`

inside this function, we have:
```js
if (duration === null && frameRate === null)
// ...
else if (duration && frameRate === null)
// ...
else
```

the documentation does not say that we choose duration or frameRate through typing `number | null`, with a default if none given.

### Example Test Code

```js
// let get a created animation named 'key'
const anim = scene.anims.get('key')

// to make the animation named 'key' last 2s:
anim.calculateDuration(anim, anim.getTotalFrames(), 2000, null)

// to make the animation named 'key' run at 8 frames per secondes:
anim.calculateDuration(anim, anim.getTotalFrames(), null, 8)
```

the correct function signature would be: `calculateDuration(target: Phaser.Animations.Animation, totalFrames: number, duration: number | null, frameRate: number | null): void`

## Additional Information

Moreover, we could make change in the code to allow undefined value too.

I don't know how to change the code to reflect typing change, it's look like typing is generated trough documentation, `@param {number | null}` would do ? Does these is something to run to generate then the typescript before sending a pull request ?

I would like to send one if it's ok, with the code change to allow `undefined`, and have the correct typing.

samme:
This PR

* Updates the Documentation

Fixes #5833

### DIFF
--- a/src/animations/Animation.js
+++ b/src/animations/Animation.js
@@ -224,8 +224,8 @@ var Animation = new Class({
      *
      * @param {Phaser.Animations.Animation} target - The target to set the values on.
      * @param {number} totalFrames - The total number of frames in the animation.
-     * @param {number} duration - The duration to calculate the frame rate from.
-     * @param {number} frameRate - The frame ate to calculate the duration from.
+     * @param {?number} duration - The duration to calculate the frame rate from.
+     * @param {?number} frameRate - The frame rate to calculate the duration from.
      */
     calculateDuration: function (target, totalFrames, duration, frameRate)
     {


photonstorm:
Thanks for this PR, it was missing the [] on the params, so I've pushed my own fix for it.

PR #5836, opened 2021-09-15
Docs: correct types in Arcade.Body#setBounce

Pythux:
This PR

* Updates the Documentation

Describe the changes below:

y argument is optional, will be equal to x if not given

### DIFF
--- a/src/physics/arcade/Body.js
+++ b/src/physics/arcade/Body.js
@@ -1899,7 +1899,7 @@ var Body = new Class({
      * @since 3.0.0
      *
      * @param {number} x - The horizontal bounce, relative to 1.
-     * @param {number} y - The vertical bounce, relative to 1.
+     * @param {?number} y - The vertical bounce, relative to 1.
      *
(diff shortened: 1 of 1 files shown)


src/physics/arcade/Body.js
```diff
@@ -1899,7 +1899,7 @@ var Body = new Class({
      * @since 3.0.0
      *
      * @param {number} x - The horizontal bounce, relative to 1.
-     * @param {number} y - The vertical bounce, relative to 1.
+     * @param {?number} y - The vertical bounce, relative to 1.
```
photonstorm, line 1902: Thank you for your PR. The way to define is a parameter is optional is to put the parameter name in square braces. So this line should be: `@param {number} [y=x] - The vertical bounce, relative to 1.`

photonstorm:
Thanks for your PR. If you want to update and fix this, I'll merge it, otherwise let me know and I'll just make the change locally.

photonstorm:
Don't worry, I've fixed this locally as I found it in several other places, too.

PR #5840, opened 2021-09-20
fix: add resizeInterval type

hanzooo:
This PR

* Fixes a bug

add resizeInterval type defined in the Phaser.Core.Config, to pass the type check.


photonstorm:
Hi - Could you please tell me what this fixes? As `resizeInterval` is already declared [here](https://github.com/photonstorm/phaser/blob/master/src/core/Config.js#L92) and [here](https://github.com/photonstorm/phaser/blob/master/src/core/typedefs/ScaleConfig.js#L15), so as a result is already in the TypeDefs on lines 5457 and 60495.

hanzooo:
> Hi - Could you please tell me what this fixes? As `resizeInterval` is already declared [here](https://github.com/photonstorm/phaser/blob/master/src/core/Config.js#L92) and [here](https://github.com/photonstorm/phaser/blob/master/src/core/typedefs/ScaleConfig.js#L15), so as a result is already in the TypeDefs on lines 5457 and 60495.

Hi, I found that we can set it in two ways，[Config.resizeInterval] (https://newdocs.phaser.io/docs/3.54.0/Phaser.Core.Config#resizeInterval) or Config.scale.resizeInterval. I use the first one, but it fail to pass the typescript type check because it is not declared in the [here](https://newdocs.phaser.io/docs/3.54.0/Phaser.Types.Core#GameConfig)

hanzooo:
> Hi - Could you please tell me what this fixes? As `resizeInterval` is already declared [here](https://github.com/photonstorm/phaser/blob/master/src/core/Config.js#L92) and [here](https://github.com/photonstorm/phaser/blob/master/src/core/typedefs/ScaleConfig.js#L15), so as a result is already in the TypeDefs on lines 5457 and 60495.

Hi, I found that we can set it in two ways，[Config.resizeInterval](https://newdocs.phaser.io/docs/3.54.0/Phaser.Core.Config#resizeInterval) or Config.scale.resizeInterval. I use the first one, but it fail to pass the typescript type check because it is not declared in the [GameConfig type](https://newdocs.phaser.io/docs/3.54.0/Phaser.Types.Core#GameConfig)

photonstorm:
Ok, have made this change locally.

PR #5842, opened 2021-09-23
Ignore scene stop if already shutdown

### LINKED ISSUE #5826: Shutdown shouldn't be repeated for stopped scenes
If you do

```js
this.scene.stop().stop().stop();
```

then [SHUTDOWN](https://newdocs.phaser.io/docs/3.55.2/Phaser.Scenes.Events.SHUTDOWN) is emitted three times.

This seems a little incorrect to me and it can also make shutdown event handlers needlessly complex.

I would rather have [SceneManager#stop](https://newdocs.phaser.io/docs/3.55.2/Phaser.Scenes.SceneManager#stop) abort if the scene is already stopped.

samme:
This PR

* Adds a new feature

Closes #5826

### DIFF
--- a/src/scene/SceneManager.js
+++ b/src/scene/SceneManager.js
@@ -1217,7 +1217,7 @@ var SceneManager = new Class({
     {
         var scene = this.getScene(key);
 
-        if (scene && !scene.sys.isTransitioning())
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

Issue #5520, opened 2021-01-23
Error stopping a scene: cannot read property 'destroy' of undefined


schontz:
## Version
* Phaser Version: 3.5.1

* Operating system: Mac and Windows

* Browser: Chrome

## Description

Shutting down a scene causes an error:

```
TypeError: Cannot read property 'destroy' of undefined
    at DisplayList.shutdown (phaser.js:163448)
    at EventEmitter.emit (phaser.js:1928)
    at Systems.shutdown (phaser.js:47907)
    at SceneManager.stop (phaser.js:99258)
```

The error in `DisplayList` is here:
https://github.com/photonstorm/phaser/blob/master/src/gameobjects/DisplayList.js#L234

I believe the issue is that some of the objects being destroyed are in turn destroying some sub-objects, which would be on `this.list`, thus making the list shorter.

This could be fixed like so:

```js
while(list.length) {
  list[0].destroy(true);
}
```

That would be unaffected by changes to the list along the way.

schontz:
I did a little more digging and it seems my original hunch is right. I have something like the following:

```js
class MySprite extends Phaser.Physics.Arcade.Sprite {
  constructor() {
    this.label = this.scene.add.text(10, 20, 'hello');
  }

  destroy() {
    this.label.destroy();
  }
}
```

When I trace `DisplayList.shutdown` I get something like so:

```
- shutdown: destroy (i = 7) = SomeObject
  - SomeObject.destroy
- shutdown: destroy (i = 6) = AnotherObject
  - AnotherObject.destroy
- shutdown: destroy (i = 5) = MySprite
  - MySprite.destroy
  - TextLabel.destroy
- shutdown: destroy (i = 4) = undefined because TextLabel was already destroyed.
```

Looking at the base [GameObject.destroy](https://github.com/photonstorm/phaser/blob/master/src/gameobjects/GameObject.js#L710), you can see that destroying removes the item from the display list, thus naively using `i` in shutdown is not a good option. I am currently getting around the issue by manually calling the code in my initial comment before shutdown runs.

daneren2005:
I am running into the same issue and for the same reason.  It is really helpful to be able to let objects manage their own dependencies without needing special logic to handle scene destruction separately.

pt-hieu:
any update on this? I've been having this problem and it's like I don't have any `destroy` or `remove` called in my code

photonstorm:
[...]

PR #5848, opened 2021-10-01
Allow objects to destroy other objects when the DisplayList is shutting down

astei:
This PR fixes a bug (#5520).

Describe the changes below:

This PR fixes `DisplayList.shutdown()` to allow users to destroy other game objects whilst another object is being destroyed without tripping up. Objects are destroyed in reverse order for full compatibility.

While checking to see if `fromScene` is `true` before destroying any objects that could be on the Display List would also fix the issue, doing so is not concise.

### DIFF
--- a/src/gameobjects/DisplayList.js
+++ b/src/gameobjects/DisplayList.js
@@ -231,14 +231,12 @@ var DisplayList = new Class({
     {
         var list = this.list;
 
-        var i = list.length;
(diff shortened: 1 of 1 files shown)


photonstorm:
Thanks for this PR. We have implemented a slightly different approach but that resolves the same problem this PR also fixed.

PR #5851, opened 2021-10-02
Warn when there is a multiFile load failure

ubershmekel:
I spent more than an hour debugging this. The animation wasn't working.
I eventually found that I was loading an invalid URL as an Aseprite texture.

* Adds a new feature

Describe the changes below:

Warn when one file fails to load out of a multiFile. E.g. when a aseprite texture fails to load because of an invalid URL.

### DIFF
--- a/src/loader/LoaderPlugin.js
+++ b/src/loader/LoaderPlugin.js
@@ -892,8 +892,8 @@ var LoaderPlugin = new Class({
                 }
                 else
                 {
-                    var srcs = file.multiFile.files.map(function(it) { return it.src }).join(', '));
-                    console.warn("failed to load multiFile into cache: " + srcs;
+                    var srcs = file.multiFile.files.map(function(it) { return it.src; }).join(', ');
+                    console.warn("failed to load multiFile into cache: " + srcs);
                 }
             }
             else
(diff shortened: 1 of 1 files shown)


samme:
I think there are no arrow functions in Phaser 3.

photonstorm:
You can't use ES6 in Phaser 3 I'm afraid.

Also, an invalid URL would clearly show up on the Dev Tools Network panel. The only time it wouldn't is if you did actually get the file, but it had invalid contents (then I would expect a runtime error from the parser tbh)

ubershmekel:
@photonstorm my case was a vite dev server. I was debugging an aseprite animation not loading and saw my JSON file loaded correctly in the network panel, but the JSON file was not in the cache so my anims was empty. Turns out the image URL resulted in an html file. Because I was loading an aseprite JSON+PNG multfile, I did not see a warning for why my JSON was not in the cache. I wasn't even looking for an image loading issue, but that's what wrecked my JSON loading.

I removed the arrow function in favor of a regular function.

photonstorm:
I merged your PR locally and it will now print a `console.warn` for every MultiFile Phaser loads, even if it's successful, because if it loads say a PNG and a JSON, it'll warn after loading the PNG because `isReadyToProcess()` returns false for the first file. Sorry, as it stands, I cannot merge this.

PR #5855, opened 2021-10-03
Fix rendering of Tiled object layers on isometric maps

lfarroco:
This PR:

* Fixes a bug

Describe the changes below:
Objects contained in object layers generated by Tiled use orthogonal positioning even when the map is isometric. In the pictures below, note that `A` has `{x:128, y:128}` and `B` has `{x:256, y:128}` - the `y` is the same, as one would expect in an orthogonal grid (but not in an isometric one).

image
image

The current way to position sprites on `createFromObjects` uses that position provided by Tiled and applies an origin offset over it - leading to issues [as described here](https://phaser.discourse.group/t/objects-position-are-wrong-in-isometric-map-created-with-tiled/8771).

This PR converts the sprites' position from cartesian to isometric before applying the offset, positioning elements over the grid as expected.

### DIFF
--- a/scripts/copy-to-examples.js
+++ b/scripts/copy-to-examples.js
@@ -58,11 +58,11 @@ if (fs.existsSync(dest))
 
         const options = {
             path: './src',
-            extensions: [ 'js' ]
+            extensions: [ '.js' ]
         };
-
+ 
         sloc(options).then((res) => {
-            console.log('Source files: ' + res.files + '\nLines of code: ' + res.sloc);
+            console.log('Source files: ' + res.sloc.files + '\nLines of code: ' + res.sloc.sloc);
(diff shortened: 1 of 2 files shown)


photonstorm:
Thanks for the PR - can you please confirm this PR will work correctly with the current changes already made to `createFromObjects` as part of 3.60?

lfarroco:
Hey! I installed locally from `master` and added the changes, it works as expected.
I also added a ratio calculation, because the original PR assumed a ratio of tile width/height of 2, but that can vary.

photonstorm:
Thanks 👍

lfarroco:
Thanks for running such an amazing project, @photonstorm !

PR #5858, opened 2021-10-07
Add static Lerp method on Vector2

### LINKED ISSUE #5684: add static method for lerp on a Vector
**When performing the lerp function on a Vector2 instance the result of that operation wil be used to overwrite the original value of that Vector2's instance. 
Meaning the lerp function is not immutable.  And after reading the documentation it is not supposed to be.
So that is why I am not posting this as a bug.

Here is the thing though; I usually do not want to lose the original Vector2's value. So have to remember to clone the original Vector2 every time a interpolation is needed. That tends to create hard to find bugs.  Or at least decrease readability of the code.

The Geom classes have static methods for things like GetPoint, even though you can also do .getPoint operations on an instance object. And as far as I know the latter does not really mutate the instance member vars. 

Could not find such a static method for interpolation, well there is Phaser.Math.Linear but that does not return a new Vector2, or even Point;

So why not include a static method for interpolation that takes 2 Vector2 in a similar fashion as GetPoint does for geom classes and returns a brand new Vector2 as the result? 

I include here a demo to show what i mean:

https://codepen.io/hendrikras/pen/abvePYd?editors=0011

GregDevProjects:

* Adds a new feature

Describe the changes below:

Adds a static method to the Vector2 class so that two vectors can be lerped (if that's a word 😛 ) without modifying the original Vectors. 

Resolves #5684.

### DIFF
--- a/src/math/LinearXY.js
+++ b/src/math/LinearXY.js
@@ -1,0 +1,30 @@
+/**
+ * @author       Greg McLean <GregDevProjects>
+ * @copyright    2021 Photon Storm Ltd.
+ * @license      {@link https://opensource.org/licenses/MIT|MIT License}
+ */
+
+/**
+ * Interpolates two given Vectors and returns a new Vector between them.
+ *
+ * Does not modify either of the passed Vectors.
+ *
+ * @function Phaser.Math.LinearXY
+ * @since 3.6.0
+ *
+ * @param {Phaser.Math.Vector2} vector1 - Starting vector
+ * @param {Phaser.Math.Vector2} vector2 - Ending vector
+ * @param {number} [t=0] - The percentage between vector1 and vector2 to return, represented as a number between 0 and 1.
+ *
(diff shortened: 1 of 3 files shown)


samme:
Since [Vector2](https://newdocs.phaser.io/docs/3.55.2/Phaser.Math.Vector2) has no static methods maybe the new method could go on Phaser.Math instead, like [Phaser.Math.RandomXY()](https://newdocs.phaser.io/docs/3.55.2/Phaser.Math#RandomXY)?

GregDevProjects:
@samme sounds good, there's a [Linear static method](https://newdocs.phaser.io/docs/3.55.2/Phaser.Math#Linear) on the Math class that does the same thing with a Number, so I named this function `LinearXY` - the `XY` is intended to specify that a Vector2 will be returned.

Any chance you'd be willing to add the `hacktoberfest-accepted` label to this PR? 😅

photonstorm:
👍

PR #5861, opened 2021-10-08
Add new `getNormal()` and `getNormalAt()` methods to `Curve` and `Path` classes

natureofcode:
This PR

* Adds a new feature

Describe the changes below:

_As Pomax wrote on the Bezier.js website,_
> In 2d, the normal is simply the normalised tangent vector, rotated by a quarter turn.

### DIFF
--- a/src/curves/Curve.js
+++ b/src/curves/Curve.js
@@ -490,6 +490,46 @@ var Curve = new Class({
     },
 
     /**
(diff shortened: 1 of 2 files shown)


natureofcode:
Saw that there are already methods `normalizeRightHand()` and `normalizeLeftHand()`, so additional methods that do the same thing even with one step less will be overkill. Close.

Issue #5863, opened 2021-10-11
Timer with very short delay


rexrainbow:
## Version

* Phaser Version: 3.60-beta.1 , or prevoius

* Operating system: win10

* Browser: Chrome, firefox

## Description

In the [test](https://codepen.io/rexrainbow/pen/abybrbw),
- `delay` is set to `1` (1 ms)
- `repeat` is set to `499`, therefore timer will run 500 times
- This timer will be stopped after 500 ms

The callback will count the times of time-out, and shown at a text game object.
I will expected that the counter will be `500` after 600ms.
It is because that [callback only invoked one time](https://github.com/photonstorm/phaser/blob/master/src/time/Clock.js#L377) in each tick. In a very short delay case, callback will be invoked many times, imo.

photonstorm:
Thanks for opening this issue, and for submitting a PR to fix it. We have merged your PR into the `master` branch and attributed the work to you in the [Change Log](https://github.com/photonstorm/phaser/blob/master/CHANGELOG.md). If you need to tweak the code for whatever reason please submit a new PR.

PR #5864, opened 2021-10-11
Invoke callback many times in very short delay case

rexrainbow:

* Fixes a bug

Describe the changes below:

#5863


rexrainbow:
Note that I did not check `event.hasDispatched` in this PR, it might cause other problems.

photonstorm:
This should be ok - `hasDispatched` should be false at the end of this loop anyway, to give the event one last chance to run (as you never run repeat count down to zero, only to 1)

Issue #5731, opened 2021-06-06
ParticleEmitter: particleClass property expects to receive a Phaser.Particle instance


gastispin:
## Version

* Phaser Version: 3.54.0

* Operating system: Any

* Browser: Any

## Description

The property `particleClass` within the class `ParticleEmitter` expects a to receive a Phaser.Particle instance, although it should **specify the type of class to use for such particle.**
This is particularly a problem when trying to animate a particle with multiple frames, as seen in [this example](http://phaser.io/examples/v3/view/game-objects/particle-emitter/custom-particles)

## Example Test Code

The issue can be found in the line 23730 in the phaser definitions file (_phaser.d.ts_)

` particleClass: Phaser.GameObjects.Particles.Particle;`

## Additional Information

The quickest workaround is just to modify the defitinions file and expect _any_, but I guess it should be officially (and properly) fixed.

`particleClass: any;`

miker2049:
I ran into this earlier, looking at the source, it doesn't look like it expects an _instance_, but rather a constructor.  Looking specifically [here](https://github.com/photonstorm/phaser/blob/0bf366a09686e16d62e996310e9110bb64325a86/src/gameobjects/particles/ParticleEmitter.js#L1934), the ParticleEmitter calls `new` on that property to make a particle, so the property is not instanced yet actually.  

Seems to me the proper type for this property would be something like this:
```js
...
particleClass: new (emitter) => Particle
...
```

Based on typescript [docs](https://www.typescriptlang.org/docs/handbook/2/classes.html#abstract-construct-signatures) for similar situation.

samme:
We had the same problem with [Group#classType](https://newdocs.phaser.io/docs/3.55.2/Phaser.GameObjects.Group#classType).

I think JSDoc `@callback` with the correct params may be the solution.

miker2049:
Oh I guess I need to be thinking in JSDoc terms, not TS.  Then why not just `@class`/`@constructor`?  Seems like it will do [what we want](https://jsdoc.app/tags-class.html), assuming that the type definition generation follows that ok.

photonstorm:
Sadly `@class` cannot be used as a data type as it's a tag.

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

PR #5868, opened 2021-10-13
Change the type for particleClass

miker2049:

* Fixes a bug

Describe the changes below:

The `particleClass` property of a `ParticleEmitter` needs to be a constructor, not an instanced class.

(For a type change like this I am not sure if I should regenerate the type definitions, but just putting this out here)

A possible fix for #5731


miker2049:
Closing because of #5870

PR #5869, opened 2021-10-13
Print a console error when file processing fails

### LINKED ISSUE #5862: Warn for file processing failures
I reproduced #5851. If you load an aseprite animation from an invalid image (e.g. wrong type or zero-length file), the files are complete, the animations are missing, but there are no other errors.

Generally I think the problem is that most file process failures are silent, which is a little confusing for users.

https://codepen.io/samme/pen/yLoLJVE?editors=0010

load aseprite with invalid image

samme:
This PR

* Adds a new feature

Closes #5862 

I used `console.error()` but I could change it to `console.warn()` if that's preferred.

### DIFF
--- a/src/loader/File.js
+++ b/src/loader/File.js
@@ -412,6 +412,7 @@ var File = new Class({
      */
     onProcessError: function ()
     {
+        // eslint-disable-next-line no-console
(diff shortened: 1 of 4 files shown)


photonstorm:
`console.error` is fine, it won't halt the JS and is accurate.

PR #5879, opened 2021-10-19
Docs: mark Scene config optional

### LINKED ISSUE #5874: Scene config should be optional.
image

Following the Phaser examples you don't have to do that. I think all configs are optional. But the argument itself is required. This is one of the most common mistakes made with TypeScript. You have to either mark the argument as optional or set a default (empty) object. 

**Solution A**

```ts
class Scene {
  constructor(config?: string | Phaser.Types.Scenes.SettingsConfig)
}
```

I think this is what you want / how the code works. You have to handle nullish values (undefined).

**Solution B**

```ts
class Scene {
  constructor(config: string | Phaser.Types.Scenes.SettingsConfig = {}) // or string in this case
}
```

`phaser: '3.55.2'`

Trivia: You don't need the `constructor` if you don't use it. Empty constructors which only contains super or nothing can be omitted.

Btw. a question: What do you think of rewrite Phaser in TypeScript or modern JavaScript?

samme:
This PR

* Updates the Documentation

Fixes #5874

### DIFF
--- a/src/scene/Scene.js
+++ b/src/scene/Scene.js
@@ -18,7 +18,7 @@ var Systems = require('./Systems');
  * @constructor
  * @since 3.0.0
  *
- * @param {(string|Phaser.Types.Scenes.SettingsConfig)} config - Scene specific configuration settings.
+ * @param {(string|Phaser.Types.Scenes.SettingsConfig)} [config] - The scene key or scene specific configuration settings.
  */
 var Scene = new Class({
 


ghost:
Great, but does this also affect the definition files (d.ts)? Are the d.ts files created from the doc comments?

samme:
They are generated from the docs at release time.

PR #5889, opened 2021-10-30
Fix PathFollower end point when given `to`

samme:
This PR 

* Fixes a bug

If you started a path follower with a `to` value, the follower would complete at the end of the path instead of the `to` value.


photonstorm:
👍

PR #5918, opened 2021-11-05
Allow Shape game object to be a mask

rexrainbow:

* Adds a new feature

Describe the changes below:

Allow Shape game object to be a mask #5900

### DIFF
--- a/src/gameobjects/components/Mask.js
+++ b/src/gameobjects/components/Mask.js
@@ -123,13 +123,13 @@ var Mask = {
      * @method Phaser.GameObjects.Components.Mask#createGeometryMask
      * @since 3.6.2
      *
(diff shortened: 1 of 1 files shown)


rexrainbow:
Is it better to remove type checking of Graphics/Shape game object? Therefore user can apply any kind of game object to be mask.

photonstorm:
Hmm - I'm happy to remove the type check, but then I swear we'll get people trying to dump sprites or whatever in there and wondering why it doesn't work. Rarely do they read the docs, or use TypeScript, which would enforce it.

rexrainbow:
> 
> 
> Hmm - I'm happy to remove the type check, but then I swear we'll get people trying to dump sprites or whatever in there and wondering why it doesn't work. Rarely do they read the docs, or use TypeScript, which would enforce it.

I see, I will keep type checking there. Thanks for suggestion.

photonstorm:
👍

PR #5921, opened 2021-11-08
fix: The window touchcancel event is not removed

teng-z:

* Fixes a bug
On mobile platforms, frequent create destroy games cause GPU memory leaks


photonstorm:
👍

PR #5935, opened 2021-12-03
Allow render settings to be specified at top level AND inside .render

vzhou842:
This PR fixes a bug with render config.

### Context

If you pass _anything_ for the `render` field in `GameConfig`, Phaser will currently ignore any render-related fields supplied at the top level of the config object. For example:

```
{
  pixelArt: true,
  render: {},
}
```

ends up with `pixelArt: false` because an object was passed to the `render` field, so the top-level `pixelArt: true` is ignored.

This is unintuitive (a bug?) and also does not match the behavior of the `scale` (`ScaleConfig`) field, which respects top-level fields if they aren't overridden in the `scale` object.

### This PR

Respects top-level render fields the same way it works for `ScaleConfig` right now:
- first pull the top-level field, e.g. `this.pixelArt = GetValue(...)`
- then, if `render` is supplied, override fields present on the render object, e.g. `this.pixelArt = GetValue(renderConfig, 'pixelArt', this.pixelArt)

### DIFF
--- a/src/core/Config.js
+++ b/src/core/Config.js
@@ -338,43 +338,112 @@ var Config = new Class({
 
         //  Renderer Settings
         //  These can either be in a `render` object within the Config, or specified on their own
-
-        var renderConfig = GetValue(config, 'render', config);
+        //  The former overrides the latter
 
         /**
          * @const {Phaser.Types.Core.PipelineConfig} Phaser.Core.Config#pipeline - An object mapping WebGL names to WebGLPipeline classes. These should be class constructors, not instances.
          */
-        this.pipeline = GetValue(renderConfig, 'pipeline', null);
+        this.pipeline = GetValue(config, 'pipeline', null);
 
         /**
          * @const {boolean} Phaser.Core.Config#antialias - When set to `true`, WebGL uses linear interpolation to draw scaled or rotated textures, giving a smooth appearance. When set to `false`, WebGL uses nearest-neighbor interpolation, giving a crisper appearance. `false` also disables antialiasing of the game canvas itself, if the browser supports it, when the game canvas is scaled.
          */
-        this.antialias = GetValue(renderConfig, 'antialias', true);
+        this.antialias = GetValue(config, 'antialias', true);
 
         /**
          * @const {boolean} Phaser.Core.Config#antialiasGL - Sets the `antialias` property when the WebGL context is created. Setting this value does not impact any subsequent textures that are created, or the canvas style attributes.
          */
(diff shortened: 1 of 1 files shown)


src/core/Config.js
```diff
@@ -338,95 +338,119 @@ var Config = new Class({
...
+            this.preserveDrawingBuffer = GetValue(renderConfig, 'preserveDrawingBuffer', this.preserveDrawingBuffer);
+            this.premultipliedAlpha = GetValue(renderConfig, 'premultipliedAlpha', this.premultipliedAlpha);
+            this.failIfMajorPerformanceCaveat = GetValue(renderConfig, 'failIfMajorPerformanceCaveat', this.failIfMajorPerformanceCaveat);
+            this.powerPreference = GetValue(renderConfig, 'powerPreference', this.powerPreference);
+            this.batchSize = GetValue(renderConfig, 'batchSize', this.batchSize);
+            this.maxTextures = GetValue(renderConfig, 'maxTextures', this.maxTextures);
+            this.maxLights = GetValue(renderConfig, 'maxLights', this.maxLights);
+        }
+
+        if (this.pixelArt)
```
vzhou842, line 448: moved here from above

photonstorm:
Thanks for the PR. I would like to find another, more compact way to do this, as it adds lots of lines of additional code to achieve what is a bit of an edge-case feature.

vzhou842:
@photonstorm what's the reasoning for supporting this for `scale` but not for `render` though? Seems like we should be consistent here, also +24 net lines does not feel particularly egregious to me (am i missing some context here around phaser trying to aggressively cut code size or anything?)

even if somehow there is a strong argument for keeping the status quo (don't support this for `render` but support it for `scale`) i feel like there needs to be a documentation improvement around this b/c this inconsistency is quite confusing. I'm happy to investigate that and try to mention this in documentation somewhere if you think that would help

vzhou842:
gonna bump this once - see my comment above^. I'm of the opinion this PR is a pretty straightforward small win, but even if ultimately there's a good reason to have different behavior for `scale` vs `render` there should probably be a mention of this in documentation somewhere. I'm happy to do the work for this

photonstorm:
There are lots of config options that cannot be overridden via the top level (in fact, all of them - input, dom, etc) with the exception of scale. Providing an empty object should not reset top-level render settings, however. I have modified the `GetValue` function so it can accept 2 sources for the property and updated the Config file to reflect this, which removes lots of lines of code and has the same end result (including what this PR set out to achieve).

vzhou842:
cool thanks!

PR #5936, opened 2021-12-04
Remove GetBounds component from PointLight

### LINKED ISSUE #5934: Reset a point light that has a body takes a lot of time
## Version

* Phaser Version: 3.55.2

* Operating system: Win10

* Browser:

## Description

First create a static group, and add a lot of PointLight, and then calling 'body.reset', the time consumed is increasing as the number of calls increases. I don't know if this is the correct usage!

## Example Test Code

```
create(){
      const group = this.physics.add.staticGroup();

      for (let i = 0; i < 20; i++) {
         for (let j = 0; j < 20; j++) {
            const p = new PointLight(this, i * 40 + 100, j * 40 + 100, 0xff0000, 20, Math.random())
               .addToDisplayList();
            group.add(p);
         }
      }

    const pl = new PointLight(this, 20, 20, 0xff0000, 20, 1)
         .addToDisplayList();
    group.add(pl);

    // Test
    setInterval(() => {
       console.time('body.reset');
       // this.pl.body.enable = false;
       // this.pl.body.enable = true;
       pl.body.reset(30, 30);
       console.timeEnd('body.reset');
    }, 1000);
}
```

samme:
This PR

* Fixes a bug

All of the [GetBounds](https://newdocs.phaser.io/docs/3.55.2/Phaser.GameObjects.Components.GetBounds) methods will fail on PointLight, because it's missing the required properties.

Closes #5934

### DIFF
--- a/src/gameobjects/pointlight/PointLight.js
+++ b/src/gameobjects/pointlight/PointLight.js
@@ -69,7 +69,6 @@ var PointLight = new Class({
         Components.AlphaSingle,
         Components.BlendMode,
         Components.Depth,
-        Components.GetBounds,
         Components.Mask,
         Components.Pipeline,
         Components.ScrollFactor,


photonstorm:
👍

PR #5937, opened 2021-12-05
fix sceneManager moveBelow and moveAbove

EmilSV:
Because the index of A and B is stored before B is remove, with splice, the code need to take into account that the index of A decrement if B was below it


photonstorm:
👍

PR #5944, opened 2021-12-15
Remove `path` from package.json dependencies

antkhnvsk:
This PR
* Fixes a bug

Describe the changes below:
Removing an unnecessary library dependency
* which replaces the native nodejs library 
* was last updated 6 years ago https://www.npmjs.com/package/path
* breaks other libraries in project during build that use require('path')

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "imports-loader": "^3.1.1",
     "jsdoc": "^3.6.7",
     "node-sloc": "^0.2.1",
+    "path": "^0.12.7",
     "remove-files-webpack-plugin": "^1.5.0",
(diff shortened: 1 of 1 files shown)


photonstorm:
I'm happy to remove this, but it will break our TS Defs generator if I do. So a PR will have to replace our use of that (so it still works) first.

artaommahe:
Can we just move this lib to the devDependencies so it would not be installed when used in other projects?

antkhnvsk:
I've changed this PR to move`path` to devDependencies. If that is not applicable, I will make another fix.

photonstorm:
That's fine, I'm happy with this change as it'll still work for defs gen.

Issue #5945, opened 2021-12-19
window.webkitAudioContext does not have argument on iOS 10


jcyuan:
![image](https://user-images.githubusercontent.com/850854/146659266-4f01e76d-3e9e-43e0-b4a1-dfcbc05fe9c2.png)

on IP6, iOS10.0.1, the `WebAudioSoundManager` failed to initialize, the browser thinks that you pass an unnecessary parameter to it.

error "TypeError: Not enough arguments"....  actually too many argument though...

after remove the parameter, all works fine:

![image](https://user-images.githubusercontent.com/850854/146659301-f5e49301-e645-4f20-840a-d520431ac60e.png)

maybe better has some way to detect if the API has an argument? if no then create without argument, otherwise create just like the current solution with `{ latencyHint: 'interactive' }` object?

photonstorm:
It might work under a try catch? But iOS10 is seriously old now.

jcyuan:
> It might work under a try catch? But iOS10 is seriously old now.

yep agree, just our proj has a serious compatible requirement... 🙁
except this, all others work very well till now. no issues found on such old devices. 👍✔

photonstorm:
Thanks for opening this issue, and for submitting a PR to fix it. We have merged your PR into the `master` branch and attributed the work to you in the [Change Log](https://github.com/photonstorm/phaser/blob/master/CHANGELOG.md). If you need to tweak the code for whatever reason please submit a new PR.

PR #5949, opened 2021-12-21
fix iOS10 webkitAudioContext failed to initialize

jcyuan:

* Fixes a bug

fix for #5945


photonstorm:
Merged but I removed the console log, doesn't feel like we need it.

PR #5953, opened 2021-12-22
destroy game objects during shutdown safely #5520

schontz:

This PR

* Fixes a bug #5520

Describe the changes below:
- Since `destroy` removes the game object from the `list`, enumerating through the list is not safe. Instead, we can simply destroy all the game objects until the list is empty.


schontz:
I could not find any automated test suite.

photonstorm:
Thanks for the PR. I'll test this locally.

PR #5956, opened 2021-12-23
Revert "Skip intersects check by argument"

samme:
This PR

* Fixes a bug

d80e4f5 was wrong — the `intersects()` check should not be skipped — but because it includes a logical mistake it didn't cause any problems. Nevertheless I want to get rid of it.

So this removes the optional `intersects` argument in `Phaser.Physics.Arcade.World#separate()`. Phaser no longer uses it and it never worked so this isn't really a breaking change.


photonstorm:
👍

PR #5957, opened 2021-12-23
Docs: correct type for TileSpriteConfig.frame

### LINKED ISSUE #5955: `frame` of TileSpriteConfig should accept `number` in TypeScript
## Version
* Phaser Version: 3.55.2
* Browser: Google Chrome 96.0.4664.110

## Description
In phaser.d.ts, the `frame` of `TileSpriteConfig` type is defined as `string`. But `frame` parameter should also accept `number`.

## Example Test Code

In create() of Scene
```
this.make.tileSprite({
  x: 0,
  y: 0,
  width: 32,
  height: 32,
  key: "keyOfTexture",
  frame: 0,  <<- type error!
})
```

## Additional Information

GameObjectFactory accepts both `string` and `number` for `frame` parameter.
So below code is working.
```
this.add.tileSprite(0, 0, 32, 32, "keyOfTexture", 0)
```

Also `// @ts-ignore` is a workaround for `this.make` way.

samme:
This PR

* Updates the Documentation

Fixes #5955


photonstorm:
👍

PR #5958, opened 2021-12-24
Add setTint() method in Mesh GameObject

rexrainbow:

* Adds a new feature

Describe the changes below:

Add `setTint()` method, `tint` setter in Mesh GameObject.
Which will set `color` property of all vertices of this Mesh Game Object.


photonstorm:
Sure, why not :)

PR #5968, opened 2022-01-06
Fix `pointToContainer` issue with `output` parameter

vforsh:
* Fixes a bug

Describe the changes below:
Provided `output` parameter was ignored when the container didn't have a parent container (was attached directly to the scene)


photonstorm:
👍

PR #5977, opened 2022-01-24
Fix: `convertTilemapLayer` creating composite bodies unintentionally

EddieCameron:

* Fixes a bug

Describe the changes below:

There is an edge case that could cause a crash when using `convertTilemapLayers` to create Matter.js bodies from a Tilemap. If any tiles had multiple colliders _and_ you were providing some body creation options, the `parts` property in the options would be modified and then concatenated with any bodies created after it. This would mean that some tiles would be combined when they shouldn't be, and on large maps would eventually hang once the convex hull got too big/complex.

This PR just clones the `options` object before modifying it. You may already have a favoured way of doing this.

### DIFF
--- a/src/physics/matter-js/MatterTileBody.js
+++ b/src/physics/matter-js/MatterTileBody.js
@@ -235,8 +235,9 @@ var MatterTileBody = new Class({
         }
         else if (parts.length > 1)
         {
-            options.parts = parts;
-            this.setBody(Body.create(options), options.addToWorld);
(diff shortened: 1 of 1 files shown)


samme:
Maybe [Clone()](https://newdocs.phaser.io/docs/3.55.2/Phaser.Utils.Objects#Clone) instead.

photonstorm:
Thanks for this. I have implemented your fix and credited you in the Change Log.

Issue #5986, opened 2022-01-28
Calling tween.stop(0) runs for an additional delta before stopping, so tween isn't truly 100% "reset"


Mesonyx:
**EDIT:** I created a [PR](https://github.com/photonstorm/phaser/pull/5987) that seems to fix this issue

--------------

## Version

* Phaser Version: 3.5.5
* Operating system: OSX Mojave 10.14.6
* Browser: Google Chrome

## Description

When calling `stop()` on a Tween with a `resetTo` param of **0** (indicating you wish to revert the tween back to its pre-tweened state), the tween actually runs for a single additional delta before stopping. 

I noticed this because I'm using tweens to animate the mouths of NPC's in my game using `scaleX` and `scaleY` tweens. Although imperceptible at first, their mouths noticeably shrunk in size during a long enough run-time, despite me reverting the tween to the supposed default state with `stop(0)` when they were done talking.

## Example Test Code

I've created a Code pen of the issue [here](https://codepen.io/Mesonyx/pen/WNXvLWX)

Clicking toggles the tween on and off with `stop(0)`. 

After starting/stopping about a dozen times, the box is definitely smaller when compared with the reference shape immediately below it.

I've also output the `scaleX` of the object being tweened, so we can definitively verify that the box is not actually resetting to its full, pre-tween `scaleX` value of 1.

## Additional Information

I believe this behavior occurs because the `seek()` method in the `Tween.js` class runs with a `do/while` loop, so the condition is checked *after* the tween iterates a single delta (line [972](https://github.com/photonstorm/phaser/blob/master/src/tweens/tween/Tween.js#L972)).

From my understanding of how Tweens work from the docs/reading the code, would it work to simply skip this loop entirely if the  the `toPosition` is equal to **0**?

photonstorm:
Thanks for opening this issue, and for submitting a PR to fix it. We have merged your PR into the `master` branch and attributed the work to you in the [Change Log](https://github.com/photonstorm/phaser/blob/master/CHANGELOG.md). If you need to tweak the code for whatever reason please submit a new PR.

PR #5987, opened 2022-01-28
Bugfix: Tween Reset Fix

Mesonyx:
This PR:
* Fixes a bug

Describe the changes below:
A fix for #5986 

Because the `do/while` loop iterates one time even when `toPosition` is 0, calling `stop(0)` doesn't fully reset the tween to the expected position, instead reverting to zero and iterating the tween by a single delta. Individually, this isn't noticeable, but over time the single-delta changes add up.

Instead, if `toPosition` is 0, the delta-iteration loop is now skipped entirely.

I've verified that `tween.stop()` still works as expected with no `setTo` parameter as well as *0* and non-zero parameters. Additionally,  `tween.play()`, `tween.pause()`, `tween.resume()` and `tween.restart()` all still function as expected.

### DIFF
--- a/src/tweens/tween/Tween.js
+++ b/src/tweens/tween/Tween.js
@@ -967,13 +967,16 @@ var Tween = new Class({
             this.state = TWEEN_CONST.ACTIVE;
         }
 
-        this.isSeeking = true;
-
-        do
+        if (toPosition > 0)
         {
-            this.update(0, delta);
+            this.isSeeking = true;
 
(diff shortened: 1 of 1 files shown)


photonstorm:
I don't see any immediate harm in doing this but will test the Tween examples after this merge. My only worry is that if you don't call `update` at least once, I don't think the Tween seek will actually reset its targets at all. It just won't do anything.

photonstorm:
All looks good so far. Will keep this merged for the next beta!

PR #5989, opened 2022-01-31
Bugfix: hitAreaCallback is not a function

ArifZx:

* Fixes a bug

Describe the changes below:

Screen Shot 2022-01-31 at 09 38 45

Screen Shot 2022-01-31 at 09 40 02

### DIFF
--- a/src/input/InputManager.js
+++ b/src/input/InputManager.js
@@ -971,7 +971,7 @@ var InputManager = new Class({
         y += gameObject.displayOriginY;
 
         var input = gameObject.input;
-        var hitAreaCallback = input?.hitAreaCallback;
+        var hitAreaCallback = input && input.hitAreaCallback;
 
         if (typeof hitAreaCallback === 'function' && hitAreaCallback(input.hitArea, x, y, gameObject))
         {
--- a/src/input/InputManager.js
+++ b/src/input/InputManager.js
@@ -971,8 +971,9 @@ var InputManager = new Class({
         y += gameObject.displayOriginY;
 
         var input = gameObject.input;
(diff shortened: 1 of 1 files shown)


src/input/InputManager.js
```diff
@@ -971,8 +971,9 @@ var InputManager = new Class({
         y += gameObject.displayOriginY;
 
         var input = gameObject.input;
+        var hitAreaCallback = input?.hitAreaCallback;
```
natureofcode: Unfortunately, **Optional chaining operator (?.)** is not **ES5**-compliant, so it can't be used in Phaser 3 code.

samme:
What's the bug?

src/input/InputManager.js
ArifZx: got it noted and updated

ArifZx:
> What's the bug?

sometimes hitAreaCallback is not a function

samme:
Sometimes how? How do you produce the bug?

ArifZx:
> Sometimes how? How do you produce the bug?

Still don't know to reproduce that, but found causes hitAreaCallback is undefined
Screen Shot 2022-02-02 at 16 30 40

Maybe after clear interactive, but the object still in queue (_list)

samme:
I just don't see how it could happen. In [clear()](https://newdocs.phaser.io/docs/3.55.2/Phaser.Input.InputPlugin#clear) `gameObject.input` is set to null.

photonstorm:
Sorry but I'm going to close this until an example can be provided showing how it fails. The error is quite clear IMHO, in that a function wasn't provided, or that you cleared the input and then still tried to call an input event. Please provide an example, thanks.

PR #5997, opened 2022-02-08
Remove the load complete handlers when stopping a scene

### LINKED ISSUE #5726: Pause and stop scene do not work during preload
## Version

* Phaser Version: 3.55.2

* Operating system: Ubuntu

* Browser: any

## Description

When calling `scene.pause` on a scene during its preload:

**Expected**: Either the loading pauses, or the loading continues but `create` is not called until the scene is resumed, or the loading continues and `create` is called, but after the `create` call, the scene goes back to being paused, and `update` is not called.

**Actual**: The pause is ignored.
<hr />

When calling `scene.stop` on a scene during its preload:

**Expected**: The scene stops, and even if the loading finishes, `create` should not be called.

**Actual**: `create` is called which usually results in an error. However, this can lead to unpredictable behavior. For example, if you add a timer in `create`, suddenly, if you ever restart the scene, there will be two timers, since the old timer was added after the scene was destroyed.

## Example Test Code

Example code: https://codepen.io/prakol16/pen/dyvmXJP

Note that this uses the `rexAwait` loader plugin, but that is just so that I can simulate a long-loading file. The bug is there even when loading ordinary files.

## Solutions
Probably something like checking if the scene has been destroyed/paused before calling `create` in `loadComplete` in `SceneManager.js`.

samme:
This PR

* Fixes a bug

If you did, e.g.,

    this.scene.start('scene').stop('scene');

with a scene with a load queue, it would shut down, then call create, then change to running.

Fixes #5726

### DIFF
--- a/src/scene/SceneManager.js
+++ b/src/scene/SceneManager.js
@@ -1221,6 +1221,11 @@ var SceneManager = new Class({
 
         if (scene && !scene.sys.isTransitioning() && scene.sys.settings.status !== CONST.SHUTDOWN)
         {
+            var loader = scene.sys.load;
+
+            loader.off(LoaderEvents.COMPLETE, this.loadComplete, this);
+            loader.off(LoaderEvents.COMPLETE, this.payloadComplete, this);
+
             scene.sys.shutdown(data);
         }
 


photonstorm:
👍

PR #6000, opened 2022-02-11
Fixed WebGL blending that was broken in 3.50. Fix #5565, fix #5996

### LINKED ISSUE #5565: Blend modes are not set on WebGL Graphics primitives drawn to a RenderTexture
## Version
* Phaser Version:  v3.52.0 
* Operating system: Windows 10 Version 1909 (OS Build 18363.1379)
* Browser: Firefox 85.0.2 (64-bit)

## Description

Under WebGL, calling setBlendMode() on a Graphics object has no effect when drawing that Graphics object to a RenderTexture.

The workaround is to add the Graphics object to a Container, and draw that Container to the RenderTexture.

This came up whilst drawing primitives to a Graphics object using additive blending, then drawing that RenderTexture to the scene using multiplicative blending, in order to implement a light/shadow effect.

## Example Test Code

See [https://codepen.io/danarcher/pen/OJbWYeW](https://codepen.io/danarcher/pen/OJbWYeW). The code is drawing concentric circles using additive blending.

Set `workaround` to `true` to see the workaround in effect, whereby concentric circles of increasing (additive) brightness are visible.

The workaround isn't necessary when using the canvas rather than WebGL renderer (change `Phaser.WEBGL` to `Phaser.CANVAS` in the example to demonstrate).

## Additional Information

There is no call to the renderer to set the blend mode from [GraphicsWebGLRenderer](https://github.com/photonstorm/phaser/blob/v3.52.0/src/gameobjects/graphics/GraphicsWebGLRenderer.js) or [RenderTextureWebGLRenderer](https://github.com/photonstorm/phaser/blob/v3.52.0/src/gameobjects/rendertexture/RenderTextureWebGLRenderer.js).

There is a call to the renderer to set the blend mode from [ContainerWebGLRenderer](https://github.com/photonstorm/phaser/blob/v3.52.0/src/gameobjects/container/ContainerWebGLRenderer.js#L99).

### LINKED ISSUE #5996: Blending broken using RenderTextures & WebGL
## Version
* Phaser Version: 3.55.2
* Browser: Chrome

## Description
Additive blending is not working when rendering to a `RenderTextures` when using `WebGL`.

## Example Test Code

Minimal sample with 2 circles that should be blended: https://codepen.io/sjb933/pen/wvPgpWq?editors=1010

**You can fix the blending by either:**

1. Changing `USE_WEBGL` to `false` will activate canvas rendering which blends as expected.
2. Use a version of `Phaser <3.50`. You can uncomment the second script line in the HTML file to import older `Phaser@3.24.1`

I'm including the same sample code below
```javascript
// With webgl disabled (canvas mode), blending works regardless of Phaser version
// With webgl enabled, blending works in Phaser < 3.50, and doesn't work in latest
const USE_WEBGL = true;

let config = {
  type: USE_WEBGL ? Phaser.WEBGL : Phaser.CANVAS,
  width: 600,
  height: 400,
  scene: {
    create: create,
  }
};

let game = new Phaser.Game(config);

function create() {
  rt = this.add.renderTexture(0, 0, 600, 400);
  
  layer1 = createLayer(this, 200, 200, 0xff0000);
  layer2 = createLayer(this, 400, 200, 0x0000ff);
  
  // draw the layers with additive blending
  layer1.setBlendMode(Phaser.BlendModes.ADD);
  layer2.setBlendMode(Phaser.BlendModes.ADD);
  rt.draw(layer1);
  rt.draw(layer2);
}

function createLayer(self, x, y, color) {
  let graphics = self.make.graphics();
  graphics.fillStyle(color);
  graphics.fillCircle(x, y, 150);
  return graphics;
}
```

sjb933:
This small PR fixes a WebGL blending bug that was introduced in v3.50.0.

**Fixes https://github.com/photonstorm/phaser/issues/5565 and https://github.com/photonstorm/phaser/issues/5996**

### Detail
It looks like when the `renderDirect()` method was introduced, the WebGL blending was mistakenly removed. 
- You can see the [original code from the previous version 3.24.1](https://github.com/photonstorm/phaser/blob/v3.24.1/src/gameobjects/rendertexture/RenderTexture.js#L985-L988)
- I reintroduced this code in case of non-direct rendering

### DIFF
--- a/src/gameobjects/rendertexture/RenderTexture.js
+++ b/src/gameobjects/rendertexture/RenderTexture.js
@@ -1167,6 +1167,10 @@ var RenderTexture = new Class({
         }
         else
         {
+            if (!this._eraseMode)
+            {
+                this.renderer.setBlendMode(gameObject.blendMode);
+            }
             gameObject.renderWebGL(this.renderer, gameObject, this.camera);
         }
 


sjb933:
Found the code was removed in this commit: https://github.com/photonstorm/phaser/commit/e03c514b0313b68b8da1d05b3a14664b066dfe38 which seems to be in error.

The commit reports to fix erase mode, but the relevant code that was removed in this commit only applies to non-erase mode.

photonstorm:
👍

PR #6003, opened 2022-02-12
Add enableCapture to createCursorKeys

codergautam:
This was an annoying issue for me, I was using createCursorKeys, and I also had a html textbox, and I couldn't figure out why space wouldn't work. After a long time I found out the reason, and hopefully this helps other people too.

This PR:

* Adds a new feature

Changes:

Adds a param to createCursorKeys, which allows it to emit keypress events as normal.

### DIFF
--- a/src/input/keyboard/KeyboardPlugin.js
+++ b/src/input/keyboard/KeyboardPlugin.js
@@ -395,9 +395,11 @@ var KeyboardPlugin = new Class({
      * @method Phaser.Input.Keyboard.KeyboardPlugin#createCursorKeys
      * @since 3.10.0
      *
+     * @param {boolean} [enableCapture=true] - Automatically call `preventDefault` on the native DOM browser event for the key codes being added.
+     *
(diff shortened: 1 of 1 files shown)


src/input/keyboard/KeyboardPlugin.js
```diff
@@ -395,9 +395,11 @@ var KeyboardPlugin = new Class({
      * @method Phaser.Input.Keyboard.KeyboardPlugin#createCursorKeys
      * @since 3.10.0
      *
+     * @param {boolean} [enableCapture=true] - Automatically call `preventDefault` on the native DOM browser event for the key codes being added.
+     *
      * @return {Phaser.Types.Input.Keyboard.CursorKeys} An object containing the properties: `up`, `down`, `left`, `right`, `space` and `shift`.
      */
-    createCursorKeys: function ()
+    createCursorKeys: function (enableCapture=true)
```
photonstorm: Please remove the `=true` part, this isn't ES5 (so can't be merged as-is).

codergautam:
Like that?

PR #6010, opened 2022-02-15
Fix WebAudio for Crodova based apps

kglogocki:

* Fixes a bug

Describe the changes below:

When using Cordova,  WebAudio doesn't work. It was fixed for Capacitor by adding check in `src/loader/File.js` that `xhr.responseUrl` starts with `capacitor://`. We are using different scheme for Cordova and it was not working for us. I've added configuration option to specify custom scheme to properly recognise if file is local and load sound file properly.


photonstorm:
Thanks for the PR. Can you please remove all of the dist files, package-lock and types from it, as I cannot merge those.

kglogocki:
I removed dist files, package-lock and types. I also merged latest changes from master.

photonstorm:
Thanks, I've merged this in and tweaked it a bit, so you can now specify an array of local schemes in the config, rather than just one. It's also available via the new `LoaderPlugin.localSchemes` property, too.

PR #6012, opened 2022-02-16
Fix kerning in BitmapText rendering

arbassic:
This PR:

- [ ] Updates the Documentation
- [ ] Adds a new feature
- [x] Fixes a bug

I noticed a quite major bug in the GetBitmapTextSize function, causing kerning not being applied to BitmapText objects. In the PR the bug is fixed and bitmap textfields are rendered properly.

Details:
---

Kerning values are read properly from glyph data. Local 'x' variable is modified by the kerning value also properly. But at the end of the loop's code local 'x' variable was not used to describe the character object, which is then stored in the 'characters' return array (out).

In the fix 'x' variable is used as intended to describe particular characters in the output array instead of general xAdvance.
Also xAdvance is now modified by kerningOffset to provide consistency between adjacent characters (necessary to place the next character at proper position)

-----

Below is an example illustration of how the bug impacts on characters in a textfield ([Jost](https://fonts.google.com/share?selection.family=Jost:wght@500) font):

pre

Original font sample:
Zrzut ekranu 2022-02-18 o 14 33 26

------

On the second image, that uses the modified code, the Y character is placed as in the original font. Space is slightly smaller and the general effect more consistent.

now

### DIFF
--- a/src/gameobjects/bitmaptext/GetBitmapTextSize.js
+++ b/src/gameobjects/bitmaptext/GetBitmapTextSize.js
@@ -386,7 +386,7 @@ var GetBitmapTextSize = function (src, round, updateOrigin, out)
             i: charIndex,
             char: text[i],
             code: charCode,
-            x: (glyph.xOffset + xAdvance) * scale,
+            x: (glyph.xOffset + x) * scale,
             y: (glyph.yOffset + yAdvance) * scale,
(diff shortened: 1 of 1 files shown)


photonstorm:
Lovely, thanks. I will test with some fonts that have any kerning defined.

PR #6023, opened 2022-02-25
Container game object should supply container reference to child rend…

### LINKED ISSUE #5853: Child spines don't inherit parent container's alpha
* Phaser Version: 3.55.2+

## Description

The child spines of a regular container do not use the parent's alpha, but child spines of a spineContainer do.

## Example Test Code

Running example here <https://jsfiddle.net/stupot/scojxkn8/26/>

The example will open using a regular container, with 2 spines in it.  The container alpha is set to 0.2, but the spines appear solid.

Change value in line 63 to change container type.  Setting it to 1 will use a spineContainer and the spines will now appear semi-transparent, as expected.

I would expect the spines to use the parent alpha regardless of which container type is used, in the same was positio/rotation/visible are honoured.

spayton:
fixes #5853


photonstorm:
👍

PR #6028, opened 2022-02-28
Improve the scene start logic

samme:
This PR

* Adds a new feature

These changes avoid some problems when starting scenes (#5999):

- If you start a scene that is already starting (`START`, `LOADING`, or `CREATING`), the start operation is ignored. (But since #5997 you can force the scene to restart by shutting it down first, if you want.)
- If you start a scene that is sleeping, it's shut down before starting again. This matches how Phaser currently handles paused scenes.

### DIFF
--- a/src/scene/SceneManager.js
+++ b/src/scene/SceneManager.js
@@ -1125,7 +1125,9 @@ var SceneManager = new Class({
     },
 
     /**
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

Issue #5954, opened 2021-12-22
hitAreaCallback type definition


SylvainBreton:
## Version

* Phaser Version: 3.55 but same code on 3.60

* Operating system: Ubuntu

## Description

In phaser/types/phaser.d.ts, Phaser.Types.Input.HitAreaCallback is defined as follow :

type HitAreaCallback = (hitArea: any, x: number, y: number, gameObject: Phaser.GameObjects.GameObject)=>void;

I think that there is an issue here as the definition is different from the one given in As this definition is different from the one given in phaser/src/input/typedefs/HitAreaCallback.js

## Additional Information

Change phaser/types/phaser.d.ts l62860 to 
type HitAreaCallback = (hitArea: any, x: number, y: number, gameObject: Phaser.GameObjects.GameObject)=>boolean;

samme:
I think this is from a problem in the tsgen parser. It has all the function types return `void`.

xmahle:
I have a fix for this issue, but I can't make a pull request: `ERROR: Permission to photonstorm/phaser.git denied to xmahle.

cc @samme @photonstorm

xmahle:
Also noticed that you are not returning `null` types, so `nullable` flag is ignored in the parser. I fixed that also...

samme:
@xmahle Push to your own origin remote <https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/configuring-a-remote-for-a-fork>

xmahle:
Nice...

photonstorm:
Sorry, forgot to close this issue after merging this PR a while ago!

PR #6029, opened 2022-02-28
Fix issue: hitAreaCallback type definition #5954

xmahle:

* Fixes a bug

Describe the changes below:

When parser created a `Typedef` and it was a function, we ignored the `returns` and just used static `dom.type.void`

This code is copied `createFunction` which picked the return type correctly.


photonstorm:
Thanks for this PR, as with the other one, please remove all files from it other than `Parser.ts`, as they're all build files.

photonstorm:
👍

PR #6030, opened 2022-02-28
Handle nullable in parser.parseType

xmahle:

* Fixes missing type script return type `null`

Describe the changes below:
Nullable flag was ignored, which makes all functions that can return `null` not containing the type for it.

This adds the null return type, to all such functions.

This is an issue, because ts doesn't know the function can return null, which can easily lead to bugs.

**Please Note**
This will make all ts based projects to fail type check, if there is no null check, when using functions that can return null.

**Example**
before:
```
/**
 * Retrieves a Scene.
 * @param key The Scene to retrieve.
 */
getScene(key: string | Phaser.Scene): Phaser.Scene;
```
after:
```
/**
 * Retrieves a Scene.
 * @param key The Scene to retrieve.
 */
getScene(key: string | Phaser.Scene): Phaser.Scene | null;
```

### DIFF
--- a/scripts/tsgen/src/Parser.ts
+++ b/scripts/tsgen/src/Parser.ts
@@ -486,7 +486,7 @@ export class Parser {
                     continue;
                 }
 
-                let param = dom.create.parameter(paramDoc.name, this.parseType(paramDoc));
+                let param = dom.create.parameter(paramDoc.name, this.parseType(paramDoc, dom.type.undefined));
                 parameters.push(param);
 
                 if (optional && paramDoc.optional != true) {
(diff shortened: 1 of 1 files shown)


scripts/tsgen/src/Parser.ts
```diff
@@ -523,6 +523,9 @@ export class Parser {
 
                 types.push(type);
             }
+            if (typeDoc.nullable) {
+                types.push(dom.type.null);
```
xmahle: I think we should use 'undefined' instead... since this get's applied to the function parameters also,

photonstorm:
Please can you remove all files from the PR other than `Parser.ts` as the rest are build files, thanks.

photonstorm:
👍

PR #6031, opened 2022-03-01
Fixed the type definition of SceneUpdateCallback

arosemena:

* Fixes a bug

Describe the changes below:

Fixes the `SceneUpdateCallback` type definition, it was missing the `time` and `delta` parameters

### DIFF
--- a/src/scene/typedefs/SceneUpdateCallback.js
+++ b/src/scene/typedefs/SceneUpdateCallback.js
@@ -3,4 +3,6 @@
  * @since 3.0.0
  *
  * @this Phaser.Scene
+ * @param {number} time - The current time. Either a High Resolution Timer value if it comes from Request Animation Frame, or Date.now if using SetTimeout.
+ * @param {number} delta - The delta time in ms since the last frame. This is a smoothed and capped value based on the FPS rate.
  */


photonstorm:

arosemena:
Apologies @photonstorm I missed that part, I've gone ahead and updated the correct typedef file and reverted the changes done directly at `phaser.d.ts`, let me know if there's anything else I should add, thanks!

photonstorm:
@arosemena thanks - I wonder if we should make both parameters optional? As this change will impact lots of TS devs otherwise.

arosemena:
Types should still be compatible, extra parameters can be ignored because we're using this from the consumer side, if phaser is always passing those values then they shouldn't be optional as we know they'll be there and having them as optional would require additional work to check they are defined which would be redundant because we know they are there.

Here's a sample where you can confirm this:

```typescript
type Before = (this: Scene) => void
type After = (this: Scene, time: number, delta: number) => void

const oldUpdate: Before = function() {} // valid old code
const newUpdate: After = oldUpdate // no type error
const newUpdate2: After = function(time, delta) {} // no type error, this would not work with the old type
```

photonstorm:
👍

PR #6039, opened 2022-03-12
Destroy game objects safely in layer.destroy()

rexrainbow:

* Fixes a bug

Describe the changes below:

The logic references from #5953 

Children of layer might be destroyed internally, thus the list length will be changed. 

Also bypass the `fromScene` parameter to children.


photonstorm:
👍

PR #6050, opened 2022-03-22
Fix CDN source url in README.md

lawrence-laz:
This PR:

* Updates the Documentation

Describe the changes below:

CDN urls were missing the `https` prefix in README.md. This PR adds the missing pieces.


photonstorm:
Thanks for the PR, but the URLs don't need the protocol added. The format we currently use allows for both HTTP and HTTPS formats.

lawrence-laz:
Are you saying that the snippet as it currently is should work, or that the user is expected to modify it to their own needs, i. e. prefix it with either https or http?

Because I copy pasted it and it didn't work on Chrome, and since examples are usually copy paste friendly this took me by surprise.

PR #6058, opened 2022-03-30
Docs: correct generateFrameNames() example

### LINKED ISSUE #6054: generateFrameNames: documentation wrong
Documentation states: "The `end` value tells it to look for 6 frames". 
I believe the 'end' value is the last index, not number of frames.
https://github.com/photonstorm/phaser/blob/1a086fc57c8faa53885fd1a4e20d617b958d3677/src/animations/AnimationManager.js#L587

samme:
This PR

* Updates the Documentation

Fixes #6054


photonstorm:
👍

PR #6060, opened 2022-03-31
Allow moving particles to (0, 0)

samme:
This PR

* Adds a new feature

It wasn't possible to move particles to (0, 0), so I added it.

`moveToX` and `moveToY` now default to `null` instead of `0`.

NB both `moveToX` and `moveToY` must now be set (even if 0) to enable to move-to behavior. This is a slight change from v3.55.2.


photonstorm:
👍

PR #6066, opened 2022-04-02
Fix kerning for BitmapText on Canvas renderer

TJ09:

* Fixes a bug

Describe the changes below:

#6012 resolved an issue where BitmapText did not take kerning into account when positioning letters. However, the canvas renderer for BitmapText uses almost exactly the same code, and was not updated to apply the same fix. This copies the change to xAdvance, ensuring that the kerning offset is used when calculating the next letter's position on non-WebGL.


photonstorm:
👍

PR #6067, opened 2022-04-02
Don't reassign the return values from require()

TJ09:

* Fixes a bug

Describe the changes below:

This is a bit weird, but when processing Phaser with Google's closure-compiler, it doesn't like when a variable from require() is reassigned to a new value (see: google/closure-compiler#3363).

This should be functionally the same, but in a way that pleases closure-compiler. These render modules are the only ones I saw that run into this case.


photonstorm:
👍

PR #6070, opened 2022-04-02
Disable context menu on input target instead of document.body

### LINKED ISSUE #6064: disableContextMenu does not respect input.mouse.target
## Version

* Phaser Version: ^3.54.0

* Operating system: all

* Browser: all

## Description

When setting `disableContextMenu: true`, Phaser disables the context menu on `document.body` even when `input.mouse.target` ~`scale.parent`~ is set to something else.
This results in a disabled context menu on the whole page, even parts that should not be under the control of Phaser.

I think the context menu should only be disabled on `input.mouse.target` ~`scale.parent` when it is set (or maybe even the canvas itself?)~ .

## Example Test Code

Relevant code parts:

https://github.com/photonstorm/phaser/blob/59fbcc5ca3f45388b833b07f94aefbc25cd1e24e/src/input/mouse/MouseManager.js#L287

https://github.com/photonstorm/phaser/blob/59fbcc5ca3f45388b833b07f94aefbc25cd1e24e/src/input/touch/TouchManager.js#L212

## Additional Information

I'm happy to help with this, when there are no objections.

lukashass:
This PR

* Fixes a bug

Describe the changes below:

Fixes #6064

### DIFF
--- a/src/input/mouse/MouseManager.js
+++ b/src/input/mouse/MouseManager.js
@@ -284,7 +284,7 @@ var MouseManager = new Class({
      */
     disableContextMenu: function ()
     {
-        document.body.addEventListener('contextmenu', function (event)
+        this.target.addEventListener('contextmenu', function (event)
         {
             event.preventDefault();
             return false;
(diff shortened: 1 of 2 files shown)


photonstorm:
👍

PR #6084, opened 2022-04-21
Allow input debug hit area colour to be adjusted after enabling.

spayton:

* Fixes a bug

Describe the changes below:

The docs say that the Debug Shape for input debugging, can be accessed and modified via the Game Object property 'hitAreaDebug' but preUpdate function always uses the colour specified when enabling, so changes to the 'strokeColor' property are ignored.

This update allows changes to the gameObject.hitAreaDebug.strokeColor to affect the colour of the hit area outline.

Possibly this is a new feature, but it feels more like this was the intended behaviour.

### DIFF
--- a/src/input/InputPlugin.js
+++ b/src/input/InputPlugin.js
@@ -2486,10 +2486,11 @@ var InputPlugin = new Class({
         if (debug)
         {
             debug.isFilled = false;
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

Issue #6042, opened 2022-03-14
Images with normal maps are not loaded properly


hexus:
## Version

* Phaser Version: 3.60.0-beta.4
* Operating system: Windows 10

## Description

Loading images with normal maps works in 3.55.x, but appears to fail in 3.60.0-beta.4.

## Example Test Code

Reproducible from the existing Phaser 3 examples that load normal maps.

For example: https://phaser.io/examples/v3/view/loader/image/load-normal-map-from-object

Switching this to "development build" loads 3.60.0-beta.4 and demonstrates both textures failing to draw.

## Additional Information

In my local testing, the both files are loaded successfully by the browser, but the texture is not assigned to any sprite that tries to use the texture key. If I'm able to track down the cause, I'll leave details here in a comment and/or open a PR.

photonstorm:
Thank you for submitting this issue. A fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

PR #6094, opened 2022-04-28
FIX: ImageFile addCache pending destruction status

VanaMartin:
This PR fixes a bug with failed loading of linked files if one of them is already scheduled for destruction,  addrresses #6042 reported earlier. 

Describe the changes below:

This migh not be the only change necessary as the change expects all status codes above FILE_COMPLETE to be results of successful load. This is quite efficient approach yet quite implicit. Alternatively a list of acceptable states should be enough.


photonstorm:
Hmm, this is a bit of a weird way of solving it (as they're meant to be constants). But, it works, so let's go with it for now!

PR #6104, opened 2022-05-06
Fix the GetTilesWithinShape() area

### LINKED ISSUE #5640: TilemapLayer#getTilesWithinShape does not return valid results when run with Phaser.Geom.Line
## Version

* Phaser Version: 3.52

* Operating system: Windows

* Browser: Observed on Chrome and Firefox

## Description

getTilesWithinShape does not return valid results when run with Phaser.Geom.Line. When a line spans two rows of tiles, tiles in the lower row can be left off of the results. It seems like the logic checking for the shape bounding box isn't valid for a line. I haven't tested with other shapes.

As an example - the tiles in the bottom row aren't returned by the function as expected.
phaserbug

## Example Test Code

https://jsfiddle.net/gva74z1x/2/

samme:
This PR

* Fixes a bug

The shape `right` and `bottom` coordinates weren't actually getting rounded up.

Fixes #5640


photonstorm:
👍

PR #6141, opened 2022-06-11
Let texture `scaleMode` override the `antialias` setting under `CANVAS`

cirras:
**This PR**
* Updates the Documentation
* Fixes a bug

My understanding has always been that linear texture filtering or canvas image smoothing will be used if `antialias` is true, or if the texture in question has its `scaleMode` set to `LINEAR` (0).

This holds true for the WebGL renderer, but doesn't seem to be case under the Canvas renderer.

I had a look and noticed that there are a number of statements of the form:
```
ctx.imageSmoothingEnabled = !(!antialias || frame.source.scaleMode);
```
This will short-circuit if `antialias` is disabled, never checking the `scaleMode`.
If `antialias` is true but `scaleMode` is `NEAREST`, then image smoothing will be disabled when drawing that texture.
However, the inverse isn't true. If `antialias` is false but `scaleMode` is `LINEAR`, image smoothing will still be disabled when drawing that texture.

My changes give `scaleMode` priority over `antialias`. In the above example, the logic is reduced to:
```
ctx.imageSmoothingEnabled = !frame.source.scaleMode;
```
In a case like this, we should be able to cut the `antialias` check out entirely since `antialias` already controls the default value of `TextureSource.scaleMode`.

### DIFF
--- a/src/gameobjects/blitter/BlitterCanvasRenderer.js
+++ b/src/gameobjects/blitter/BlitterCanvasRenderer.js
@@ -42,7 +42,7 @@ var BlitterCanvasRenderer = function (renderer, src, camera, parentMatrix)
     //  Blend Mode + Scale Mode
     ctx.globalCompositeOperation = renderer.blendModes[src.blendMode];
 
-    ctx.imageSmoothingEnabled = !(!renderer.antialias || src.frame.source.scaleMode);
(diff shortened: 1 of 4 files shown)


photonstorm:
Sure, this sounds good to me.

PR #6142, opened 2022-06-15
Fixes GameObjectFactory#group method's signature

shiqimei:

* Fixes a bug

Describe the changes below:

Since the first parameter could be the `config` argument, every type presented in the second parameter should be in the first one too.

Perhaps, we could define aliases for these two parameters and so that we don't need to maintain it ourselves. Like

```javascript

/** @typedef {Phaser.GameObjects.GameObject[]|Phaser.Types.GameObjects.Group.GroupConfig[]} GroupChildren */
/** @typedef {Phaser.Types.GameObjects.Group.GroupConfig|Phaser.Types.GameObjects.Group.GroupCreateConfig} GroupConfig */

/**
 * Creates a new Group Game Object and adds it to the Scene.
 *
 * Note: This method will only be available if the Group Game Object has been built into Phaser.
 *
 * @method Phaser.GameObjects.GameObjectFactory#group
 * @since 3.0.0
 *
 * @param {GroupChildren | GroupConfig} [children] - Game Objects to add to this Group; or the `config` argument.
 * @param {GroupConfig} [config] - A Group Configuration object.
 *
 * @return {Phaser.GameObjects.Group} The Game Object that was created.
 */
GameObjectFactory.register('group', function (children, config)
{
    return this.updateList.add(new Group(this.scene, children, config));
});
```

### DIFF
--- a/src/gameobjects/group/GroupFactory.js
+++ b/src/gameobjects/group/GroupFactory.js
@@ -15,7 +15,7 @@ var GameObjectFactory = require('../GameObjectFactory');
  * @method Phaser.GameObjects.GameObjectFactory#group
  * @since 3.0.0
  *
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

Issue #6144, opened 2022-06-17
SetTexture does not update hitarea


rexrainbow:
## Version

* Phaser Version: 3.55.2

* Operating system: win11

* Browser: Chrome, firefox

## Description

When texture size changed by `setTexture` method, hitarea does not update with this new texture size.

## Example Test Code

[Test code](https://codepen.io/rexrainbow/pen/JjpzBqv)
1. Create an image with 64x64 texture size, set hitarea with current texture size 64x64
2. Change texture and new texture size is 32x32.
3. Click between 32x32 and 64x64 area, the pointerdown event will fire. 

[Zone game object will change hitarea.width, hitarea.height when resizing](https://github.com/photonstorm/phaser/blob/master/src/gameobjects/zone/Zone.js#L172-L176), might need update hitarea with new texture size when `setTexture()`.

rexrainbow:
Issue closed due to PR merged

PR #6145, opened 2022-06-17
Update hitarea when changing frame

rexrainbow:

* Fixes a bug

Describe the changes below:

Update hitarea when changing frame, for issue #6144


photonstorm:
Hmm, I'll merge this and give it a test - as it feels like this may introduce some quite unintended side effects. Especially with regard to the changing state of a pointer over an animating sprite.

PR #6148, opened 2022-06-17
add "aac" sound type support

Ariorh1337:

* Adds a new feature

Describe the changes below:
Adding "aac" audio format support

### DIFF
--- a/src/device/Audio.js
+++ b/src/device/Audio.js
@@ -31,6 +31,7 @@ var Audio = {
     dolby: false,
     m4a: false,
     mp3: false,
+    aac: false,
     ogg: false,
     opus: false,
     wav: false,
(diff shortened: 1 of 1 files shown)


samme:
Needs a JSDoc `@property`.

photonstorm:
Thanks for your PR, but this feature is already tested for under the `Audio.m4a` property.

Ariorh1337:
> but this feature is already tested for under the `Audio.m4a` property

you are right, but files with aac format are not available for download due to an exception, this pr resolves the exception and makes it possible to use files with aac format

photonstorm:
> you are right, but files with aac format are not available for download due to an exception, this pr resolves the exception and makes it possible to use files with aac format

What exception exactly? On Safari on my Mac `Audio.m4a` is `true`, which would allow me to load an aac file.

Ariorh1337:
aac format is also `example.aac` files [Wikipedia](https://en.wikipedia.org/wiki/Advanced_Audio_Coding#Container_formats)
to be able to load them needs current PR, otherwise loader doesnot want to get load it
image

samme:
```js
this.load.video('example', { url: 'example.aac', type: 'mp4' });
```

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

Issue #6015, opened 2022-02-17
Matter Game Object Missing Types


jbromberg:
## Version

* Phaser Version: 3.55.2

* Operating system: MacOS 12.0.1

* Browser: Chrome

## Description

When setting a Phaser game object to have matter physics, the types are missing in the object. I can call functions like setVelocity() and they work as expected, however TypeScript is telling me that this method does not exist on the game object.

## Example Test Code

class Player extends Phaser.GameObjects.Container {
// ...
}

  addPlayer(spawnPoint: VecXY, username: string, character: string) {
    const player = new Player(this.scene, spawnPoint, username, character);
    return this.scene.matter.add.gameObject(player);
  }

const player = factory.addPlayer({ x: 0, y: 0 }, "", "");

// This line of code is giving me an error
player.setVelocity(0);

jbromberg:
What I'm really trying to do is have my Player class extend Phaser.Physics.Matter.Sprite but the super does not allow me to use a Phaser container to create it, only a single texture.

BMorearty:
I agree, this is a bug in the TypeScript definition.

To be more specific: the TypeScript type definition for the `Phaser.Physics.Matter.Factory.gameObject` function in phaser.d.ts is:

    gameObject(
      gameObject: Phaser.GameObjects.GameObject, 
      options?: Phaser.Types.Physics.Matter.MatterBodyConfig | MatterJS.Body, 
      addToWorld?: boolean): 
      Phaser.Physics.Matter.Image | Phaser.Physics.Matter.Sprite | Phaser.GameObjects.GameObject;

The return type of `Phaser.Physics.Matter.Image | Phaser.Physics.Matter.Sprite | Phaser.GameObjects.GameObject` is incorrect. By inspecting the `MatterGameObject` function in `MatterGameObject.js`, I believe the correct TypeScript return type that should be specified is:

    Phaser.GameObjects.GameObject &
    Phaser.Physics.Matter.Components.Bounce &
    Phaser.Physics.Matter.Components.Collision &
    Phaser.Physics.Matter.Components.Force &
    Phaser.Physics.Matter.Components.Friction &
    Phaser.Physics.Matter.Components.Gravity &
    Phaser.Physics.Matter.Components.Mass &
    Phaser.Physics.Matter.Components.Sensor &
    Phaser.Physics.Matter.Components.SetBody &
    Phaser.Physics.Matter.Components.Sleep &
    Phaser.Physics.Matter.Components.Static &
[...]

PR #6176, opened 2022-07-26
Fix TypeScript return type in Matter gameObject.

BMorearty:
The declared TypeScript return type for the
`Phaser.Physics.Matter.Factory.gameObject` function was wrong. As a
result, code like this--although correct--was giving a TypeScript error
saying the `setVelocity` method didn't exist:

    const player = this.matter.add.gameObject(player);
    player.setVelocity(0);

* Updates the Documentation
* Adds a new feature
* Fixes a bug

Describe the changes below:


BMorearty:
This fixes #6015.

photonstorm:

PR #6193, opened 2022-08-20
Add raw character index to BitmapText text bounds

JaroVDH:
This PR:
* Adds a new feature

Describe the changes below:
Character text bounds for BitmapText are currently indexed with the already wrapped text.
This minor change includes the actual index to the bounds.

From the wording I think the current index was always meant to be the actual base index, but for backwards compatibility I only added it as a new property.

What this solves:
The current index doesn't allow you to find the bounds by starting from the text you provide, rather than the internal wrapped text. For example if I had a text like `Hello  there!` that gets wrapped due to max width like this:
```
Hello
there!
```
then I can't find the bounds of `there` without mapping my text to the interal wrapped text, because everything starting from `t` is shifted over 1 because of the dropped space character.


photonstorm:
Thanks for the PR. I'm happy to merge this, but can you please remove the changes to the TS defs from the PR. Instead, they need adding to the typedefs we use to generate the defs from - in this case, it needs adding to `Phaser.Types.GameObjects.BitmapText.BitmapTextSize` (which is in `gameobjects/bitmaptext/typedefs` folder)

JaroVDH:
I should have noticed that, who can maintain a 5MB file manually.
Fixed it.

Took a bit to make my auto formatter to stop fussing about though 😅

photonstorm:
Thanks - all merged.

PR #6196, opened 2022-08-24
Add texture add/remove key events

samme:
This PR

* Adds a new feature (#5973)

This makes using `once()` for a specific texture much easier.

- Phaser.Textures.Events.ADD_KEY
- Phaser.Textures.Events.REMOVE_KEY


photonstorm:
👍

PR #6197, opened 2022-08-24
Exclude __WHITE in getTextureKeys()

samme:
This PR

* Fixes a bug

[getTextureKeys()](https://newdocs.phaser.io/docs/3.55.2/Phaser.Textures.TextureManager#getTextureKeys) would include `'__WHITE'`, which didn't seem right.


photonstorm:
👍

PR #6198, opened 2022-08-24
Destroy Arcade Physics debug graphic with world

samme:
This PR

* Fixes a bug

Phaser.Physics.Arcade.World would create a debug graphic but not destroy it. Ordinarily this isn't a problem, but if you do, e.g.,

```
this.physics.shutdown();
this.physics.start();
```

while the scene is running then the graphics will stack up.


photonstorm:
👍

PR #6202, opened 2022-08-26
Fix wrong alpha in snapshot pixel

samme:
This PR

* Fixes a bug

The renderer and Render Texture `snapshotPixel()` methods would give [incorrect pixel `alpha` and `alphaGL` values](https://phaser.discourse.group/t/get-pixel-of-rendertexture/7442/20).

There was unnecessary division.


photonstorm:
👍

PR #6204, opened 2022-08-29
Fix "shader too complex" game crash on high-end Android

rpn-ollie-james:
This PR
* Fixes a bug

Fixes issue on high end Android devices that handle a lot of textures, ie. 128
Previous code created 128 if statements and WebGL would fail to compile, crashing the game

### DIFF
--- a/src/textures/events/index.js
+++ b/src/textures/events/index.js
@@ -16,6 +16,6 @@ module.exports = {
     LOAD: require('./LOAD_EVENT'),
     READY: require('./READY_EVENT'),
     REMOVE: require('./REMOVE_EVENT'),
-    REMOVE_KEY: require('./REMOVE_KEY_EVENT'),
+    REMOVE_KEY: require('./REMOVE_KEY_EVENT')
 
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -2219,9 +2219,9 @@
       }
     },
     "node_modules/minimist": {
-      "version": "1.2.5",
-      "resolved": "https://registry.npmjs.org/minimist/-/minimist-1.2.5.tgz",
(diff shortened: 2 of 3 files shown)


photonstorm:
This is exactly how I used to have it (see https://github.com/photonstorm/phaser/commit/c8d3a20754c5769dc475d5ba0f41d94b02f6ad43) - but using this approach utterly tanks performance on iOS15.

So it's likely going to need some hybrid solution or device check that swaps shader at boot time.

Also, I can't merge this (aside from breaking iOS) because it uses ES6 string literals.

rpn-ollie-james:
Interesting, I will look into adding a device check, unless there's a Phaser 3 standard way of checking for iOS?
Although, doing some Googling it seems like they may have fixed the performance issue in iOS 15.4?

And wow, I didn't realise backticks were only ES6, I always code in Typescript :)

src/textures/events/index.js
```diff
@@ -16,6 +16,6 @@ module.exports = {
     LOAD: require('./LOAD_EVENT'),
     READY: require('./READY_EVENT'),
     REMOVE: require('./REMOVE_EVENT'),
-    REMOVE_KEY: require('./REMOVE_KEY_EVENT'),
+    REMOVE_KEY: require('./REMOVE_KEY_EVENT')
```
rpn-ollie-james, line 19: I didn't touch this file but it came up on running lint so I just fixed it..

rpn-ollie-james:
Have updated with check for OS.iOS and removed use of ES6 string literal

rpn-ollie-james:
Oh, and the package change is a minor version bump because npm said it was a critical security vulnerability and it was a non-breaking change..

photonstorm:
Thank you for your work on this, but I'm going to close this PR off as the 3.60 Beta 11 completely overhauled the mobile pipeline, and improved performance no end meaning there's no longer a requirement for this approach.

PR #6206, opened 2022-08-29
Show WebGL error code

zpxp:

* Fixes a bug

Describe the changes below:

Show the webgl error code value if it is not present within the object map. Currently it just prints "Framebuffer status: undefined", now it will print "Framebuffer status: 69420" etc


photonstorm:
👍

PR #6208, opened 2022-09-03
Fix function Graphics -> strokeRoundedRect

temajm:
This PR

* Fixes a bug

If you call the strokeRoundedRect function in Graphics, and pass to arguments a radius greater than half of the smaller side of the rectangle, then the library will draw a completely different figure.

Adding a maximum radius to the code fixes this bug.

Before:
image

After (Desired result):
image


photonstorm:
👍

PR #6211, opened 2022-09-06
Fix: BitmapText width calculation

monteiz:
The BitmapText width should be calculated using the actual glyphs' `width`, not their `xAdvance` property.

This PR

* Fixes a bug

The `xAdvance` property defines how distant in horizontal space the next glyph should be drawn. It does not necessarily coincide with the glyph `width` property.
Consider for example a glyph with a shadow on the right side. The glyph should advance by its total width minus the right shadow offset. The total bitmaptext however should be calculated taking into account the shadow as well.


monteiz:
After a second thought, the `xAdvance` approach is the most accurate one. Even in presence of a shadow, what it really counts is the glyphs advancement.

Closed.

PR #6216, opened 2022-09-08
Improve visible children performance

JernejHabjan:
This PR

* Improves performance

using native filter instead of manually looping and appending children to array

PR #6217, opened 2022-09-08
Improve stable sort performance

JernejHabjan:
This PR

* Improves performance

[Issue created here](https://github.com/photonstorm/phaser/pull/6216). I'm not sure if this is ok solution, but simple array.sort could be used

### DIFF
--- a/src/core/Config.js
+++ b/src/core/Config.js
@@ -593,6 +593,12 @@ var Config = new Class({
                 this.renderType = CONST.CANVAS;
             }
         }
+
+
+        /**
+         * @const {boolean} Phaser.Core.Config#es2019 - Use native ES2019 features
+         */
+        this.es2019 = GetValue(config, 'es2019', false);
     }
 
 });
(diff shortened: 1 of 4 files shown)


photonstorm:
This may need putting behind a Game Config flag because Array sort is unstable on lots of older browsers, and without a stable sort, it causes all kinds of rendering errors on the likes of IE.

JernejHabjan:
@photonstorm not sure how to proceed here - created additional commit, but not sure this is the right way

photonstorm:
Leave this with me - I think we can switch this flag automatically if, for example, we detect IE vs. Chrome. Plus, I can't merge it because the PR contains files that shouldn't be updated :) (like the TS defs). But I'll have a play with this now, as I think we can use it somehow.

photonstorm:
Ok, this feature is now in the master branch (and enabled based on browser detection), so worth you re-testing your previous demos, please.

JernejHabjan:
Commented the perf improvement on issue

Issue #6212, opened 2022-09-06
Tilemap, when read second time, the prefix added


wahur666:
When loaded a Tiled map editor generated JSON, with the following structure:

![kép](https://user-images.githubusercontent.com/9078042/188695948-8693c57d-fba3-45cc-8c86-9e15e09b8c33.png)

Using the following code:
```js
import * as Phaser from "phaser";
import map1 from "./map2.json";

export default class PreloadScene extends Phaser.Scene {
    constructor() {
        super("preload");
    }
    createMap(gameScene) {
        return  gameScene.make.tilemap({key: "map1"});
    }
    preload() {
        this.load.tilemapTiledJSON("map1", map1);
    }
    create() {
        const t1 = window.tileMap = this.createMap(this);
        const t2 = window.tileMap2 = this.createMap(this);
        console.log(t1.objects);
        console.log(t2.objects);
    }
}
```

The first time the map loaded correctly, with the object layer names, for example: `Pads/Pad1`

```js
Array(10) [ {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…} ]
0: Object { name: "Resources", opacity: 1, type: "objectgroup", … }
1: Object { name: "Pads/Pad1", opacity: 1, type: "objectgroup", … }
2: Object { name: "Pads/Pad2", opacity: 1, type: "objectgroup", … }
3: Object { name: "Pads/Pad3", opacity: 1, type: "objectgroup", … }
4: Object { name: "Nuke", opacity: 1, type: "objectgroup", … }
5: Object { name: "Bases/Base1", opacity: 1, type: "objectgroup", … }
6: Object { name: "Bases/Base2", opacity: 1, type: "objectgroup", … }
7: Object { name: "Water", opacity: 1, type: "objectgroup", … }
8: Object { name: "Terrain", opacity: 1, type: "objectgroup", … }
9: Object { name: "Obstacle", opacity: 1, type: "objectgroup", … }
```

But when it is loaded for the second time, the result is the following:

```js
Array(10) [ {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…} ]
0: Object { name: "Resources", opacity: 1, type: "objectgroup", … }
1: Object { name: "Pads/Pads/Pad1", opacity: 1, type: "objectgroup", … }
2: Object { name: "Pads/Pads/Pad2", opacity: 1, type: "objectgroup", … }
[...]

PR #6219, opened 2022-09-09
Fix function ParseJSONTiled

temajm:
This PR
* Fixes a bug

A call to the ParseJSONTiled function changes the object passed as the second argument of the function, so that when this function is called again with the same object, the function returns a different result than the first call to this function.

This PR also fixes a problem described in https://github.com/photonstorm/phaser/issues/6212

before:
image

after (Desired result):
image

### DIFF
--- a/src/tilemaps/parsers/tiled/ParseJSONTiled.js
+++ b/src/tilemaps/parsers/tiled/ParseJSONTiled.js
@@ -7,6 +7,7 @@
 var AssignTileProperties = require('./AssignTileProperties');
 var BuildTilesetIndex = require('./BuildTilesetIndex');
 var CONST = require('../../const/ORIENTATION_CONST');
+var DeepCopy = require("../../../utils/object/DeepCopy");
 var Formats = require('../../Formats');
 var FromOrientationString = require('../FromOrientationString');
 var MapData = require('../../mapdata/MapData');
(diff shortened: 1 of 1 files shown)


photonstorm:
Thanks for the PR. I'd be more inclined to see what is it about `ParseObjectLayers` that mutates the data (because most of the things this parser does is perfectly safe, just copying numbers across)

temajm:
> ParseObjectLayers

The ``ParseObjectLayers`` function edits the json object passed to it, so when the function ``gameScene.make. tilemap({key: "map1"});``the function passes this changed json object as an argument to the ``ParseJSONTiled`` function, which then edits the same object again, since the changed object has a name with a prefix and the ``ParseObjectLayers`` function reapplies the prefix.

Solution options:
 1) It is necessary to pass a copy of the json object to the ``ParseObjectLayers`` function so that its changes to that object do not edit the main object.
 2) It's necessary to forbid the ``ParseObjectLayers`` function to edit the object, but use ordinary variables instead of storing the changed data.
 3) It is necessary to make a copy of exactly those objects which are edited by the ``ParseObjectLayers`` function.

photonstorm:
Looking at it, nothing in `ParseObjectLayers` mutates the json, so it must be even deeper than that - likely in `CreateGroupLayer`.

photonstorm:
Just checked and nope, nothing in there does it either! However, it does do this: `layers: json.layers,` which returns the whole original object, so I guess something outside of the parser then edits this, thinking the data is safe. It may be enough just to set that to be a cloned object.

temajm:
You can clone only json.layers, but in my solution I clone the whole json, because besides ``mapData.objects = ParseObjectLayers(json);`` the ``ParseJSONTiled `` function calls other parsers like ``mapData.images = ParseImageLayers(copyJsonData);`` and I suspect they can edit the object too (I haven't checked if they edit the json object)

photonstorm:
Thanks, I'm going to do this the same way you have, but just with a different property name. This is now in the master branch (and attributed to you in the Change Log).

PR #6220, opened 2022-09-10
Rope and Video already on update list

samme:
This PR

* Fixes a bug

I think it was harmless but `make.rope()` and `make.video()` had an unnecessary branch for adding to the update list. The game objects are already added to the update list by `BuildGameObject()`.


photonstorm:
👍

PR #6221, opened 2022-09-10
Update video texture after loading

samme:
This PR

* Fixes a bug

I found that `loadURL()` (without `noAudio = true`) didn't show anything and didn't emit the `VIDEO_CREATED` event, although I could see the video element had loaded. I assume `loadMediaStream()` is the same.

I added a load event handler that calls `updateTexture()` and this seems to fix it — the video appears and `VIDEO_CREATED` is emitted.


photonstorm:
👍

PR #6223, opened 2022-09-19
Give TilemapLayer origin (0, 0)

samme:
This PR

* Updates the Documentation
* Fixes a bug

Tilemap Layers have the default origin (0.5, 0.5) but they always render as (0, 0). So visually you'll get the wrong results from `getBounds()`, `camera.cull()`, etc.

I changed the default origin to (0, 0) and marked the `originX/Y` and `displayOriginX/Y` properties read-only.


photonstorm:
👍

PR #6228, opened 2022-09-25
Tween seek use less then instead of less or equal

EmilSV:
* Fixes a bug
Tween seek  could do an infinite loop if the seek value was 1. 
using <  instead of <= would fix that


photonstorm:
👍

Issue #6233, opened 2022-10-03
MultiPipeline shaders cause crash on older Android devices


arbassic:
## Version

* Phaser Version: v3.60.0-beta.9

* Operating system: Android 6.0.1 / Android 8.1.0

* Browser: Mozilla/5.0 (Linux; Android 8.1.0; TM-MID792WP Build/O11019; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/105.0.5195.136 Mobile Safari/537.36

## Description

When initializing Phaser on Multi.frag and Multi.vert load, an error occurs:

> **Fragment Shader failed:ERROR:** 0:4: 'highp' : precision is not supported in fragment shader

## Example Test Code

Pseudocode:
```
Create instance of Phaser.Game

Display a Phaser.Sprite

Observe how error raise
```

## Additional Information

It is a rare case occurring only on old Android devices but globally it may break UX quite often!

Possible workaround would be to add IF statement to choose between highp and mediump using the proper GL constant, just like here https://github.com/photonstorm/phaser/issues/6120

```c
#ifdef GL_FRAGMENT_PRECISION_HIGH
#define highmedp highp
#else
#define highmedp mediump
#endif
precision highmedp float;
```

or alternatively:

```c
#ifdef GL_FRAGMENT_PRECISION_HIGH
precision highp float;
#else
precision mediump float;
#endif
```

Fix should be made in following files:
- src/renderer/webgl/shaders/src/Multi.frag
- src/renderer/webgl/shaders/src/Multi.vert
- src/renderer/webgl/shaders/Multi-vert.js
- src/renderer/webgl/shaders/Multi-frag.js

-------------

After official **mdn web docs**

> [In WebGL 1, "highp float" support is optional in fragment shaders](https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#in_webgl_1_highp_float_support_is_optional_in_fragment_shaders)

photonstorm:
Thanks for opening this issue, and for submitting a PR to fix it. We have merged your PR into the `master` branch and attributed the work to you in the [Change Log](https://github.com/photonstorm/phaser/blob/master/CHANGELOG.md). If you need to tweak the code for whatever reason please submit a new PR.

PR #6234, opened 2022-10-03
fix: use webgl 1 pattern to always choose best precision available

arbassic:
* Fixes a bug

Describe the changes below:

The bug fixed by the PR:
[#6233 MultiPipeline shaders cause crash on older Android devices ](https://github.com/photonstorm/phaser/issues/6233)

------

[In WebGL 1, "highp float" support is optional in fragment shaders](https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#in_webgl_1_highp_float_support_is_optional_in_fragment_shaders)

Should have minimal impact on performance (done once per the program run, during the shader compilation) and provides more portability to Phaser on older devices


photonstorm:
👍

PR #6235, opened 2022-10-04
Simplify scene types

ubershmekel:
`SettingsConfig | CreateSceneFromObjectConfig` is a valid scene type. Now it will be allowed in GameConfig.

* Fixes a bug

Describe the changes below:

I'm just trying to use typescript with phaser. I can't pass an array of objects with both `key` and `preload` into `Phaser.Types.Core.GameConfig.scene` because the way the type is set up right now it's either an array of  `Phaser.Types.Scenes.SettingsConfig` or an array of `Phaser.Types.Scenes.CreateSceneFromObjectConfig`. Definitely not an array of the union. But Phaser does allow the union of the types in there.

On the way I already updated another location that had the same type signature.

Having `SceneType` will also make it easier for me to declare an object that's a `Phaser.SceneType` rather than `Phaser.Types.Scenes.SettingsConfig | Phaser.Types.Scenes.CreateSceneFromObjectConfig`.


ubershmekel:
If you'd rather just see my repo to see the typing issue, here's the object I'm generating:

https://github.com/ubershmekel/vite-phaser-ts-starter/blob/2d9f7d7d1394e736dc8c59750ae8bec6cadd96e3/src/menu-scene.ts#L7

And this line throws an error because of the currently defined type:

https://github.com/ubershmekel/vite-phaser-ts-starter/blob/2d9f7d7d1394e736dc8c59750ae8bec6cadd96e3/src/main.ts#L24

> Type 'SettingsConfig | CreateSceneFromObjectConfig' is not assignable to type 'Function'.
  Type 'SettingsConfig' is missing the following properties from type 'Function': apply, call, bind, prototype, and 5 more.

photonstorm:

ubershmekel:
Thank you for explaining that. I fixed the PR so please take another look.

Btw it seems you haven't run `tsgen` in a while and a lot changed. Like `Smootherstep` switches places with `SmoothStep`. Maybe it's because I ran `tsgen` on windows?

E.g. scripts/tsgen/test/output.txt gets the following errors in it.
```
+../../../types/phaser.d.ts(84058,43): error TS2694: Namespace 'Phaser.Math' has no exported member 'Vec2'.
+../../../types/phaser.d.ts(89248,40): error TS2694: Namespace 'Phaser.Scenes' has no exported member 'Scene'.
```

Though it seems I can't update this PR even though I updated my fork of master. See https://github.com/photonstorm/phaser/compare/master...ubershmekel:phaser:master

Should I open a new PR?

photonstorm:
`tsgen` was last built when I released 3.60 Beta 10, so around two weeks ago. Lots of changes since then, yes, but that is what the output debug is for - fixing TS errors.

I would open a new PR, yes.

PR #6236, opened 2022-10-04
Skip the Group createMultiple() post-creation options if not set

samme:
This PR

* Adds a new feature

This is more efficient and will let authors set alpha, position, etc. in `createCallback` or a game object constructor without having those values overwritten.

This is a breaking change.

The post-creation options are:

- `setAlpha`
- `setDepth`
- `setOrigin`
- `setRotation`
- `setScale`
- `setScrollFactor`
- `setXY`


photonstorm:
👍

PR #6261, opened 2022-10-21
Reverse order in which indexes are substracted when sorting gameobjects for inputs

LaP0573:
This PR
* Fixes a bug

Describe the changes below:

When setting the `topOnly` property to false on the `InputPlugin`, the game objects receive the events in reverse render object (i.e bottom-most object receives before top-most).

This feels counterintuitive as they should be receiving them from top to bottom.

This PR aims to fix that by modifying the `sortGameObjects` method.

### DIFF
--- a/src/input/InputPlugin.js
+++ b/src/input/InputPlugin.js
@@ -2678,7 +2678,7 @@ var InputPlugin = new Class({
             var indexA = Math.max(list.indexOf(childA), 0);
             var indexB = Math.max(list.indexOf(childB), 0);
 
-            return indexB - indexA;
+            return indexA - indexB;
         });
     },
 
--- a/src/input/InputPlugin.js
+++ b/src/input/InputPlugin.js
@@ -632,12 +632,12 @@ var InputPlugin = new Class({
                 //  Only the top-most one counts now, so safely ignore the rest
                 if (this._temp.length)
                 {
-                    this._temp.splice(1);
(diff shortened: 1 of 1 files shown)


photonstorm:
Hi, I've checked this and it doesn't appear to be doing what you say. Take the following code:

```js
{
    preload ()
    {
        this.load.image('eye', 'assets/pics/lance-overdose-loader-eye.png');
    }

    create ()
    {
        for (let i = 0; i < 14; i++)
        {
            const s = this.add.sprite(100 + i * 30, 100 + i * 30, 'eye').setInteractive().setName('eye ' + i);

            s.on('pointerdown', () =>
            {
                console.log(s.name, s.x, s.y);
            });
        }

        //  If you disable topOnly it will fire events for all objects the pointer is over
        //  regardless of their place on the display list
        this.input.setTopOnly(false);
    }
}

const config = {
    type: Phaser.WEBGL,
    parent: 'phaser-example',
    width: 800,
    height: 600,
    scene: Example
};

```

When run, if I click the bottom-right sprite at the point where several overlap (the one at the very top of the display list), I get this in the console:

```text
eye 13 490 490
eye 12 460 460
eye 11 430 430
eye 10 400 400
```

So it's clearly emitting the event for 'eye 13' first, then 'eye 12', and so on. This is the correct order, as 'eye 13' is at the top of the display list.

Please provide a code sample that demonstrates the events emitting in the wrong order (open it as an issue if you like)

PR #6263, opened 2022-10-25
When calling the 'Phaser/Geom/Intersects#LineToLine' method without passing in the third parameter, 'out', avoid creating a new Phaser.Geom.Point, as it will be totally inaccessible.

Trissolo:

* Fixes a bug

Describe the changes below:
Not quite a bug fix😅: just a very small performance improvement by not creating an inaccessible object.
In fact, the third parameter, 'out', must have an external reference (e.i. it must be passed in by the user), otherwise creating and populating an unreachable new Point is just a waste of memory and computing time.

### DIFF
--- a/src/geom/intersects/LineToLine.js
+++ b/src/geom/intersects/LineToLine.js
@@ -15,7 +15,7 @@
  *
  * @param {Phaser.Geom.Line} line1 - The first Line to check.
  * @param {Phaser.Geom.Line} line2 - The second Line to check.
- * @param {(object|Phaser.Geom.Point|Phaser.Math.Vector2)} [out] - An optional Point-like object in which to store the coordinates of intersection, if needed.
+ * @param {Phaser.Types.Math.Vector2Like} [out] - An optional point-like object in which to store the coordinates of intersection, if needed.
  *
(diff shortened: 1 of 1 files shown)


src/geom/intersects/LineToLine.js
```diff
@@ -17,13 +15,12 @@ var Point = require('../point/Point');
  *
  * @param {Phaser.Geom.Line} line1 - The first Line to check.
  * @param {Phaser.Geom.Line} line2 - The second Line to check.
- * @param {Phaser.Geom.Point} [out] - A Point in which to optionally store the point of intersection.
+ * @param {(object|Phaser.Geom.Point|Phaser.Math.Vector2)} [out] - An optional Point-like object in which to store the coordinates of intersection, if needed.
```
samme: I think `Phaser.Types.Math.Vector2Like` is best here.

Trissolo: Done, I think.
I pushed another commit to my forked repository, but I'm not sure if this is the right way: this is my second PR ever! 🤦‍♂️

photonstorm:
👍

PR #6271, opened 2022-10-30
Replace the big number with Infinity

temajm:
This PR
* Fixes a bug

Describe the changes below:

I think it makes more sense to use the Infinity definition than to use a large number instead.


photonstorm:
This won't work, I'm afraid:

```js
let counter = Infinity;
counter = counter - 100;
console.log(counter); // outputs 'Infinity'
```

You can't use it as a seed for a decrementing value, as we do in all those cases.

We _could_ use `Number.MAX_SAFE_INTEGER` instead, but there's no real point, it just takes up more bytes for no benefit in this case.

PR #6275, opened 2022-11-07
Fixed getPostPipeline

neki-dev:
If pass parameter as **string** (pipeline name) to `getPostPipeline`, then get error:

```
Uncaught TypeError: Right-hand side of 'instanceof' is not an object
```

This happens when there is pipeline with name different from the one are looking for.
I changed the condition to solve this problem.


photonstorm:
Thanks for your PR and finding this bug. I have used a slightly different approach to resolve it, but have attributed it to you in the Change Log.

PR #6286, opened 2022-11-16
Updated SpineGameObject type definition

AlbertMontagutCasero:
This PR
* Fixes a bug

Describe the changes below:

The type exposed in the definition file does not correspond to the source code, which makes it incompatible. This commit fixes that.


photonstorm:
👍

PR #6291, opened 2022-11-23
Update spine build config to match newer spine runtimes.

orjandh:

* Fixes a bug

Describe the changes below:

The spine plugin build configs have been updated to match the source code of newer versions of the spine runtime. This will no longer work for spine-runtime 3.8, but I guess that should not be a problem? I think the build readme also is slightly outdated, but I am not sure.

### DIFF
--- a/plugins/spine/tsconfig.both.json
+++ b/plugins/spine/tsconfig.both.json
@@ -7,18 +7,18 @@
         "preserveConstEnums": true,
         "outFile": "src/runtimes/spine-both.js",
         "sourceMap": true,
-        "declaration": true
+        "declaration": true,
+        "target": "es6"
     },
     "include": [
-        "spine-runtimes/spine-ts/core/src/**/*",
-        "spine-runtimes/spine-ts/canvas/src/**/*",
-        "spine-runtimes/spine-ts/webgl/src/**/*"
+        "spine-runtimes/spine-ts/spine-core/src/**/*",
+        "spine-runtimes/spine-ts/spine-canvas/src/**/*",
+        "spine-runtimes/spine-ts/spine-webgl/src/**/*"
     ],
     "exclude": [
-        "spine-runtimes/spine-ts/build",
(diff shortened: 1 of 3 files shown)


photonstorm:
Sorry but these need splitting up. I cannot replace the 3.8 plugin with 4.0. It needs to be offered as an additional plugin so that users can pick. Also, have you actually tested this? I don't mean the build, but does the Phaser plugin even work after upgrading to 4.0?

orjandh:
I understand if this cannot be merged, just thought I'd share it. I think having two versions is a great idea.

I have not done very thorough testing, but the plugin build seemed to work fine when I tested it. I don't have a spine license myself, so I am not able to do edits and new exports. But perhaps I can find some example files for the different versions and do more proper tests.

spayton:
@orjandh can you check your plugin to see if it really is using the v4.1 runtime?  The v3.8 plugin will contain the string "3.8.75", v4.1 won't.

I tried this, had to do a little extra config to get it to find the spine-core.  Then the "npm run plugin.spine.runtimes" finishes, apparently ok with nothing displayed, but it hadn't rebuilt the contents of spine\src\runtimes, so building the plugin just uses existing runtime, which of course works.

orjandh:
You are completely right. I had assumed everything went fine, but it just "failed" silently. Apparently the typescript outFile parameter cannot be used with module="none" if the source is using ES modules. As the spine runtimes are now using ES modules, this does not work anymore.

spayton:
Yeah, I did exactly the same and was surprised that it worked when I was expecting trouble.  I hacked at it a little further to get it to build some runtimes, then also the plugin, but that failed at runtime, not necessarily because the v4 interface was incompatible, but because I still had some module problems... left it there for a rainy day.

photonstorm:
Closing this because it sadly doesn't work. We'll cover the Spine 4 upgrade ourselves.

Issue #6301, opened 2022-11-30
Tilemap Collisions do not work properly in 3.60.0 Beta 15


chrisl8:
## Version

* Phaser Version: 3.60.0 Beta 15

* Operating system: Windows

* Browser: Google Chrome

## Description

Tilemap collisions in Beta 15 are not working correctly, compared to Beta 14 where they worked as expected.

## Example Test Code

In this example the yellow box falls through the purple "floor" boxes in the tilemap:

https://labs.phaser.io/view.html?src=src\tilemap\collision\simple%20map.js&v=3.60.0-beta.15

while in Beta 14, the yellow box stops when it hits the purple "floor" boxes in the tilemap:

https://labs.phaser.io/view.html?src=src\tilemap\collision\simple%20map.js&v=3.60.0-beta.14

## Additional Information

I brought this up in Discord already, so Phontonstorm is aware. I just wanted to make it official to help track the solution.

chrisl8:
I've narrowed it down to some change within `src/tilemaps/`.

I believe that commit 4bf7e1978fee94973a7a33e62a4e0a5ea2b76936 is what broke it.

chrisl8:
I believe that this line

https://github.com/photonstorm/phaser/blob/1747cb67bea18c223bca5d418662fce74b160f70/src/tilemaps/components/WorldToTileXY.js#L28

should be

`if (snapToFloor === undefined) { snapToFloor = true; }`

chrisl8:
The reason I believe that this is the correct code for this line is because that is how the line was written in [WorldToTileX.js](https://github.com/photonstorm/phaser/commit/4bf7e1978fee94973a7a33e62a4e0a5ea2b76936#diff-e51f3af259c5644d73d7c277ce6b1b15d3a912dc6fe014ce9318decd5f06061a) and [WorldToTileY.js](https://github.com/photonstorm/phaser/commit/4bf7e1978fee94973a7a33e62a4e0a5ea2b76936#diff-be05bc6e794ac6ddf5ca6bebb3a0a7cfb29793d196eae6d958414630b22dbc23) prior to this change.

See here:

https://github.com/photonstorm/phaser/commit/4bf7e1978fee94973a7a33e62a4e0a5ea2b76936#diff-e51f3af259c5644d73d7c277ce6b1b15d3a912dc6fe014ce9318decd5f06061a

I tested my PR code with my own games and with the example code and it appears to fix the problem.

photonstorm:
[...]

PR #6302, opened 2022-11-30
Fix Issue #6301 causing tilemap collisions to fail.

chrisl8:

* Fixes a bug

Describe the changes below:

Check specifically if `snapToFloor` is undefined, vs. false. Reverting a change made earlier, in order to fix a bug with tile map collisions.

This is related to issue #6301


photonstorm:
👍

Issue #3611, opened 2018-04-23
Set default origin (globally) for all GameObjects


NemoStein:
It's really strange to me that images/sprites are aligned in the center, not in the top-left corner, which makes it hard to build UIs.

So, is it possible to add a option to set the default origin of gameobjects?

---
**EDIT:** I just found out about the `Phaser.Display.Bounds` namespace, which eases the work with positioning of objects.
Is it the "Phaser 3 Way™" of aligning objects?
If so, shouldn't a GameObject have a delegate to it?
```javascript
// Something like this
this.add.image(0, 0, 'image').bounds.setLeft(100)
```

photonstorm:
It's quite normal for game frameworks to have x/y represent the center. You can just do `this.add.image(x, y, key).setOrigin(0)` if you want to have x/y represent the top-left represent instead. There are no plans to make a configurable 'global' origin as it would introduce an extra look-up for every single GO created, which isn't worth it when it's so simple to adjust the origin at run-time.

PR #6304, opened 2022-12-01
Add the ability to set a default value for the origin

temajm:
This PR
* Adds a new feature

I think we should add the ability for developers to set a default value for the origin parameter in objects

This was asked for in 2018: https://github.com/photonstorm/phaser/issues/3611

!!! Before merging this PR, please answer the questions that were spelled out in the TODO comments  !!!


temajm:
setDefaultOrigin can be renamed to resetOrigin if desired

photonstorm:
Thank you for spending time doing this. However, I do wish you had asked first. There's a reason the issue requesting it is closed, and the reason I gave there still stands today. It's not worth the extra look-up time required for _every single_ Game Object that is created to then check all the way up through the Scene into the game config and out.

Plus, we cannot use the `x?y?z?` approach as it's ES6 only. The v3 codebase is ES5.

The reason some Game Objects don't default to 0.5 is that it doesn't make sense in those specific cases. This is especially true for Tilemaps, which you nearly always want to align with world 0x0 rather than have to calculate their center. Text, again, is nearly always left-aligned, although this one is a bit more common to be centered. Either way, it has been like this for many years, and it's not worth the massive change it would introduce (to devs updating their games, not to the API as such) to change those defaults at this stage.

Phaser 4 already has this feature. It's too late to retrofit it to v3, I'm afraid.

PR #6309, opened 2022-12-05
Add `lineSpacing` property and `setLineSpacing` method to BitmapText

arbassic:
A similar feature to the `Text`'s `lineSpacing`

Should work in both Canvas and WebGL modes
For a text displaying a single line `lineSpacing` has no influence on height nor its bounds


photonstorm:
Sure, sounds like a nice and non-intrusive addition. Will need adding to the TypeDefs too, but I can do that.

PR #6320, opened 2022-12-13
Fixed text related paths in phaser-core.js

pavle-goloskokovic:
This PR Fixes a bug:

Several text related paths were incorrect.
This PR fixes those paths and enables `phaser-core.js` to be used as entry without errors.


photonstorm:
👍

PR #6330, opened 2022-12-24
Compare `maxSpeed` with updated velocity

### LINKED ISSUE #6329: Arcade Physics: maxSpeed not applied consistently
## Version

* Phaser Version: v3.55.2

* Browser: All browsers

## Description

When using the `maxSpeed` property with a given body in arcade physics, the expected behaviour is that the speed of that body will be capped on each physics update step. In practice, it appears that this only applies on every other step. As a result, arcade bodies with a `maxSpeed` value will end up moving slightly faster than they should.

## Example Test Code

https://codesandbox.io/s/phaser-3-max-speed-issue-iu9beh

## Additional Information

The cause of this behaviour [appears to be in the 'computeVelocity' function.](https://github.com/photonstorm/phaser/blob/master/src/physics/arcade/World.js#L1340) When deciding whether to scale the body's velocity, a comparison is made to confirm that `speed > maxSpeed`. At this point, speed has not actually been updated to represent the body's new velocity.  As far as I can tell, changing this comparison to `body.velocity.length() > maxSpeed` should fix the undesired behaviour.

Bambosh:

* Fixes a bug

Describe the changes below:
Fixes #6329

### DIFF
--- a/src/physics/arcade/World.js
+++ b/src/physics/arcade/World.js
@@ -1337,7 +1337,7 @@ var World = new Class({
 
         body.velocity.set(velocityX, velocityY);
 
-        if (maxSpeed > -1 && speed > maxSpeed)
+        if (maxSpeed > -1 && body.velocity.length() > maxSpeed)
         {
             body.velocity.normalize().scale(maxSpeed);
             speed = maxSpeed;


photonstorm:
👍

PR #6332, opened 2022-12-27
fix the bug when Edge and Opera browsers are not defined

ArtemSiz:

* Updates the Documentation
* Adds a new feature
* Fixes a bug

Describe the changes below:
The problem is that two browsers are incorrectly defined: Edge and Opera. The reason was found - the value of the string "navigator.userAgent" has changed.
Browsers where the error was detected:
Edge 108.0.1462.54
Opera 94.0.4606.38

I found the problem using the latest version of Phaser in my project, you can also see it here: https://jsfiddle.net/artem_syzonenko/4ng20okq/

edge_v_108 0 1462 54

opera_1_94 0 4606 38
opera_2_94 0 4606 38

### DIFF
--- a/src/device/Browser.js
+++ b/src/device/Browser.js
@@ -54,9 +54,13 @@ function init ()
 {
     var ua = navigator.userAgent;
 
-    if ((/Edge\/\d+/).test(ua))
+    if ((/Edg\/\d+/).test(ua))
     {
         Browser.edge = true;
+        Browser.es2019 = true;
+    }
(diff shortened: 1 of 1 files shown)


photonstorm:
Yeah ok, I don't mind making this change now. Opera still used the old user agent when Phaser 3 was released, but we can swap to the OPR variant quite safely now I reckon.

ArtemSiz:
Thanks for the quick reply!

PR #6337, opened 2023-01-08
Add optional update argument to CanvasTexture methods

samme:
This PR

* Adds a new feature

This lets you skip updates during multiple draws so you can call `refresh()` or `update()` yourself when finished.

- `clear()`
- `draw()`
- `drawFrame()`


photonstorm:
👍

PR #6339, opened 2023-01-09
Upgrade spine-runtimes to 4.1.24 and fixed issues

justintien:
This PR

* Updates the Documentation
* Adds a new feature
* Fixes a bug

Describe the changes below:

1. Upgrade spine-runtimes to 4.1.24
2. Update script `plugin.spine.runtimes` comand
3. fix: SpinePlugin spineFileCallback support 7th parameter (settings { prefix }) (Avoid diff key but the atlas image filename same issue)
4. fix: Remove monkeypatch about setBlendMode

In fact, I based on v3.55.2...

But I can't PR based on v3.55.2, because you have no branch v3.55.2


photonstorm:
I assume this doesn't use the new Phaser Spine Runtimes that Esoteric built? https://github.com/EsotericSoftware/spine-runtimes/tree/4.1/spine-ts/spine-phaser/src

There's no 3.55.2 branch but I wouldn't merge even if there was, as once published you can't modify again anyway, so this will need to be a 3.60 release version of the plugin - and for that it will need to sit in its own plugins folder (`plugins/spine4`) please.

justintien:
> I assume this doesn't use the new Phaser Spine Runtimes that Esoteric built? https://github.com/EsotericSoftware/spine-runtimes/tree/4.1/spine-ts/spine-phaser/src
> 
> There's no 3.55.2 branch but I wouldn't merge even if there was, as once published you can't modify again anyway, so this will need to be a 3.60 release version of the plugin - and for that it will need to sit in its own plugins folder (`plugins/spine4`) please.

@photonstorm 

Spine official https://github.com/EsotericSoftware/spine-runtimes/tree/4.1/spine-ts/spine-phaser/src
It seems to be unfinished, and I used v3.55.2 in production now, so...

I close this PR.

spine4 folder is good idea. (Sorry I didn't think of that)

PR #6347, opened 2023-01-17
TilemapFilterCallback return value type fix

pavle-goloskokovic:
This PR:

* Updates the Documentation
* Fixes a bug

TilemapFilterCallback return value type was fixed to return `boolean` instead of `GameObject` and related docs updates.


photonstorm:
👍

PR #6348, opened 2023-01-17
Improve the missing frame/texture handling when creating animations

samme:
This PR

* Adds a new feature
* Fixes a bug

Fixed
-----

The "missing frame" console warning in `generateFrameNumbers()` was printing the loop index instead of the frame name.

Changed
-------

Before if you gave a bad texture `key` in an animation frame config, there was no console message and an invalid animation frame was created (undefined `frame`). Now a console warning is printed and no animation frame is created.

Before if you gave a bad texture key as `frames` in an animation config, there was no console message. Now a console warning is printed.

Before if you gave a bad texture key to `generateFrameNames()` or `generateFrameNumbers()` they would print a "missing frame" warning for each animation frame config. Now they print one "texture not found" warning and abort without trying to select frames.

### DIFF
--- a/src/animations/Animation.js
+++ b/src/animations/Animation.js
@@ -399,6 +399,13 @@ var Animation = new Class({
         {
             textureKey = frames;
 
+            if (!textureManager.exists(textureKey))
+            {
+                console.warn('Texture "%s" not found', textureKey);
+
+                return out;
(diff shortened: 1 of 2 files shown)


photonstorm:
👍

PR #6352, opened 2023-01-19
Fix Texture#has()

samme:
This PR

* Fixes a bug

It wasn't returning a boolean.


photonstorm:
Yeah I guess a strict boolean is more useful here 👍

PR #6355, opened 2023-01-20
Spatial sound

alxwest:

* Adds a new feature

Describe the changes below:

----

### New Features - Spatial Sound

Spatial sound takes advantage of the Web Audio API (https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API/Web_audio_spatialization_basics). This allows to configure sounds to come from a specific position or to track objects with x and y properties. Currently this is designed for 2D but could be 3D capable with a few modifications.

* Works only with `WebAudioSound` and `WebAudioSoundManager` and only in 2D (ignores ang changes to Z or Orientation).
* The `WebAudioSoundManager.setAudioDestination` method accepts any object with `x` and `y` properties. If this method is not used in conjunction with the source configuration, it will default to the center of the canvas.
* The `SoundConfig.source` is a new property that allows `WebAudioSound` to be configured in 2D space. The property can be set to either an object with `x` and `y` properties or a `SpatialSoundConfig`. If set to a object `WebAudioSound` will track it's position. If using  `SoundConfig.source` it can be set in the `SoundConfig` parameter when initializing `WebAudioSound` or when calling method `WebAudioSound.play`.

The following are examples of implementing the above.

* Adding a source coordinates on `WebAudioSound` initialization.

```js
var config = {
                source: {x:200, y:100}
             };
var music = this.sound.add('music', config);
```

* Adding a source object on `WebAudioSound.play`

```js
var enemy = this.image.add('tank');
rumbleSound.play({source:enemy});
```

* Adding the destination coordinates on `WebAudioSoundManager.setAudioDestination`

```js
game.sound.setAudioDestination({x:200, y:100});
```

* Adding a destination object on `WebAudioSoundManager.setAudioDestination`

```js
var player = this.image.add('mario');
game.sound.setAudioDestination(player);
```
----

[Examples for the feature can be found here](https://github.com/alxwest/phaser3-examples/tree/spatical-sound/public/src/3.60/audio/Web%20Audio)

* [spatial with destination coordinates.js](https://github.com/alxwest/phaser3-examples/blob/spatical-sound/public/src/3.60/audio/Web%20Audio/spatial%20with%20destination%20coordinates.js)
* [spatial with destination object.js](https://github.com/alxwest/phaser3-examples/blob/spatical-sound/public/src/3.60/audio/Web%20Audio/spatial%20with%20destination%20object.js)
* [spatial with source coordinates.js](https://github.com/alxwest/phaser3-examples/blob/spatical-sound/public/src/3.60/audio/Web%20Audio/spatial%20with%20source%20coordinates.js)
* [spatial with source object.js](https://github.com/alxwest/phaser3-examples/blob/spatical-sound/public/src/3.60/audio/Web%20Audio/spatial%20with%20source%20object.js)

### DIFF
--- a/src/sound/BaseSoundManager.js
+++ b/src/sound/BaseSoundManager.js
@@ -453,6 +453,18 @@ var BaseSoundManager = new Class({
     },
 
     /**
+     * Sets the destination for the spatial sound.
+     * Currently only WebAudio is supported.
+     *
+     * @method Phaser.Sound.BaseSoundManager#setAudioDestination
+     * @override
+     * @since 3.6.0
+     *
+     * @param {Phaser.Types.Sound.SpatialSoundConfig|object} [destination] - An object with x and y fields
+     */
+    setAudioDestination: NOOP,
+
+    /**
      * Stops all the sounds in the game.
      *
      * @method Phaser.Sound.BaseSoundManager#stopAll
--- a/src/sound/BaseSoundManager.js
+++ b/src/sound/BaseSoundManager.js
@@ -458,7 +458,7 @@ var BaseSoundManager = new Class({
      *
      * @method Phaser.Sound.BaseSoundManager#setAudioDestination
      * @override
-     * @since 3.6.0
+     * @since 3.60.0
      *
      * @param {Phaser.Types.Sound.SpatialSoundConfig|object} [destination] - An object with x and y fields
      */
--- a/src/sound/html5/HTML5AudioSoundManager.js
+++ b/src/sound/html5/HTML5AudioSoundManager.js
@@ -160,21 +160,6 @@ var HTML5AudioSoundManager = new Class({
     },
 
     /**
-     * Sets the destination for the spatial sound.
-     * Currently only WebAudio is supported.
-     *
-     * @method Phaser.Sound.BaseSoundManager#setAudioDestination
-     * @override
-     * @since 3.6.0
-     *
-     * @param {Phaser.Types.Sound.SpatialSoundConfig|object} [destination] - An object with x and y fields
-     */
-    setAudioDestination: function ()
-    {
-
-    },
-
-    /**
      * Unlocks HTML5 Audio loading and playback on mobile
      * devices on the initial explicit user interaction.
      *
--- a/src/sound/html5/HTML5AudioSoundManager.js
+++ b/src/sound/html5/HTML5AudioSoundManager.js
@@ -160,6 +160,21 @@ var HTML5AudioSoundManager = new Class({
     },
 
     /**
+     * Sets the destination for the spatial sound.
+     * Currently only WebAudio is supported.
+     *
+     * @method Phaser.Sound.BaseSoundManager#setAudioDestination
+     * @override
+     * @since 3.6.0
+     *
+     * @param {Phaser.Types.Sound.SpatialSoundConfig|object} [destination] - An object with x and y fields
+     */
+    setAudioDestination: function ()
+    {
+
+    },
+
+    /**
      * Unlocks HTML5 Audio loading and playback on mobile
      * devices on the initial explicit user interaction.
      *
--- a/src/sound/noaudio/NoAudioSoundManager.js
+++ b/src/sound/noaudio/NoAudioSoundManager.js
@@ -207,22 +207,6 @@ var NoAudioSoundManager = new Class({
     },
 
     /**
-     * Sets the destination for the spatial sound.
-     * Currently only WebAudio is supported.
-     *
-     * @method Phaser.Sound.BaseSoundManager#setAudioDestination
-     * @override
-     * @since 3.6.0
-     *
-     * @param {Phaser.Types.Sound.SpatialSoundConfig|object} [destination] - An object with x and y fields
-     */
-    setAudioDestination: function ()
-    {
-
-    },
-
-
-    /**
      * Stops any sounds matching the given key.
      *
      * @method Phaser.Sound.NoAudioSoundManager#stopByKey
--- a/src/sound/noaudio/NoAudioSoundManager.js
+++ b/src/sound/noaudio/NoAudioSoundManager.js
@@ -207,6 +207,22 @@ var NoAudioSoundManager = new Class({
     },
 
     /**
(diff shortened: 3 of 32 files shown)


src/sound/html5/HTML5AudioSoundManager.js
```diff
@@ -159,6 +159,21 @@ var HTML5AudioSoundManager = new Class({
...
+     *
+     * @method Phaser.Sound.BaseSoundManager#setAudioDestination
+     * @override
+     * @since 3.6.0
+     *
+     * @param {Phaser.Types.Sound.SpatialSoundConfig|object} [destination] - An object with x and y fields
+     */
+    setAudioDestination: function ()
+    {
+
```
photonstorm: There's no need for this method in here, because it's already in `BaseSoundManager`.

src/sound/BaseSoundManager.js
```diff
@@ -452,6 +452,18 @@ var BaseSoundManager = new Class({
         this.emit(Events.RESUME_ALL, this);
     },
 
+    /**
+     * Sets the destination for the spatial sound.
+     * Currently only WebAudio is supported.
+     *
+     * @method Phaser.Sound.BaseSoundManager#setAudioDestination
+     * @override
+     * @since 3.6.0
```
photonstorm: Please update from 3.6.0 to 3.60.0

src/sound/noaudio/NoAudioSoundManager.js
```diff
@@ -206,6 +206,22 @@ var NoAudioSoundManager = new Class({
...
+     *
+     * @method Phaser.Sound.BaseSoundManager#setAudioDestination
+     * @override
+     * @since 3.6.0
+     *
+     * @param {Phaser.Types.Sound.SpatialSoundConfig|object} [destination] - An object with x and y fields
+     */
+    setAudioDestination: function ()
+    {
+
```
photonstorm: As with HTML5Audio, there's no need for this method.

src/sound/webaudio/WebAudioSound.js
```diff
@@ -396,6 +400,17 @@ var WebAudioSound = new Class({
         return source;
     },
 
+    /**
+     * This method is only used internally and creates a PannerNode
+     *
+     * @method Phaser.Sound.WebAudioSound#createSpatialPanner
+     * @since 3.6.0
```
photonstorm: Please update from 3.6.0 to 3.60.0

src/sound/webaudio/WebAudioSoundManager.js
```diff
@@ -270,6 +270,24 @@ var WebAudioSoundManager = new Class({
         }
     },
 
+    /**
+     * Sets the destination for the spatial sound.
+     * Currently only WebAudio is supported.
+     *
+     * @method Phaser.Sound.BaseSoundManager#setAudioDestination
+     * @override
+     * @since 3.6.0
```
photonstorm: 3.60.0

src/sound/webaudio/WebAudioSoundManager.js
```diff
@@ -270,6 +270,24 @@ var WebAudioSoundManager = new Class({
         }
     },
 
+    /**
+     * Sets the destination for the spatial sound.
+     * Currently only WebAudio is supported.
+     *
+     * @method Phaser.Sound.BaseSoundManager#setAudioDestination
```
photonstorm, line 277: Should be `WebAudioSoundManager` not `BaseSoundManager` (and doesn't need override jsdoc tag)

No need to say 'only WebAudio' because this is the WebAudio class anyway.

Might be worth giving an example of exactly what the x/y coordinates relate to - i.e. which scale?

photonstorm (changes requested):
Thanks for this PR. I'm happy to merge this, but I think it would benefit from a couple of small changes (aside from those listed in the review).

1) Please ensure the panner node is destroyed properly in `WebAudioSoundManager.destroy` method, or it will leak memory.

2) I do wonder if we should only create the spatial panner node in the first place if the user requests it, rather than just making it by default for everyone (when most won't use it).

3) There is are getter/setters for pan, mute, etc in WebAudioSound and it feels like this ought to have an equivalent added too.

4) Only create the spatial panner node if the audio context supports it (like we check for `createStereoPanner` in the same way before making that)

alxwest:
> Please ensure the panner node is destroyed properly in WebAudioSoundManager.destroy method, or it will leak memory.

I followed the same logic as stereo panner node (confusingly `pannerNode`) in `WebAudioSound.destroy` `spatialNode` is also destroyed in the same way. I can't see anything in `WebAudioSoundManager.destroy` for the other nodes?

> I do wonder if we should only create the spatial panner node in the first place if the user requests it, rather than just making it by default for everyone (when most won't use it).

As I've designed spatial sound to be initialised in 3 places (new `WebAudioSound`, `WebAudioSound.play`, `WebAudioSoundManager.setAudioDestination`), I've opted for simplicity as chaining the audio nodes becomes convoluted.

NB calling `WebAudioSoundManager.setAudioDestination` without a spatial panner node present on the sound has no effect.

It might be better to create `2DWebAudioSound` for clarity?

> There is are getter/setters for pan, mute, etc in WebAudioSound and it feels like this ought to have an equivalent added too.

Is it alright to expose `SpatialSoundConfig` instead of the 14 properties, which I feel will bloat the class for properties the majority won't touch?

> Only create the spatial panner node if the audio context supports it (like we check for createStereoPanner in the same way before making that)

Done. 👍

photonstorm:
Ok I've merged this in and fixed a few issues, thanks.

One thing I've noticed is that the volume fall-off is _incredibly_ tiny. If the destination is say 400x300 and you set the source to just 380x300 then you can barely hear it. Set it at 200x300 and there is complete silence.

I assume this is a result of the defaults for the rolloff and cone angles? Are those the defaults the WebAudio API provides, or did you select them?

alxwest:
Correct, I corrected that with my branch (while working on my game incorporating 2D sound). I'll summit further PRs on the bugs I found and possible changes needed.

photonstorm:
Feel free to submit another PR, but just be aware I've modified the code quite a lot since your original one (it's all in the main repo)

PR #6359, opened 2023-01-20
Few type fixes

pavle-goloskokovic:
This PR 

* Updates the Documentation

Updates `fontSize` type to avoid errors when passing a number value withing `TextStyle` config when passed to `this.add.text` call, and adds missing `lineSpacing` property.

Optional `config` param for `Phaser.GameObjects.GameObjectCreator#graphics` method.

`ParticleEmitter` docs reference updated method name `setEmitterFrame`.


photonstorm:
👍

PR #6360, opened 2023-01-22
Set blocked.none in processX(), processY()

samme:
This PR

* Fixes a bug

These needed `blocked.none = false` when an edge was blocked.


photonstorm:
👍

PR #6382, opened 2023-02-02
Spatial sound

alxwest:

* Fixes a bug

Describe the changes below:

Fixes the sound "jumping" when playing/resuming.

Reduces redundant audio processing; IE if listener or follower's x  or y are updated if they've changed. It won't ask the web API to recalculate the "spatial pan/gain", which I've found can be processor heavy.


photonstorm:
👍

PR #6383, opened 2023-02-03
Change BaseCamera to AlphaSingle

samme:
This PR

* Fixes a bug

Not really a bug but BaseCamera had `alphaTopLeft` etc. properties that didn't do anything.


photonstorm:
👍

PR #6384, opened 2023-02-09
container#getBounds now also support getTextBounds for bounds calcula…

EmilSV:
* Fixes a bug

Describe the changes below:

container#getBounds now use getTextbounds for text gameobjects like BitmapText to for bounds calculations.

### DIFF
--- a/src/gameobjects/container/Container.js
+++ b/src/gameobjects/container/Container.js
@@ -412,16 +412,25 @@ var Container = new Class({
                 if (entry.getBounds)
                 {
                     entry.getBounds(tempRect);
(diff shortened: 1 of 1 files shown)


photonstorm:
Thanks for the PR. I'm not sure this order will work, however. As `BitmapText` has functions called `getBounds` _and_ `getTextBounds` so it'll just fall into the first condition and never the important one. I think the order needs swapping.

EmilSV:
Yes you are correct i swap the order now.

photonstorm:
👍

PR #6386, opened 2023-02-11
Fix renderFlags after changing scaleX/scaleY

mizunokazumi:
The value of renderFlags depends on the order in which scaleX/scaleY changes are made

```js
go.scaleX = 0
go.scaleY = 1
go.renderFlags === 15
```
```js
go.scaleY = 1
go.scaleX = 0
go.renderFlags === 11
```

Because of this, when changing the scale properties in pairs (for example, in a tween), the object passes the willRender check even if one of the scale properties is set to zero


photonstorm:
👍

Issue #6395, opened 2023-02-19
link in the shown javascript in line 5 is wrong.


Sathvik1007:
in file `phaser/src/physics/matter-js/typedefs/MatterWorldConfig.js` the 5th line link is not shown .
![aa](https://user-images.githubusercontent.com/124372859/219956615-d41a293e-6899-490a-905e-5c5bc907eac4.jpg)

photonstorm:
Thanks for reporting this. It's fixed in our JSDocs and now appears correctly: https://newdocs.phaser.io/docs/3.70.0/Phaser.Types.Physics.Matter.MatterWorldConfig

PR #6396, opened 2023-02-19
#6395 changed line into world

Sathvik1007:
the link in line 5 is not a link and there is only method regarding gravity in Phaser.Physics.Matter.World so i changed its link to that source.


Sathvik1007:
a

photonstorm:
Sorry, but this modification is incorrect. `Phaser.Physics.World` is for Arcade Physics, not Matter, but these are the Matter defs you have changed.

PR #6400, opened 2023-02-21
Adds tileOffset param to Phaser.Tilemaps.Tilemap#addTilesetImage

imothee:
This PR
* Adds a new feature
* Fixes a bug
 
Describe the changes below:

Currently addTilesetImage cannot set the tileOffset which means that changing the tileOffset requires you to update each tile after the Tileset is created.

This PR passes the tileOffset through to the Tileset constructor in Phaser.Tilemaps.Tilemap#addTilesetImage


photonstorm:
👍

PR #6414, opened 2023-03-05
correct the docs for generateFrameNumbers

ZekeLu:
This PR

* Updates the Documentation

Describe the changes below:

1. the function used in the example should be `this.anims.generateFrameNumbers` instead of `this.anims.generateFrameNames`.
2. the end frame in the example should be `11` instead of `12` (see https://github.com/photonstorm/phaser/commit/6761ec92f5c95d5e736c22cf4b3d3bb248b1ad90).
3. mark the "config" parameter as optional (the one for `generateFrameNames` is already marked as optional).


photonstorm:
👍

PR #6419, opened 2023-03-18
Fix ParticleEmitter#setParticleLifespan

kainage:

* Fixes a bug

Describe the changes below:

Fixes ParticleEmitter#setParticleLifespan which was calling `onChange` on the value instead of the `ops`


photonstorm:
👍

PR #6421, opened 2023-03-19
Fix error stopping a scene without a loader

samme:
This PR

* Fixes a bug (v3.60.0-beta.6)

Shutting down a scene without a LoaderPlugin would cause an error.

https://github.com/photonstorm/phaser/commit/0679f1aa649ee2c8eddf03ae94f09d3603b140ce


photonstorm:
👍

PR #6432, opened 2023-03-25
Change GetBounds() default output type to Rectangle

samme:
This PR

* Adds a new feature

This is more consistent with the rest of the API, IMO.


photonstorm:
Yeah fair enough.

PR #6436, opened 2023-03-26
Fix putTileAt() with empty tile

### LINKED ISSUE #6353: Tilemap putTilesAt crashes in 3.60.0-beta*
## Version
* Phaser Version: 3.60.0-beta.18

## Description
This is the error that I get when I switch to 3.60.0-beta.18 (or more in general to 3.60.0-beta*) while it works fine with 3.55.2.
I've put "debugger" before invoking "putTilesAt" from my code.

The zip file contains a Visual Studio Code project, just open "test.code-workspace" and npm install (for default I've set 3.60.0-beta.18 to see it running correctly "npm i phaser@3.55.2")
Then start the npm script "dev" and once compiled if you have Live Server extension just click "go live" and all should start (and then hang).
So open the developer tools and reload and the execution should pause at line 107 of src\js\pages\game\scenes\MazeScene.js
Please note that I've reduced this project in order to isolate the error so there are a lot of functions that points to nothing but obviously they are not necessary for showing the bug. 

If you follow the code you'll end up at the definition of PutTileAt somewhere inside Phaser library.
Near the end of that method you should find these 2 lines:
var sid = tiles[index][2];
var set = layer.tilemapLayer.tileset[sid];
the "sid" value is wrong ando set becomes undefined and then crashes at the next line.

I just don't know what I did wrong on the tileset definition that instead works perfectly with 3.55.2 and I haven't a clue. 
I've also made static the json that contains the layers data and you can find it on the console, I've done a console.log of it (there are many layers but it hangs at once on the first one named "colliders").

Thanks

## Example Test Code

The zip attached contains a Visual Studio Code project, just open "test.code-workspace" and npm install (for default I've set 3.60.0-beta.18 to see it running correctly "npm i phaser@3.55.2")

[_bug_putTilesAt.zip](https://github.com/photonstorm/phaser/files/10460589/_bug_putTilesAt.zip)

samme:
This PR

* Fixes a bug

Fixes #6353.

Adds `Phaser.Tilemaps.Tilemap#tiles`, avoiding calling `BuildTilesetIndex()` for each put tile.

### DIFF
--- a/src/tilemaps/Tilemap.js
+++ b/src/tilemaps/Tilemap.js
@@ -4,6 +4,7 @@
  * @license      {@link https://opensource.org/licenses/MIT|MIT License}
  */
 
+var BuildTilesetIndex = require('./parsers/tiled/BuildTilesetIndex');
 var Class = require('../utils/Class');
 var DegToRad = require('../math/DegToRad');
 var Formats = require('./Formats');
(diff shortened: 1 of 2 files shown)


photonstorm:
Thanks for this. I think it may be worth nulling the array in `Tilemap.destroy` too.

I've had a quick look to see if it's safe to index the tileset once in the constructor (i.e. check there isn't another method you could call that would invalidate this index), and I _think_ it's ok. Will look more carefully tomorrow.

photonstorm:
👍

PR #6453, opened 2023-04-05
Add methods to clear EmitZone or DeathZone

rexrainbow:

* Adds a new feature

Describe the changes below:

Add methods to clear EmitZone or DeathZone

### DIFF
--- a/src/gameobjects/particles/ParticleEmitter.js
+++ b/src/gameobjects/particles/ParticleEmitter.js
@@ -1731,6 +1731,21 @@ var ParticleEmitter = new Class({
     },
 
     /**
+     * Clear all Particle Death Zone from this Emitter.
+     *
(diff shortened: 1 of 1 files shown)


photonstorm:
Hmm, do we really need this along with `removeDeathZone` and `removeEmitZone` that we already have?

rexrainbow:
`removeDeathZone`, `removeEmitZone` require a zone object as parameter, it is not easy/directly to remove all zones. 
I use `emitter.emitZones.length = 0` in my test code. So I guess that it is nice to have clear method, or invoking `removeDeathZone`, `removeEmitZone` without no parameter to clear all zones is a good solution, too.

photonstorm:
👍

photonstorm:
Merged, but renamed to be plural.

PR #6456, opened 2023-04-06
Add `skipBatch` in StampConfig parameter of stamp method

rexrainbow:

* Adds a new feature

Describe the changes below:

Method `stamp` and `repeat` can draw a single frame or repeat frames. 
`repeat` has `skipBatch` parameter, but `stamp` does not, so I add `skipBatch` into StampConfig of  `stamp` method.

(BTW, method `draw` has *skipBatch* version `batchDraw`.)


photonstorm:
Sure, why not.

PR #6457, opened 2023-04-06
Change default callback scope to the timer event

samme:
This PR

* Adds a new feature

This is a breaking change but seems a lot more useful.


photonstorm:
👍

Issue #6398, opened 2023-02-20
3.60.0-beta.19 - adding empty BitmapText to container causes wrong container bounds


ultimoistante:
## Version

* Phaser Version: 3.60.0-beta.19

* Operating system:  Manjaro Linux

* Browser: tested on Chrome Version 110.0.5481.100 (64-bit), Firefox 110.0 (64-bit)

## Description

While testing an application I'm writing on 3.60.0-beta.18 against new release 3.60.0-beta.19, I've found an issue on container bounds when an empty BitmapText instance is added to a container. Using exactly the same code, these are the results of [parentContainer].getBounds() calls on both versions:

3.60.0-beta.18 **(working good)**:

{
    "type": 5,
    "x": 759.5000012516975,
    "y": 693.7500004470348,
    "width": 146.99999749660492,
    "height": 83.99999856948853
}

3.60.0-beta.19:

{
    "type": 5,
    "x": -45.33333333333333,
    "y": -18,
    "width": 951.8333320816358,
    "height": 795.7499990165234
}

photonstorm:
This is a result of  #6384 - any thoughts @EmilSV ?

EmilSV:
I did some testing and Phaser.GameObjects.BitmapText#getTextBounds or GetBitmapTextSize do not use it parent container for its bounds calculations.
To fix that you can either add it to container GetBounds like this:
```
    getBounds: function (output)
    {
        if (output === undefined) { output = new Rectangle(); }

        output.setTo(this.x, this.y, 0, 0);

        if (this.parentContainer)
        {
            var parentMatrix = this.parentContainer.getBoundsTransformMatrix();
            var transformedPosition = parentMatrix.transformPoint(this.x, this.y);

            output.setTo(transformedPosition.x, transformedPosition.y, 0, 0);
        }
        var matrix = this.getBoundsTransformMatrix();
        if (this.list.length > 0)
        {
            var children = this.list;
            var tempRect = new Rectangle();
            var tempVector = new Vector2();
            var hasSetFirst = false;

            output.setEmpty();

            for (var i = 0; i < children.length; i++)
            {
                var entry = children[i];

                if (entry.getTextBounds)
                {
                    var textBounds = entry.getTextBounds().global;
                    var tlX = textBounds.x;
                    var tlY = textBounds.y;
    
                    var brX = textBounds.x + textBounds.width;
                    var brY = textBounds.y + textBounds.height;
[...]

PR #6461, opened 2023-04-08
reverted using getTextBounds for cotainer getBounds()

EmilSV:
* Fixes a bug

Describe the changes below:
#6384
was leading to bugs #6398

I made container use getTextBounds() based on a fix for my own game.
But my game that was using an old version of the phaser 3.60 beta before BitmapText had its own getBounds() 6943989
I did some test and my change #6384 was unnecessary as commit 6943989 already solved my original problems.
I should have tested more before made the pull request

### DIFF
--- a/src/gameobjects/container/Container.js
+++ b/src/gameobjects/container/Container.js
@@ -407,28 +407,19 @@ var Container = new Class({
             {
                 var entry = children[i];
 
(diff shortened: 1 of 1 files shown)


photonstorm:
No worries, thanks for this PR to fix it again!

Issue #6425, opened 2023-03-22
ParticleEmitter renders flickering squares


EmilSV:
## Version
* Phaser Version: 3.60 beta 20

## Description
![billede](https://user-images.githubusercontent.com/3351370/227007287-888a7f09-9fb8-4351-a787-b64ade337d8c.png)
When a lot of Particle Emitter is on screen squares will begin to appear in weird patterns.
This only happen in the newest version of phaser after the updated particle system around beta 18

## Example Test Code
```javascript
class ParticlesSquare
    extends Phaser.Scene
{
     preload()
    {
        this.load.image("smokeTest", "https://labs.phaser.io/assets/particles/smoke0.png");
    }

     create()
    {

        const mainCam = this.cameras.main;
        const width = 1600;
        const height = 1200;

        mainCam.setBackgroundColor("#FFFFFF");
        mainCam.setZoom(0.50);
        mainCam.centerOn(width/2,height/2)

        for (let x = 0; x < width; x +=100)
        {
            for (let y = 0; y < height; y += 200)
            {
                this.add.particles(0, 0, "smokeTest", {
                    x,
                    y,
                    lifespan: { min: 3000, max: 5000 },
                    speed: { min: 5, max: 25 },
                    angle: { min: 270 - 10, max: 270 + 10 },
                    timeScale: 0.35,
                    gravityX: -2,
                    gravityY: -5,
                    scale: {
                        start: 0.33,
                        end: 1.2,
                        ease: Phaser.Math.Easing.Cubic.In
                    },
                    rotate: { min: 0, max: 360 },
                    alpha: { start: 0.6, end: 0 },
                    quantity: 1,
                    tint: [0x000000, 0x141414, 0x292929, 0x2e2e2e],
                    frequency: 25,
                    blendMode: Phaser.BlendModes.NORMAL,
                    particleBringToTop: true,
                });

            }
        }

    }
}
const config = {
    type: Phaser.AUTO,
    width: 800,
    height: 600,
    backgroundColor: '#000000',
    parent: 'phaser-example',
    scene: ParticlesSquare
};

const game = new Phaser.Game(config);
```

Gathaa:
Hi , the first issue maybe because of the hardware because rendering a lot of particles require good specs , also update the latest version of phaser then test it  .
another potential solution is to use a **Particle Container** in Phaser. A Particle Container is a container that can be used to group and render large numbers of particles efficiently 

Code :
[...]

PR #6462, opened 2023-04-09
fixed particle using a white texture if flush happened

EmilSV:
*  This PR Fixes a bug #6425

Describe the changes below:

When the pipeline runs flush it reset the texture and because "emitter.texture.glTexture" is undefined it could not do that probably.
But even if it could "textureUnit" would be wrong as it is cached before rendering the particles.
This PR tries to fix both issues by doing flush in "Phaser.GameObjects.Particles.Emitter#renderWebGL" and using "emitter.frame.glTexture" instead of "emitter.texture.glTexture"


photonstorm:
👍

PR #6471, opened 2023-04-14
fix: memory leak in WebAudioSoundManager

wjaykim:

* Fixes a bug

Describe the changes below:

`unlockHandler` was not removed because `this` was not binded to `bodyRemove` function, causing serious memory leak. As you see performance debugger with chrome, event listeners count get continuously increased as user presses keyboard.

This fixes the above bug by binding `document.body` to `bodyRemove` function.


photonstorm:
Merging, because I don't _think_ this has any side effects, but will keep an eye on it.

PR #6474, opened 2023-04-15
fix: Crash in PutTileAt

wjaykim:

* Fixes a bug

Describe the changes below:

If mutliple layer exists in tilemap and tilesets were added in each layers, calling putTileAt crashes the game because it tries to get tileset from its layers' own tileset array. Because tile index were gotten from tilemap.tiles, its index differ from index in layer specific tileset array.


photonstorm:
👍

PR #6480, opened 2023-04-19
BitmapText now reset texture unit on flush

EmilSV:
* Fixes a bug

Describe the changes below:
When a flush happen while BitmapText / DynamicBitmapText is rendering characters  the textureUint was not update.


photonstorm:
👍

PR #6499, opened 2023-05-10
avoid TypeError in addSpriteSheet method of TextureManager

charlieschwabacher:

* Fixes a bug

Describe the changes below:

The variable `texture` is set to null on line 1057.   On 1061 we attempt to read its `key` property which is not available, throwing a TypeError.  Instead we can use the key from the source argument as documented on line 1048.

### DIFF
--- a/src/textures/TextureManager.js
+++ b/src/textures/TextureManager.js
@@ -1058,7 +1058,7 @@ var TextureManager = new Class({
 
         if (source instanceof Texture)
         {
-            key = texture.key;
+            key = source.key;
             texture = source;
         }
         else if (this.checkKey(key))


samme:
What code produces the error?

greymd:
Agree with this PR.
It is clear that an error occurs when the `source` is an instance of Texture.
`key` does not exist as a null attribute.

https://github.com/photonstorm/phaser/blob/8c44dffcc7b5e1cfff92c22cfa639068c5f0cdd0/src/textures/TextureManager.js#L1055-L1063

> What code produces the error?

Attempting to create a SpriteSheet from an existing texture results in an error.

```typescript
new Phaser.Textures.TextureManager(this.game)
.addSpriteSheet(
  "key",
  this.textures.get("key"),
  {frameWidth: 300, frameHeight: 300, startFrame:0, endFrame: -1}
); // => got error
```

photonstorm:
👍

Issue #6501, opened 2023-05-11
Min/max width and height in game config are ignored by Phaser.Game constructor


NikitaShpanko:
## Version

* Phaser Version: **3.60.0**

* Operating system: **any** (found on Windows 11)

* Browser: **any** (found on Google Chrome)

## Description

According both to Phaser documentation and TypeScript type declarations, we can pass minimum and maximum game dimensions to the `Phaser.Game` constructor using `min` and `max` fields either in the game config itself or in its `scale` field.

Yet in the actual instance of `Phaser.Game` the respective parameters (`game.config.minWidth`, `game.config.minHeight`, `game.config.maxWidth`, `game.config.maxHeight`) remain on their default values (`0`) as if we didn't initialize them at all.
`

## Example Test Code

``` ts
const game = new Phaser.Game({
    min: {
        width: 640,
        height: 360,
    },
    max: {
        width: 1280,
        height: 720,
    },
    // we'll try 'em both just to be sure:
    scale: {
        min: {
            width: 640,
            height: 360,
        },
        max: {
            width: 1280,
            height: 720,
        },
    },
})

console.log(game.config.minWidth, game.config.minHeight) // expected:  640 360, actual: 0 0
console.log(game.config.maxWidth, game.config.maxHeight) // expected: 1280 720, actual: 0 0
```

## Additional Information

As far as I understand the way to fix this bug is to alter these lines in the source code:
https://github.com/photonstorm/phaser/blob/master/src/core/Config.js#L107
https://github.com/photonstorm/phaser/blob/master/src/core/Config.js#L112
https://github.com/photonstorm/phaser/blob/master/src/core/Config.js#L117
https://github.com/photonstorm/phaser/blob/master/src/core/Config.js#L122

from
``` ts
this.minWidth = GetValue(scaleConfig, 'minWidth', 0, config);
this.maxWidth = GetValue(scaleConfig, 'maxWidth', 0, config);
this.minHeight = GetValue(scaleConfig, 'minHeight', 0, config);
this.maxHeight = GetValue(scaleConfig, 'maxHeight', 0, config);
```
to
``` ts
this.minWidth = GetValue(scaleConfig, 'min.width', 0, config);
this.minHeight = GetValue(scaleConfig, 'min.height', 0, config);
this.maxWidth = GetValue(scaleConfig, 'max.width', 0, config);
this.maxHeight = GetValue(scaleConfig, 'max.height', 0, config);
```

wpederzoli:
[...]

PR #6502, opened 2023-05-11
add proper key value to scale width and height

wpederzoli:

* Fixes a bug

Describe the changes below:
This should fix the bug described here:
https://github.com/photonstorm/phaser/issues/6501

When getting the `scale` from the config object, we're currently not getting the proper key. This PR introduces getting `min.width, min.height, max.width, max.height`


photonstorm:
👍

PR #6523, opened 2023-06-12
Changes Tiled Map properties value to an array instead of object

likwidgames:

* Fixes a bug

Describe the changes below:

Tiled outputs Map properties as an array of objects, rather than a singular object.


photonstorm:
Some versions of Tiled do, yes, but not all. So I cannot merge this as it'll break as many versions as it helps fix. This needs to be conditional and checked for in the parser.

Plus, you'll need to undo all of the eslint breaking modifications please (quotes, etc)

photonstorm:

PR #6533, opened 2023-06-28
Timeline restart method

monteiz:
This PR

* Updates the Documentation
* Adds a new feature

Describe the changes below:

# Pull Request: Restart Method for Timeline Object

## Description
This pull request adds a new method called `restart` to the Timeline object. The `restart` method resets the timeline back to the start by setting the elapsed time to zero and marking all events as incomplete. After resetting, the timeline is automatically started by calling the `play` method.

## Motivation
The addition of the `restart` method provides a convenient way to reset and restart a timeline without needing to call separate methods. This improves the usability and readability of the Timeline object.

## Changes Made
- Added a new method `restart` to the Timeline object.
- The `restart` method resets the elapsed time to zero.
- All events in the timeline are marked as incomplete.
- The `restart` method calls the `reset` method to handle the reset logic.
- After resetting, the timeline is automatically started by calling the `play` method.

## Usage Example
```javascript
const timeline = this.add.timeline();

// Add timeline events...

timeline.restart();

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "phaser",
-  "version": "3.60.0",
+  "version": "3.60.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "phaser",
-      "version": "3.60.0",
+      "version": "3.60.1",
       "license": "MIT",
       "dependencies": {
         "eventemitter3": "^5.0.0"
--- a/src/physics/arcade/World.js
+++ b/src/physics/arcade/World.js
@@ -1009,6 +1009,21 @@ var World = new Class({
     },
 
     /**
+     * Advances the simulation by a single step.
+     *
+     * @method Phaser.Physics.Arcade.World#step
+     * @fires Phaser.Physics.Arcade.Events#WORLD_STEP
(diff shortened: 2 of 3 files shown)


photonstorm:
Thanks for your PR. Having a quick look at it - the `restart` method you've added calls `reset()` and then `play()`.

However, calling `reset()` on its own automatically calls `play` already:

https://github.com/photonstorm/phaser/blob/master/src/time/Timeline.js#L439-L449

This means via the new `restart` method, what actually happens is:

1) The Timeline is reset, which calls `play` automatically.
2) `play` is called again via the `restart` method but with no `fromStart` parameter, meaning it will then call `reset` again for a second time, which will then call `play` again for a third time :)

monteiz:
Yeah, sorry, it was a quick one :P

So, [here](https://github.com/photonstorm/phaser/blob/fb3c85888b444b5cc0e5ba6d0262adf0cda76bd2/src/time/Timeline.js#L153), you wanted to write `.reset` and not `.restart`?

photonstorm:
Yeah. It used to be `restart` but I changed it and forgot to update the docs.

monteiz:
I see :)
Well, we spotted a typo and that's good ;)

Issue #6541, opened 2023-07-07
Make Tiled tile layer ID accessible from Phaser.Tilemaps.LayerData


rui-han-crh:
In Tiled, unique layers are identified by their layer ID. Two layers may have the same name in Tiled, but can be disambiguated via their ID. While all other properties, such as name, x, y, alpha, height and width can be accessed from a `LayerData` object, the ID is seeming not parsed from the map.

### Request
To make the id property accessible from the LayerData, after creation of a tilemapLayer via `createLayer`.

### Example usage

```ts
const tilemapLayer = this.tilemap.createLayer(this.tilemap.layers[i].name, this.tilemap.tilesets) as TilemapLayer;

console.log(tilemapLayer.layer.name); // -> Logs name as shown in Tiled
console.log(tilemapLayer.layer.id); // -> Logs unique layer id as shown in Tiled
```

photonstorm:
Thanks for opening this issue, and for submitting a PR to fix it. We have merged your PR into the `master` branch and attributed the work to you in the [Change Log](https://github.com/photonstorm/phaser/blob/master/CHANGELOG.md). If you need to tweak the code for whatever reason please submit a new PR.

PR #6542, opened 2023-07-07
Expose layer id from Tiled map parser in LayerData and ObjectLayer

rui-han-crh:

This PR

* Adds a new feature

Describe the changes below:

Before this change, it was not possible to access a Tiled layer (either tilemap layer or object layer) ID directly from the LayerData or ObjectLayer.

This PR adds a new field called `id` to both LayerData and ObjectLayer, as well as adds one additional property in the layer data config, `id`, to retrieve the id property from the already parsed map data.

This would allow users to be able to access the unique layer id of Tiled layers and use it for disambiguation purposes, in the event that a map is given with non-unique layer names.

Related issue: #6541

### DIFF
--- a/src/tilemaps/mapdata/LayerData.js
+++ b/src/tilemaps/mapdata/LayerData.js
@@ -39,6 +39,15 @@ var LayerData = new Class({
         this.name = GetFastValue(config, 'name', 'layer');
 
         /**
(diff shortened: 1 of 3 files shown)


photonstorm:
👍

PR #6545, opened 2023-07-14
Add backgroundAlpha property in Circle effect  controller

rexrainbow:

* Adds a new feature

Describe the changes below:

Add `backgroundAlpha` property in Circle effect  controller, so that background could be transparent (set `backgroundAlpha` to `0`)


photonstorm:
👍

PR #6555, opened 2023-07-21
JSDocs update

AlvaroEstradaDev:
Updated the JSDocs for the particle emitter to fix an incompatibility issue with its typescript constructor description.


photonstorm:
Yeah good catch. The parameters are optional, but I'll fix that after I merge.

PR #6561, opened 2023-07-26
fix JSDoc fired event: Phaser.Loader.LoaderPlugin#ADD -> Phaser.Loader.Events#ADD

julescubtree:

* Updates the Documentation

Describe the changes below:

The JSDoc pointed to the non-existent `Phaser.Loader.LoaderPlugin#ADD` event, so I've changed it to the correct `Phaser.Loader.Events#ADD`.


julescubtree:
I only updated the */src/* files because I assume there's a build step you take care of for the */dist/* files.

photonstorm:
Nice catch, thanks.

And yes, you're right - please never commit dist files!

PR #6569, opened 2023-08-02
Added two optional parameters to GameObjects.Components.Texture#setTexture

Trissolo:
This PR

* Adds two optional parameters to GameObjects.Components.Texture#setTexture

Describe the changes below:

The params are `updateSize` and `updateOrigin` and are used as the second and third arguments in the `setFrame()` call. This gives you full control when setting the frame, rather than just specifying which frame to use.


photonstorm:
Yeah ok, why not.

PR #6573, opened 2023-08-06
remove duplicate declarations

allowing:


photonstorm:
Thanks for the PR, but if I merge this, it will completely break the use of the RTree within Arcade Physics, causing collisions to fail.

PR #6576, opened 2023-08-08
Update return type of emitParticle

AlvaroEstradaDev:
The function emitParticle can also return undefined when the particle max limit has been reached.


photonstorm:
👍

PR #6577, opened 2023-08-09
Fix Geom.Intersects#GetLineToPoints

### LINKED ISSUE #6467: GetLineToPolygon doesn't work as expected
Hello,

I don't know if it's a mistake from my side or not, but it seems  that Phaser.Geom.Intersects.GetLineToPolygon does not work with the last vertice (the one which is not closed) of the polygon.

Here's an example 
https://codepen.io/Abspirit/pen/PoyPNOK

the green dot is not supposed to be here or I did not get what it's supposed to do ?
And I also don't get why when I use the other "line4" (quoted in the example) it does not work.
(because X1 = X2 ?). I'm lost.

Thanks for your answer

Trissolo:
This PR

* Fixes a bug

An oversight in the for loop prevented the intersection test between the given `Geom.Line` and the line segment between the first and last point.

Also fixes  #6467

### DIFF
--- a/src/geom/intersects/GetLineToPoints.js
+++ b/src/geom/intersects/GetLineToPoints.js
@@ -46,9 +46,9 @@ var GetLineToPoints = function (line, points, isRay, out)
     out.set();
     tempIntersect.set();
 
-    var prev = points[0];
+    var prev = points[points.length - 1];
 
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

PR #6578, opened 2023-08-10
Uses current faces to do hit-testing

rexrainbow:

* Fixes a bug

Describe the changes below:

Hit-testing of Mesh won't work if faces is changed after invoking `setInteractive()`. Uses current faces to do hit-testing could eliminate this bug.


photonstorm:
👍

PR #6584, opened 2023-08-16
Remove unused code in Geom.Polygon#setTo

Trissolo:
This PR

* Removes unused code

In the `Geom.Polygon#setTo` method, the variable `y0`, which temporarily stores the upper bound of the Polygon, has no real use. Hence I removed the related code.


photonstorm:
👍

PR #6587, opened 2023-08-17
Added missing description of the anims property of the particle emitter for JSDoc

AlvaroEstradaDev:
- Created a description of the object handed to the function setAnim
- Added the property anim to ParticleEmitterConfig
- Changed the parameters description of setAnim to match better


photonstorm:
👍 the typedef would benefit from having the default values added to it, but I'll do that here.

PR #6596, opened 2023-08-25
Timeline improvement

rexrainbow:

* Adds a new feature

Describe the changes below:

- Do nothing in `update()` method if `timeline.complete` is `true`
- Add `COMPLETE` event, fired when `timeline.complete` is `true`, after [events execution loop](https://github.com/photonstorm/phaser/blob/master/src/time/Timeline.js#L252).

Although user can fire `complete` inside a timeline event.
However, it might not be a good idea that `clear`, or `add` new timeline events during events execution loop (i.e. change items of `timeline.events`), thus I add `COMPLETE` event fired after events execution loop.


photonstorm:
A few things I need to fix in here, but I like the features, so merging :)

PR #6602, opened 2023-08-30
Update InputConfiguration.js: Clarify alphaTolerance

johnhyde:
This PR
* Updates the Documentation

Describe the changes below:
Adds documentation to make it clear that the max alphaTolerance is 255, rather than 1 (as in setAlpha functions).


photonstorm:
👍

PR #6608, opened 2023-09-08
Added the use of Vector2Like

AlvaroEstradaDev:
Changed the types of follow and followOffset to use the interface Vector2Like, this enables more objects to be used and not only gameobjects and better describes followOffset


photonstorm:
👍

PR #6610, opened 2023-09-08
EmitterOpRandomMinMaxConfig type correction

AlvaroEstradaDev:
Added the description of the "int" boolean property that was missing

### DIFF
--- a/src/gameobjects/particles/typedefs/EmitterOpRandomMinMaxConfig.js
+++ b/src/gameobjects/particles/typedefs/EmitterOpRandomMinMaxConfig.js
@@ -6,4 +6,5 @@
  *
  * @property {number} min - The minimum value.
  * @property {number} max - The maximum value.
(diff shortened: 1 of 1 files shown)


samme:
Shouldn't it be

> If true, only integers are selected from the range

or something similar?

AlvaroEstradaDev:
> Shouldn't it be
> 
> > If true, only integers are selected from the range
> 
> or something similar?

I took "EmitterOpRandomStartEndConfig.js" description of random as reference.

samme:
But that's completely different. Here `int: true` means force the random value to be an integer.

AlvaroEstradaDev:
> But that's completely different. Here `int: true` means force the random value to be an integer.

Oh, yeah, I misread the code. I'm gonna change it.

photonstorm:
👍

PR #6612, opened 2023-09-11
Update Path.js

AlvaroEstradaDev:
* Added a defaultDivisions property, this brings it to parity with singular curves implementation
* Added a stepRate parameter to the getPoints method, so it can pass this parameter to the curve's getPoints method when given


photonstorm:
👍

PR #6613, opened 2023-09-11
Accessing tiles out of layer bounds

XT60:
# Error I was getting
`phaser/src/tilemaps/parsers/tiled/ParseTileLayers.js`
Function: `ParseTileLayers`
Line: `188`
Error: `Cannot set properties of undefined (setting '16')`

# Cause
Tiled Map Editor produces chunks that are aligned on a 'chunk grid' starting at (0, 0) and increasing in size by the dimensions of the chunks. Since not all layers have to align perfectly with this grid, some parts of the chunks within a layer might extend beyond the layer bounds. This, in turn, caused the parser to access array indexes that were either negative or greater than its length

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "phaser",
-  "version": "3.61.0-beta.2",
+  "version": "3.61.0-beta.3",
   "release": "Yotsuba",
   "description": "A fast, free and fun HTML5 Game Framework for Desktop and Mobile web browsers.",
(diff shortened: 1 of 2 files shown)


samme:
Is this a [Tiled Infinite Map](https://doc.mapeditor.org/en/stable/manual/using-infinite-maps/)?

XT60:
> Is this a [Tiled Infinite Map](https://doc.mapeditor.org/en/stable/manual/using-infinite-maps/)?

yes

PR #6619, opened 2023-09-13
EmitterOpOnEmitCallback update

AlvaroEstradaDev:
Changed the parameters to be optional since not all the onEmit calls actually pass parameters to the function, so there should be a warning about that.


photonstorm:
👍

PR #6620, opened 2023-09-14
Type clarification for emit zone data

AlvaroEstradaDev:
This PR

* Updates the Documentation

Describe the changes below:

* Updated the definition of EmitZoneObject to exclusively represent EdgeZone or RandomZone objects, excluding their configurations. Previously, the emitZones property was either an array of RandomZones or an array of EdgeZones, but not an array that could hold either. With this change, emitZones is now represented as an array of EmitZoneObjects.
* EmitZoneData takes over the role previously held by EmitZoneObject.


photonstorm:
👍

PR #6623, opened 2023-09-17
Add setDisplaySize method in BitmapText

rexrainbow:

* Adds a new feature

Describe the changes below:

Since BitmapText does not have original size, I set scaleX, scaleY to `1` to get original size, 
then calculate new scaleX, scaleY by displayWidth, displayHeight.

### DIFF
--- a/src/gameobjects/bitmaptext/static/BitmapText.js
+++ b/src/gameobjects/bitmaptext/static/BitmapText.js
@@ -860,7 +860,7 @@ var BitmapText = new Class({
      *
      * Calling this will adjust the scale.
      *
-     * @method Phaser.GameObjects.Components.Size#setDisplaySize
+     * @method Phaser.GameObjects.BitmapText#setDisplaySize
      * @since 3.61.0
      *
      * @param {number} width - The width of this BitmapText Game Object.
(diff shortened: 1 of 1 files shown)


photonstorm:
Does it fix something to add this? Or is it just a feature?

rexrainbow:
Since `displayWidth`, `displayHeight` properties does not have setter, I thought it is a new feature.

Because that BitmapText has `scaleX`, `scaleY` properties, it can have setter of `displayWidth`, `displayHeight` propertie. 
But BitmapText does not store original size (size when scale is `1`) directly, I have to get this original size first. This solution is not straightforward I agree.

photonstorm:
I think I'd rather just store the original size somewhere and use that. Although I guess it will need resetting every time you change the font size, for example.

rexrainbow:
> I think I'd rather just store the original size somewhere and use that. Although I guess it will need resetting every time you change the font size, for example.

It is a better solution.

zekeatchan:
Hi @rexrainbow. Thanks for the PR. We have merged it with the `master` branch. It will be part of the next release. Do test it out and let us know if you encounter any issues.

PR #6627, opened 2023-09-21
Add read-only originX, originY properties

rexrainbow:

* Adds a new feature

Describe the changes below:

Plane game object can have read-only originX, originY properties.


photonstorm:
👍

PR #6634, opened 2023-09-23
Update matter.d.ts

paxperscientiam:
This PR adds a missing property to the MatterJS BodyType type.

Describe the changes below:
* I added BodyType.deltaTime: number to match property change as described here: https://github.com/photonstorm/phaser/discussions/6452


photonstorm:
👍

PR #6637, opened 2023-09-25
Removed interface EmitterOpRandomStartEndConfig

AlvaroEstradaDev:
This PR

* Updates the Documentation

Describe the changes below:

* The interface EmitterOpRandomStartEndConfig, even though it is another path to configure a random emit value, does not fully describe that all the options for the EmitterOpEaseConfig are available. Instead, it is better to add to the description of the EmitterOpEaseConfig that there is a random start value option available.


photonstorm:
👍

PR #6641, opened 2023-10-01
Add getter of tint property

rexrainbow:

* Adds a new feature

Describe the changes below:

Since setter of `tint` will set `tintTopLeft`, `tintTopRight`, `tintBottomLeft`, `tintBottomRight` together, i.e. value of these 4 properties are equal. So that it can return `tintTopLeft` value as the getter of `tint` property, imo.


photonstorm:
👍

PR #6668, opened 2023-11-13
Update GetPoints.js

### LINKED ISSUE #6439: Better calculation in Geom.Line.GetPoints and Actions.Spread
## Version

* Phaser Version: 3.55.2

* Operating system: Windows

* Browser: 

## Description

Hello,

I have the feeling that Phaser.Geom.Line.GetPoints does not work quite like people expect as it does not include the end coordinate of the line in the results.
If I have a line between (0, 0) and (10, 0) and asks 3 points on that line I would expect [(0, 0), (5, 0), (10, 0)]
instead, I get [(0, 0), (3.33, 0), (6.66, 0)]

This is unintuitive, especially when trying to use Actions.PlaceOnLine() to distribute objects evenly on half the screen for example.

Same issue with Actions.Spread(), if I spread the alpha value from 0 to 1, no object will have an alpha of 1 and of course it is more noticeable with a small set of objects. It can't be spotted easily in the Phaser3 examples repo due to the large number of objects used in the PlaceOnLine and Spread examples.

I hope I did not miss something.

## Example Test Code

```JavaScript
const link1 = this.add.text(0, 0, "Episode 1").setOrigin(.5)
const link2 = this.add.text(0, 0, "Episode 2").setOrigin(.5)
const link3 = this.add.text(0, 0, "Episode 3").setOrigin(.5)

const line = new Phaser.Geom.Line(width/4, height/4*3, width/4*3, height/4*3)
Phaser.Actions.PlaceOnLine([link1, link2, link3], line);
Phaser.Actions.Spread([link1, link2, link3], 'alpha', 0, 1);
```

## Additional Information

I think the current calculation method makes sense for closed shapes (circle, ellipse, rectangle) as you would not want duplicate coordinates for the start and end point in most cases, but for open shapes like lines or curves having the end point is important.

This logic produce the expected result:
```Javascript
function GetPoints(x1, y1, x2, y2, quantity) {

    const dx = x2 - x1;
    const dy = y2 - y1;
    const distance = Math.sqrt(dx * dx + dy * dy)
    const step_size = distance / (quantity - 1)

    let points = [[x1, y1]]

    for (let i = 1; i < quantity; i++) {
        position = i / quantity
        x = x1 + (dx / distance) * (step_size * i)
        y = y1 + (dy / distance) * (step_size * i)

        points.push([x, y])
    }
    return points
}

let p = GetPoints(10, 0, 40, 0, 3)
```
image

sB3p:
This PR

* Adds a new feature
* Updates existing code

Fixes #6439 partially based on @AlitaStudio proposed solution.
This PR only applies to `Geom.Line.GetPoints` calculation metioned in the issue. `Actions.Spread` calculation is not affected by this PR.
`Geom.Line.GetPoints` and `Phaser.Actions.PlaceOnLine` are affected by this PR.
Merging this PR will potentially cause issues for users later on (after updating Phaser version) as this will change the x/y coords of GameObjects that are placed on a `Phaser.Geom.Line` via `Phaser.Actions.PlaceOnLine`.

Example of how `Phaser.GameObjects.GameObject` were placed on a line before this change:
place_on_line_before

Example of how `Phaser.GameObjects.GameObject` will be placed on a line after this change:
place_on_line

### DIFF
--- a/src/geom/line/GetPoints.js
+++ b/src/geom/line/GetPoints.js
@@ -43,12 +43,16 @@ var GetPoints = function (line, quantity, stepRate, out)
     var x2 = line.x2;
     var y2 = line.y2;
 
+    var dx = x2 - x1;
+    var dy = y2 - y1;
+
+    var distance = Math.sqrt(dx * dx + dy * dy);
+    var stepSize = distance / (quantity - 1);
+
     for (var i = 0; i < quantity; i++)
     {
-        var position = i / quantity;
-
-        var x = x1 + (x2 - x1) * position;
-        var y = y1 + (y2 - y1) * position;
+        var x = x1 + (dx / distance) * (stepSize * i);
+        var y = y1 + (dy / distance) * (stepSize * i);
 
         out.push(new Point(x, y));
     }


rgk:
The feature was added to `Actions.PlaceOnLine` when supplied with an ease function instead of to `Geom.Line.GetPoints` to reduce breaking changes in the master branch.

Note: Use `Geom.Line.GetEasedPoints` with the `linear` ease function for the results you are looking for.

PR #6670, opened 2023-11-14
Added correct parameters for create

stevenwithaph:
This PR 

* Updates the Documentation

Describe the changes below:

Added correct parameters to TextureManager.create, source can take an either a HtmlIamgeElement, or an array.
Width and height are optional.


photonstorm:
Good catch. Can also take a canvas element or array of them! Will update for that locally.

PR #6672, opened 2023-11-15
Correcting animation frame-step when all frames have duration value provided

TomMalitz:
This PR:

* Fixes a bug

Describe the changes below:

I noticed the total animation duration was not correct when defining an animation with frames that each have a duration value provided.

For example, for a 5-frame animation with each frame's duration set to 100ms I was seeing a total animation duration of 1000ms instead of 500ms. I looked into the source and saw that the next tick for the frame-step is always adding the calculated `msPerFrame` to the `currentFrame.duration` (100ms + 100ms in this case; resulting in 200ms for each frame, and 1000ms for the entire animation).

I made an update to the logic that takes into consideration if all the frames for the animation have duration values provided. In this case, the calculated `msPerFrame` is not needed when determining the `nextTick` value.

This change results in the example 5-frame animation correctly taking 500ms overall, with 100ms for each frame.

I have tested and verified the change in my own test project, but please take a look and let me know if there is anything I am not considering.

Phaser is amazing, thanks!

### DIFF
--- a/src/animations/Animation.js
+++ b/src/animations/Animation.js
@@ -114,9 +114,9 @@ var Animation = new Class({
          * @type {boolean}
          * @since 3.0.0
          */
-        this.allFramesHaveDuration = !this.frames.some(function (frame)
+        this.allFramesHaveDuration = this.frames.every(function (frame)
         {
-            return frame.duration === undefined;
+            return frame.duration !== undefined;
         });
 
         /**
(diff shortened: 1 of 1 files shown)


rgk:
The issue #6712 is related to this and has been solved.

Sorry your PR was not used, but it helped us solve the issue, thank you! You have been credited in the Change Log

PR #6676, opened 2023-11-21
Remove DESTROY event before destroy child game object

### LINKED ISSUE #6675: GameObjects added to Layers are destroyed twice on Scene shutdown causing crashes
## Version

* Phaser Version: 3.70.0

* Browser: 

## Description

Adding a GameObject to a Layer throws unhandled exceptions when the Scene they're added to shuts down (such as when transitioning to a new scene). 

The Scene destroys the Layer, then the Layer destroys the Sprite which causes it to emit its DESTROY event. This event is caught by the Layer which removes the Sprite from its list as well as attempts to destroy it again because the Layer is being destroyed by the scene. The destroy attempted by the Layer throws an exception in Sprite.preDestroy when this.anims is null because it was already removed by the original destroy.

## Example Test Code

https://codepen.io/crockergd/pen/RwvMxNG

The demo attempts to add a Sprite to a Layer, then transition to a different Scene which displays a small green square. Currently, the demo crashes while trying to stop the current scene. The demo completes successfully if the line adding SceneA's rect to the layer is removed, or if the Phaser dependency is set to a version before 3.70.0.

## Additional Information

The exception thrown by the demo doesn't trace the source of the issue well, I've attached an image of a call stack from a debugger as well.
 
layercrashstack

rexrainbow:

* Fixes a bug

Describe the changes below:

Remove DESTROY event before destroy child game object. Like [`removeAll()`](https://github.com/photonstorm/phaser/blob/master/src/gameobjects/layer/Layer.js#L690) does.

Fix #6675

### DIFF
--- a/src/gameobjects/layer/Layer.js
+++ b/src/gameobjects/layer/Layer.js
@@ -1053,6 +1053,8 @@ var Layer = new Class({
 
         while (list.length)
         {
+            list[0].off(GameObjectEvents.DESTROY, this.remove, this);
+
             list[0].destroy(fromScene);
         }
 


rexrainbow:
Closed this PR as duplication #6667

samme:
This looks like a simpler fix than #6667

PR #6686, opened 2023-12-05
Fix removing destroyed Container children

samme:
This PR

* Fixes a bug

There were two problems with destroying Container children. I fixed them by hooking the child's DESTROY event to a new `onChildDestroyed()` method instead of `remove()`.

1.  If you destroyed a game object in an exclusive Container, the game object would (momentarily) move onto the scene display list and emit an ADDED_TO_SCENE event.

2.  If you added a Sprite to a nonexclusive Container and stopped the scene, you would get a TypeError (evaluating 'this.anims.destroy'). This happened because the `fromChild` argument in the DESTROY event was misinterpreted as `destroyChild` in the Container's `remove()`, and the Container was calling the Sprite's `destroy()` again.

### DIFF
--- a/src/gameobjects/container/Container.js
+++ b/src/gameobjects/container/Container.js
@@ -438,7 +438,7 @@ var Container = new Class({
      */
     addHandler: function (gameObject)
     {
-        gameObject.once(Events.DESTROY, this.remove, this);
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

PR #6694, opened 2023-12-13
Check for undefined to avoid a crash in the examples at phaser3-custo…

Byvire:
…m-build.

* Fixes a bug

Describe the changes below:

The test/index.html and other examples at
https://github.com/photonstorm/phaser3-custom-build
crash because it tries to access `sys.scenePlugin._target` when `sys.scenePlugin` is `undefined`.
When I made this change locally it got them working again.


rgk:
Simple and works 👍

PR #6715, opened 2024-01-15
Flip order of tile visibility check and camera culling

zegenie:
This PR
* Fixes a bug with the camera culling of isometric tiles by changing the order of checks to improve rendering speed

Describe the changes below:
In our game with 200x200 tiles, tile culling takes up a significant portion of render time. We tried disabling the visibility of certain tiles as a speedup measure, but realized that the culling will not check the visibility of the tile until *after* it has checked if it's in bounds. Since the bounds check is significantly heavier than the `visible` flag check, reversing the order of the checks in the code so it checks the flags before doing the camera bounds check noticeably decreases render time. On some of our lower specced computers (where the cpu is a potential bottleneck), average cull time changes from ~10ms (without the patch) to ~5ms (with the patch).

Unless there is something I'm missing, this seems like a trivial change that can decrease the time phaser spends in the render loop.

### DIFF
--- a/src/tilemaps/components/IsometricCullTiles.js
+++ b/src/tilemaps/components/IsometricCullTiles.js
@@ -50,17 +50,19 @@ var IsometricCullTiles = function (layer, camera, outputArray, renderOrder)
         {
             for (x = drawLeft; x < drawRight; x++)
             {
+                tile = mapData[y][x];
+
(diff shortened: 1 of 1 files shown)


rgk:
In your pull request the `skipCull` flag and `CheckIsoBounds` method were reversed in function. This has been fixed.

Sorry your PR was not used, but it helped us solve the issue, thank you! You have been credited in the Change Log.

PR #6719, opened 2024-01-27
docs: clarify overlapOnly param

Stan-Stani:
in Phaser.Physics.Arcade.Collider

Explain that the callback will trigger on any overlap with any part of a tileLayer

* Updates the Documentation


rgk:
If you add this note to @classdesc, I believe that would be better suited.

Stan-Stani:
@rgk Like this?

photonstorm:
👍

PR #6722, opened 2024-01-29
Improve the warnings in addTilesetImage()

samme:
This PR

* Adds a new feature

I tried to clarify the two warning messages, and the second one now prints the valid tilesets if you pass a bad tileset name. Should save some developer anguish.


rgk:
:v:

PR #6723, opened 2024-01-29
Exclude the renderers by feature flag

samme:
This PR

* Adds a new feature

This includes `Phaser.Renderer.Canvas` or `Phaser.Renderer.WebGL` only if the corresponding feature flags `CANVAS_RENDERER` or `WEBGL_RENDERER` are true.

For Canvas-only builds this saves a lot.

### DIFF
--- a/src/renderer/index.js
+++ b/src/renderer/index.js
@@ -14,9 +14,17 @@
 
 module.exports = {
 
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

PR #6734, opened 2024-02-09
File creation checks if URL is a string before seeing if its Base64

### LINKED ISSUE #6733: 3.80.0 beta 2 - File.base64 check throws exception if url is not a string
On this line https://github.com/phaserjs/phaser/blob/a4d66864aecf94bccd8c94bcec94aad26ca00466/src/loader/File.js#L243

`url` maybe an object (like when use loader.multiatlas), it throws an exception. Should it check if it's string here first?

yaustar:
This PR

* Fixes a bug

Fixes #6733 where `File.url` can be a object with the JSON data or a string but is assumed to only be a string when checking if a Base64 URI was used

### DIFF
--- a/src/loader/File.js
+++ b/src/loader/File.js
@@ -240,7 +240,7 @@ var File = new Class({
          * @type {boolean}
          * @since 3.80.0
          */
-        this.base64 = (url.indexOf('data:') === 0);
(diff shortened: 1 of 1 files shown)


photonstorm:
👍

PR #6770, opened 2024-03-15
bug: Add missing type definition in SpriteConfig

### LINKED ISSUE #6752: Missing type define of useSpriteSheet parameter
## Version

* Phaser Version: 3.80.1

* Operating system:

* Browser:

## Description

Parameter [`useSpriteSheet`](https://github.com/phaserjs/phaser/blob/master/src/tilemaps/components/CreateFromTiles.js#L83) is part of `spriteConfig`, which is [Phaser.Types.GameObjects.Sprite.SpriteConfig](https://github.com/phaserjs/phaser/blob/master/src/tilemaps/components/CreateFromTiles.js#L22) type.
However, type define [Phaser.Types.GameObjects.Sprite.SpriteConfig](https://github.com/phaserjs/phaser/blob/master/src/gameobjects/sprite/typedefs/SpriteConfig.js) does not include `useSpriteSheet`.

mischulz14:

* Fixes a bug

Describe the changes below:

This PR adds the missing type definition for the boolean `useSpriteSheet` property in the `SpriteConfig.js` file, that is later used conditionally in a function in the `CreateFromTiles.js` file.

Closes #6752

### DIFF
--- a/src/gameobjects/sprite/typedefs/SpriteConfig.js
+++ b/src/gameobjects/sprite/typedefs/SpriteConfig.js
@@ -6,4 +6,5 @@
  * @property {(string|Phaser.Textures.Texture)} [key] - The key, or instance of the Texture this Game Object will use to render with, as stored in the Texture Manager.
  * @property {(string|number)} [frame] - An optional frame from the Texture this Game Object is rendering with.
  * @property {(string|Phaser.Animations.Animation|Phaser.Types.Animations.PlayAnimationConfig)} [anims] - The string-based key of the animation to play, or an Animation instance, or a `PlayAnimationConfig` object.
+ * @property {(boolean)} [useSpriteSheet] - If set to `true` it uses a tileset loaded as a sprite sheet (not an image), and sets the Sprite key and frame to match the sprite texture and tile index.
  */


rexrainbow:
Parameter `useSpriteSheet` is for CreateFromTiles, not for creating sprite. 
Add a new file for CreateFromTiles dedicated instead of putting `useSpriteSheet` into [SpriteConfig](https://github.com/phaserjs/phaser/blob/master/src/gameobjects/sprite/typedefs/SpriteConfig.js) would be better, imo.

mischulz14:
> Parameter `useSpriteSheet` is for CreateFromTiles, not for creating sprite. 
> Add a new file for CreateFromTiles dedicated instead of putting `useSpriteSheet` into [SpriteConfig](https://github.com/phaserjs/phaser/blob/master/src/gameobjects/sprite/typedefs/SpriteConfig.js) would be better, imo. 

Ok, thanks for the insight, I'll refactor my code! (:

mischulz14:
@rexrainbow Upon reviewing the code again, I don't think it would make sense to create a dedictated file. 

This is because the config object that gets checked here: 

```
// line 83
 if (config.hasOwnProperty('useSpriteSheet')) // this config is a deep copy of the spriteConfig passed to the function
            {
                config.key = tile.tileset.image;
                config.frame = tile.index - 1;
            }
 ```
 
 is the spritesheet config created in the `SpriteConfig.js` file. That leads me to believe the `useSpriteSheet` property has to come from the spriteConfig declared in the upper scope of the file. 
 
 ```
 // line 22
 * @param {Phaser.Types.GameObjects.Sprite.SpriteConfig} spriteConfig - The config object to pass into the Sprite creator (i.e. scene.make.sprite).
 ```
 
HOWEVER, if you meant to introduce a completely new parameter to the function called `useSpriteSheet`, that would change things. Any feedback is appreciated (:

rexrainbow:
When user uses config(`SpriteConfig`) for creating sprite game object, then user  read the SpriteConfig source file, will user know the purpose or usage of `useSpriteSheet`  parameter if user does not survey it deeply?
User might find that it is for tileset, and it can be ignored for creating sprite game object at last. It will introduce more effort to understand the usage of SpriteConfig, imo.

photonstorm:

PR #6785, opened 2024-04-08
Throw an error for invalid BitmapText key

samme:
If you create a BitmapText object with a bad key you see something like this in the console:

> [Warning] Invalid BitmapText key: font
> [Error] TypeError: undefined is not an object (evaluating 'entry.data')

Since the error is inevitable, I've replaced the console warning so there's now a single error message and trace:

> [Error] Invalid BitmapText key: font


photonstorm:
👍

PR #6793, opened 2024-04-12
Add Phaser.Loader.LoaderPlugin#removePack

### LINKED ISSUE #6791: Unload asset pack
The loader plugin can currently load all assets from an asset-pack via `scene.load.pack(...)`, but it lacks the functionality for unloading an asset-pack.

Since the `pack` function must already have the code to read and iterate through assets in an asset-pack, maybe something like `scele.load.unloadAssetPack(...)` can be easy to implement, as it would just have to call `destroy` instead of `load` for each asset in the asset-pack.

Motivation:
Building multi-scene games for mobile with lots of assets can be difficult to do without this, as it requires the developer to track every asset it loads and manually and individually unload it when needed. This is almost like managing an asset-pack manually and it kind of defeats the purpose of the builtin asset-pack.
Pretty much all game engines support something along the lines of `unload unused assets` or `unload asset bundle` or `unload scene (which also unloads the assets)`.

samme:
This PR

* Adds a new feature (closes #6791)

> Remove the resources listed in an Asset Pack.
> 
> This removes Animations from the Animation Manager, Textures from the Texture Manager, and all other assets from their respective caches. It doesn't remove the Pack itself from the JSON cache, if it exists there. If the Pack includes another Pack, its resources will be removed too.

### DIFF
--- a/src/loader/LoaderPlugin.js
+++ b/src/loader/LoaderPlugin.js
@@ -12,6 +12,7 @@ var Events = require('./events');
 var FileTypesManager = require('./FileTypesManager');
 var GetFastValue = require('../utils/object/GetFastValue');
 var GetValue = require('../utils/object/GetValue');
+var IsPlainObject = require('../utils/object/IsPlainObject');
 var PluginCache = require('../plugins/PluginCache');
 var SceneEvents = require('../scene/events');
 var XHRSettings = require('./XHRSettings');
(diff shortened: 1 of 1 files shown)


tibimunteanu:
What is different for Spine assets? Are they managed by the Spine runtime?

samme:
It's `add.spine()` that adds to the "spineTextures" cache and the Texture Manager, and the Spine plugins prefix texture keys differently. So maybe the plugins could handle this better.

samme:
Now removes all Spine assets.

tibimunteanu:
Nice! When can we expect this to be merged?

zekeatchan:
Hi @samme. Thanks for submitting this pull request. We have implemented this and merged it with the `master` branch. It will be part of the next release.

PR #6802, opened 2024-04-25
Add Phaser.Time.Timeline#timeScale

samme:
This PR

* Adds a new feature

> The Timeline's delta time scale.
>
> Values higher than 1 increase the speed of time, while values smaller than 1 decrease it.
> A value of 0 freezes time and is effectively equivalent to pausing the Timeline.

### DIFF
--- a/src/time/Timeline.js
+++ b/src/time/Timeline.js
@@ -134,6 +134,19 @@ var Timeline = new Class({
         this.elapsed = 0;
 
         /**
+         * The Timeline's delta time scale.
+         *
+         * Values higher than 1 increase the speed of time, while values smaller than 1 decrease it.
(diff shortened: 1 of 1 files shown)


photonstorm:
Thanks for the PR. I don't mind merging it, but it won't impact any tweens that the timeline created (i.e. those will carry on playing at the time scale of the Tween Manager, not the Timeline). Does this matter do you reckon?

samme:
I think it could still be useful without that, but it definitely deserves mentioning. I will add to the docs description.

zekeatchan:
Hi @samme. Thanks for submitting this pull request. We have implemented this and merged it with the master branch. It will be part of the next release.

PR #6809, opened 2024-05-03
Add BaseSoundManager#isPlaying()

samme:
This PR

* Adds a new feature

> When a key is given, returns true if any sound with that key is playing.
> 
> When no key is given, returns true if any sound is playing.


photonstorm:
Nice 👍

PR #6815, opened 2024-05-09
JSDoc and type for parent property

lgtome:

This PR

* Updates JSDoc
* Fixes a bug

Describe the changes below:

I was using `typescript` and I found a mismatching property with `JSDoc` and `typescript`.
image
Here is the screenshot with `JSDoc` and an error from `typescript`.
I added `null` type for JSDoc.

Thank you!


photonstorm:
Thanks for the PR - could you please remove the types def from it, we only need the JSDoc change itself.

photonstorm:
👍

PR #6816, opened 2024-05-14
Add PR to fix null pointer exception.

jacklehamster:

* Fixes a bug

Describe the changes below:

Encountered the following null pointer exception:
```
Uncaught TypeError: Cannot set properties of undefined (setting 'controller')
    at FX.add (index.js:20011:39)
    at FX.addGlow (index.js:20023:27)
    at index.js:5204:132
```
This occurred only on Windows (VM on UTM, not sure if that made a difference) when trying to use special FX. This didn't happen on Mac.

The bug happened on all specials effects applied, on text or images (postFX.addGlow).
See code: https://github.com/jacklehamster/phaser-game/blob/main/src/hischool/game.ts#L2372


zekeatchan:
Hi @jacklehamster. Thanks for your PR, tested it and seems to be working fine. It's been merged with `master` branch

PR #6818, opened 2024-05-15
Set gameSize equal to baseSize, for EXPAND scale mode

rexrainbow:

* Fixes a bug

Describe the changes below:

gameSize is used for world bounds, cameras, set gameSize equal to baseSize, so that camera can render on whole game window.

### DIFF
--- a/src/scale/ScaleManager.js
+++ b/src/scale/ScaleManager.js
@@ -1084,6 +1084,9 @@ var ScaleManager = new Class({
         }
         else if (this.scaleMode === CONST.SCALE_MODE.EXPAND)
         {
+            var baseWidth = this.game.config.width;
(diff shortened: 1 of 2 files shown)


philipgriffin:
Should [scale/index.js](https://github.com/phaserjs/phaser/blob/master/src/scale/index.js) be updated too? (as it is not referenced in the docs due to not being added here)

rexrainbow:
> Should [scale/index.js](https://github.com/phaserjs/phaser/blob/master/src/scale/index.js) be updated too? (as it is not referenced in the docs due to not being added here)

Thanks, I had exposed `EXPAND` in index.js file in latest commit.

photonstorm:
👍

PR #6824, opened 2024-05-25
Fixed HexagonalTileToWorldXY.js

ukamp:
This PR
* Fixes a bug

Describe the changes below:
HexagonalTileToWorldXY is now retuning the correct CenterXY location for a tile. Only fixed and tested it for staggerAxis=Y.
For staggerAxis=X  it is still incorrect.


photonstorm:
Thanks for originally opening this PR. I'm closing it because we're no longer updating v3 and preparing v4 for launch. If you think it's still a valid addition to v4, then you're welcome to resubmit against that once it is live.

PR #6826, opened 2024-05-25
Hexagonal get tile corners

ukamp:
This PR

* Fixes a bug

Describe the changes below:
- Declaration of tempVec was missing
- Added rotation for staggerAxis = x

### DIFF
--- a/src/tilemaps/components/HexagonalGetTileCorners.js
+++ b/src/tilemaps/components/HexagonalGetTileCorners.js
@@ -35,17 +35,18 @@ var HexagonalGetTileCorners = function (tileX, tileY, camera, layer)
     }
 
     //  Sets the center of the tile into tempVec
-    var tempVec = new Phaser.Math.Vector2;
+    var tempVec = new Vector2();
     var center = HexagonalTileToWorldXY(tileX, tileY, tempVec, camera, layer);
 
     var corners = [];
 
+    // We assume a pointy-top orientation (staggerAxis == 'y')
     var b0 = 0.5773502691896257; // Math.sqrt(3) / 3
     var rotation = 0;
     var hexWidth;
     var hexHeight;
 
-    if (layer.staggerAxis === 'y')
+    if (layer.staggerAxis == 'y')
     {
         hexWidth = b0 * tileWidth;
         hexHeight = tileHeight / 2;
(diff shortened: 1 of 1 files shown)


src/tilemaps/components/HexagonalGetTileCorners.js
```diff
@@ -54,13 +54,15 @@ var HexagonalGetTileCorners = function (tileX, tileY, camera, layer)
...
+        rotation = 30 * Math.PI /180;
     }
 
     for (var i = 0; i < 6; i++)
     {
         var angle = 2 * Math.PI * (0.5 - i) / 6;
+        angle += rotation;
 
-        corners.push(new Vector2(center.x + (hexWidth * Math.cos(angle)), center.y + (hexHeight * Math.sin(angle))));
+        corners.push(new Phaser.Math.Vector2(center.x + (hexWidth * Math.cos(angle)), center.y + (hexHeight * Math.sin(angle))));
```
photonstorm: Please don't do this. Require the Vector2 and use it as per the original code. You cannot guarantee Phaser will exist in the global scope.

src/tilemaps/components/HexagonalGetTileCorners.js
```diff
@@ -35,13 +35,13 @@ var HexagonalGetTileCorners = function (tileX, tileY, camera, layer)
     }
 
     //  Sets the center of the tile into tempVec
+    var tempVec = new Phaser.Math.Vector2;
     var center = HexagonalTileToWorldXY(tileX, tileY, tempVec, camera, layer);
 
     var corners = [];
 
-    //  Hard-coded orientation values for Pointy-Top Hexagons only
```
photonstorm, line 42: Why remove a useful comment?

src/tilemaps/components/HexagonalGetTileCorners.js
```diff
@@ -35,13 +35,13 @@ var HexagonalGetTileCorners = function (tileX, tileY, camera, layer)
     }
 
     //  Sets the center of the tile into tempVec
+    var tempVec = new Phaser.Math.Vector2;
```
photonstorm: This needs to be a Vector2 and required.

src/tilemaps/components/HexagonalGetTileCorners.js
ukamp: Good Point

src/tilemaps/components/HexagonalGetTileCorners.js
ukamp: The comment is contradictory, the value is not only used for Pointy-Top Hexagons, but an aid to calculating the long diagonals. Better comment: We assume a pointy-top orientation.

photonstorm:
@greptileai

photonstorm:
Tidying up stale PRs.

PR #6846, opened 2024-06-26
Add friction for direct-control bodies

### LINKED ISSUE #6792: `Friction` property does not work with `.setDirectionControl()` method
## Version

- Phaser Version: 3.80.1

## Description

Does the `friction` property of `Phaser.Physics.Arcade.Body` works in conjunction with the `.setDirectControl()` method?
Here is a link to a good use case for platforms from the documentation page. If I set the `friction` property through `.setFriction(1, 1)` for each platform, I expect that the player being on the platform will move along with it. But it slips through. I've already implemented this mechanics with platforms on Phaser `3.55`, calculating the velocity of the platform depending on the coordinates manually. It still works on the phaser `3.8`, but I would like to use `.setDirectControl()` method, it would help remove a lot of unnecessary calculations.

## Example Test Code

Examples are available from the documentation [page](https://newdocs.phaser.io/docs/3.70.0/focus/Phaser.Physics.Arcade.Sprite-setDirectControl)
[PlatformsExample](https://labs.phaser.io/edit.html?src=src/physics\arcade\direct%20control%20platforms.js)

additionally set
```
platform1.setFriction(1, 1);
platform2.setFriction(1, 1);
platform3.setFriction(1, 1);
```

samme:
This PR

* Adds a new feature

Fixes #6792

### DIFF
--- a/src/physics/arcade/ProcessX.js
+++ b/src/physics/arcade/ProcessX.js
@@ -357,7 +357,9 @@ var RunImmovableBody1 = function (blockedState)
     //  This is special case code that handles things like vertically moving platforms you can ride
     if (body1.moves)
     {
-        body2.y += (body1.y - body1.prev.y) * body1.friction.y;
+        var body1Distance = body1.directControl ? (body1.y - body1.autoFrame.y) : (body1.y - body1.prev.y);
+
+        body2.y += body1Distance * body1.friction.y;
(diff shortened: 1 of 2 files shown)


zekeatchan:
Thanks @samme. Tested and seems to work fine👍

PR #6867, opened 2024-07-18
Add TimeStep#pauseDuration, pass in Phaser.Core.Events#RESUME

samme:
This PR

* Adds a new feature

Some people want to use the pause interval, and you would have to calculate it yourself, even though TimeStep knows what it is. So I added. 

```js
this.game.events.on('resume', (pauseDuration) => {
  console.log('%s ms have passed', pauseDuration);
});
```

PR #6870, opened 2024-07-21
Pass pauseDuration 0 in `Phaser.Core.Events#RESUME` after manual game resume

samme:
This PR

* Fixes a bug

After #6867, the other `RESUME` event (from calling `Phaser.Game#pause()`) also needed a `pauseDuration` argument. In this case it's `0` because the TimeStep wasn't actually paused.


photonstorm:
👍

Issue #6877, opened 2024-07-28
ParticleEmitter `speedX` and `speedY` Not Updating After Initial Creation


monteiz:
## Version

* Phaser Version: 3.85.0-beta.2

After creating a `ParticleEmitter`, subsequent changes to the emitter's `speedX` and `speedY` properties are not taken into account. Below is a minimal example demonstrating this issue. When the emitter is triggered on a pointer down event, the speed is supposed to be set to zero, but the particles still maintain their initial speed.

## Example Test Code

To be pasted [in this example](https://labs.phaser.io/edit.html?src=src\game%20objects\particle%20emitter\explode%20emitter.js):

```javascript
class Example extends Phaser.Scene
{
    preload ()
    {
        this.load.atlas('flares', 'assets/particles/flares.png', 'assets/particles/flares.json');
    }

    create ()
    {
        const emitter = this.add.particles(400, 250, 'flares', {
            frame: [ 'red', 'yellow', 'green' ],
            lifespan: 4000,
            speed: { min: 150, max: 250 },
            scale: { start: 0.8, end: 0 },
            gravityY: 150,
            blendMode: 'ADD',
            emitting: false
        });

        this.input.on('pointerdown', pointer => {

            emitter.explode(16);

            // Attempt to change particle speed to zero
            emitter.speedX = 0;
            emitter.speedY = 0;
            emitter.setParticleSpeed(0,0);

        });

        this.add.text(10, 10, 'Click to explode emit particles');
    }
}

const config = {
    type: Phaser.AUTO,
    width: 800,
    height: 600,
    backgroundColor: '#000',
    parent: 'phaser-example',
    scene: Example
};

const game = new Phaser.Game(config);
```

### Expected Behavior:
When the emitter is triggered by the pointer down event, the particle speed should be updated to the new values set (speedX and speedY to 0), resulting in particles emitting with no horizontal or vertical speed.

### Actual Behavior:
Despite setting speedX and speedY to 0, the particles still emit with their initial speed values.

photonstorm:
In the example you emit the particles and _then_ set the emitter speed to zero. Speed is an `onEmit` property, not an `onUpdate` one, so the particles already launched will not be modified by changing the speed.

PR #6882, opened 2024-08-03
Fixed issue #6877

liamp55:

* Fixes a bug
There was an issue with the particle emitter function in that it was not updating the X and Y coordinates.

Describe the changes below:

By changing ParticleEmitterOps.js I changed the code so that X and Y values are updated, no matter what the original speed of each was


photonstorm:
The only changed file in this commit is the documentation. Did you forget some other files?

Also, speedX/Y were _intended_ to be `onEmit` only. It wasn't an oversight or mistake.

PR #6896, opened 2024-08-23
Docs: clarify `moveAbove()`, `moveBelow()`

samme:
This PR

* Updates the Documentation

It's not well known that these methods move the target only if it's not already above/below the base.


photonstorm:
👍

Issue #6915, opened 2024-10-05
Text truncated when setLetterSpacing is used


monteiz:
## Version
* Phaser Version: 3.85.2

* Operating system: MacOS

* Browser: Chrome and Safari show the same issue

## Description

The "Y" of "PLAY" is truncated when setLetterSpacing is used.

## Example Test Code

```javascript
class Example extends Phaser.Scene
{
    create ()
    {
        this.add.text(20, 20, 'PLAY',{
				fontSize: "10rem",
				fontFamily: "Roboto Condensed, sans-serif-condensed",
				fontStyle: "bold",
			}).setColor('#ffffff');

        this.add.text(20, 200, 'PLAY',{
				fontSize: "10rem",
				fontFamily: "Roboto Condensed, sans-serif-condensed",
				fontStyle: "bold",
			}).setColor('#ffffff').setLetterSpacing(70);
    }
}

const config = {
    type: Phaser.AUTO,
    width: 800,
    height: 600,
    parent: 'phaser-example',
    scene: Example
};

const game = new Phaser.Game(config);
```

rexrainbow:
This rendering issue seems related with fontFamily

Default font family won't have truncating issue.
```js
        this.add.text(20, 200, 'PLAY', {
            fontSize: "10rem",
            // fontFamily: "Roboto Condensed, sans-serif-condensed",
            fontStyle: "bold",
        }).setColor('#ffffff').setLetterSpacing(70);
```

photonstorm:
Thank you for submitting this issue. We have fixed this and the fix has been pushed to the `master` branch. It will be part of the next release. If you get time to build and test it for yourself we would appreciate that.

monteiz:
Thank you 👋

PR #6916, opened 2024-10-07
Fix rendering bug when using letterSpacing

rexrainbow:

* Fixes a bug

Describe the changes below:

Measure text width by each character one by one when using letterSpacing (#6915)


photonstorm:
👍

Issue #6860, opened 2024-07-10
BitmapText not wrapping lines in some cases


bagyoni:
## Version

* Phaser Version: 3.80.1

* Operating system: Debian

* Browser: Chromium, Firefox

## Description

If a line of `BitmapText.text` is (1) long enough to get wrapped, and (2) it ends with a space, then the next non-empty line will never be wrapped, no matter how long it is.
In the example, the first and last lines are wrapped correctly, but the middle one is not.

## Example Test Code

```js
const text = `
the end is never the end is never the end is never the end is never the \n
the end is never the end is never the end is never the end is never the\n
the end is never the end is never the end is never the end is never the`

class Example extends Phaser.Scene
{
    preload ()
    {
        this.load.bitmapFont('shortStack', 'assets/fonts/bitmap/shortStack.png', 'assets/fonts/bitmap/shortStack.xml');
    }

    create ()
    {
        this.add.bitmapText(0, 200, 'shortStack', text, 16).setMaxWidth(400);
    }
}

const config = {
    type: Phaser.WEBGL,
    parent: 'phaser-example',
    scene: Example
};

const game = new Phaser.Game(config);
```

## Additional Information

Might be related to #6717.

zekeatchan:
Hi @bagyoni. Thanks for submitting this issue. We have fixed this and pushed it to the `master` branch. It will be part of the next release. Do test it out and let us know if you encounter any issues.

bagyoni:
Thanks for the quick response @zekeatchan. The change appears to cause the first two lines to overlap. (The one in the example I sent begins with a newline, so it won't show this). It also doesn't seem to solve the line wrapping issue on my end.
~(If this is an edge case, would it be a more conservative solution to just trim() each line before calculating the bounds, as the AI seems to suggest?)~ I guess that would create issues with fonts where spaces aren't empty.

zekeatchan:
Noticed the first 2 lines overlapping as well.

Another push has been made to `master` with updated `maxWidth` calculations. 

Could you kindly test it out and provide examples if you encounter other issues? Thanks!

bagyoni:
I'm still seeing the same issues.

```js
const text = `first line overlaps with itself
second line ends with a space 
third line not wrapped as it should be`;

class Example extends Phaser.Scene {
	preload() {
		this.load.bitmapFont(
			"atari",
			"assets/fonts/bitmap/atari-smooth.png",
			"assets/fonts/bitmap/atari-smooth.xml"
[...]

PR #6920, opened 2024-10-12
Fix calculation of idx in GetBitmapTextSize

bagyoni:
* Fixes a bug

Describe the changes below:

Fixes the way `BitmapTextCharacter.idx` is calculated in `GetBitmapTextSize` that is inconsistent with the docs and the behavior in 3.80 and before. This was caused by lines 165 and 173 changing the length of the text `idx` iterates over.
Code to demonstrate the bug: https://github.com/phaserjs/phaser/issues/6860#issuecomment-2334295340 
Also removed some code that was no longer used because it always iterates on an empty array.
Checked that it doesn't re-introduce the word-wrapping bug in #6860.


photonstorm:
👍

Issue #6929, opened 2024-10-22
AnimationState listener handling


photonstorm:
Consider alternative for AnimationState.destroy handling, specifically:

```
this.animationManager.off(Events.REMOVE_ANIMATION, this.globalRemove, this);
```

re: stopping sprites/particles from iterating through the event list every time one of them is destroyed.

Issue #6482, opened 2023-04-19
Performance issue when shutting down scene on Phaser 3.60.0


kpagan:
## Version

* Phaser Version: 3.60.0

* Operating system: Windows 10

* Browser: Firefox 112.0.1 64-bit, Chrome 112.0.5615.121 64-bit

## Description

Hello,
I was developing a space shooter game with Phaser 3.55.2 and after the 3.60.0 release I started migrating to the new version.
After the code migration I tested the game and found out that when I complete a level and click the OK button to return to the main menu scene it takes almost 10 seconds to transition to the scene.
I run a performance analysis with the Chrome profile and it seems that the removeListener, destroy and preDestroy methods to take too long. In 3.55.2 this would take just milliseconds. You can check the screenshot of the Chrome profiler.
When the scenes are shutting down I remove all listeners. However, I tried to comment out the code that cleans up but it still takes around 10 seconds.

## Additional Information

![image](https://user-images.githubusercontent.com/2040312/233131825-f1ffa03f-8d16-4ba4-8071-2fbd47773f2b.png)

photonstorm:
I can't debug screenshots I'm afraid. While it's clear that _your_ game is taking a while to destroy this Scene, it's not a global change that impacts every game in 3.60, so it's going to need a lot more debugging to narrow down - ideally the game itself, so a test case can be created.

kpagan:
Thank you for your prompt reply!
The game itself is hosted on GitHub and you can view the source code on the https://github.com/kpagan/Phaser3Airfighter/tree/phaser3.60 branch.
I don't know if I am doing something extraordinary but if you like to take a look then check `src\scenes\ui\LevelCompleteModal.ts` where you click the OK button to go to the main menu scene. The scenes that are shut down are the `src\scenes\CloudScene.ts` and the `src\scenes\UiScene.ts`.

Thanks

kpagan:
Hello!
I managed to narrow down the bottleneck! 

I use particle emitters on the enemy sprites. These sprites are pooled in a GameObjects group. When the sprite was hit I would deactivate it and return it to the group. But when the sprite was spawned again pulling it from the group then I was re-creating the particle emitter.
With the change in the API of Phaser 3.60.0 to the particle emitters it seems that the old one was still around and when the scene was shutting down then all the emitters were destroyed during that time.
Now I check if the particle emitter is already instantiated so I will not create it again and the shutdown takes a lot less time, under 1 second.

[...]

PR #6930, opened 2024-10-23
Make particle animation state optional

samme:
This change creates `Phaser.GameObjects.Particles.Particle#anims` only if the particle's emitter has at least one animation.

This should reduce memory usage and make cleanup faster (#6482, #6929) for games using non-animated particles.

It is a breaking change for games adding a **first** animation to the emitter **after** some particles have already been created. That will now cause an error. But the usual approach for adding animations will work like before:

```js
this.add.particles(0, 0, 'explosion', {
  anim: ['explode']
  // ...
});
```


photonstorm:
👍

PR #6936, opened 2024-10-30
return specific type when getting scene and T is provided

kpotschi:

* Fixes a bug

Describe the changes below:
return the expected class <T> when using `scene.get<T>()`


photonstorm:
Thank you for spending time contributing to the Phaser TypeScript Defs; I really appreciate it. However, I'm afraid I cannot merge your PR. The reason is that you've edited the Phaser defs file directly. Our defs file is generated automatically as part of our build process; it works by running a custom parser across the Phaser source code, extracting the JSDoc blocks, and then building the types from those. This means that if we were to merge your PR, your changes would be instantly lost the next time the defs are generated.

To fix the defs permanently, you need to edit the underlying JSDocs in the source code, causing the defs error. Sometimes, the error results from a limitation in the TS parser, but mostly because of incorrectly typed JSDocs. Please look at the command `npm run help` to see the options available to you for building and running our TS parser for yourself so you can edit the JSDoc blocks, run the parser, and see if the end result is resolved or not.

kpotschi:
thanks rich, I'll see what I can do!

PR #6957, opened 2024-11-19
Fix Arcade Physics Body destroy

samme:
This PR

* Fixes a bug (v4.0.0-b.1)

There was an error destroying an Arcade Physics body:

> TypeError: this.world.pendingDestroy.set is not a function


photonstorm:
👍

PR #6958, opened 2024-11-20
Fix TypeError when using scene injection map

samme:
This PR

* Fixes a bug (v4.0.0-beta.2)

When [a scene injection map](https://phaser.io/sandbox/zyjiE6in) is used, `scene.game` can be undefined and cause an error. I changed to `scene.sys.game`.


photonstorm:
👍

PR #6961, opened 2024-11-23
Use an optional chaining operator when calling preUpdate

justinleewells:

* Fixes a bug

Describe the changes below:

This simply ensures gameObject.preUpdate exists before attempting to `call` it. Doing so allows developers to forego the `preUpdate` method when creating custom GameObjects.

If we would prefer to avoid using an optional chaining operator, I can submit another pull request that instead checks for the existence of the method in the `if (gameObject.active)` conditional.

### DIFF
--- a/src/gameobjects/UpdateList.js
+++ b/src/gameobjects/UpdateList.js
@@ -153,7 +153,7 @@ var UpdateList = new Class({
 
             if (gameObject.active)
             {
-                gameObject.preUpdate.call(gameObject, time, delta);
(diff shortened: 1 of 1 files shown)


samme:
But game objects without `preUpdate()` shouldn't be on the Update List, since its only job is to call `preUpdate()`. 🙂 

`add.existing()` and `addToUpdateList()` do check for `preUpdate()`.

justinleewells:
@samme 
simpsons-bushes

PR #6966, opened 2024-11-26
Fix closest() and furthest()

samme:
This PR

* Fixes a bug (v4.0.0-beta.2)

Arcade Physics `closest()` and `furthest()` could error:

> [Error] TypeError: null is not an object (evaluating 'closest.center')

04694e3


photonstorm:
👍

Issue #6965, opened 2024-11-26
Failed to execute 'setTransform' on 'CanvasRenderingContext2D': 6 arguments required, but only 1 present.


rafa-fie:
## Version

* Phaser Version: 3.70 (but tested also with 3.87)

* Operating system: Windows

* Browser: CEF (Chromium Embedded Framework): CEF 52 (default browser from Sketchup 2017)

## Description

When open that browser I have an error:
Uncaught TypeError: Failed to execute 'setTransform' on 'CanvasRenderingContext2D': 6 arguments required, but only 1 present.

this problem do not invoke in newest version of Sketchup Browser (like CEF 88 and above)

![image](https://github.com/user-attachments/assets/64596e97-187d-440d-a74a-510e4b32a1e0)

## Example Test Code

in the config file I am using Phaser.CANVAS, because that renderer is better for my case (using phaser to visualize pdf so thin lines are necessary)

when Phaser.Auto then it is working but that renderer isn't good in my case

## Additional Information

I have solved that issue by changing line
`return t.setTransform(this),t}`
into
`return t.setTransform(this.a, this.b, this.c, this.d, this.e, this.f),t}`

not sure if it is safe solution, but it's working ;P
maybe in future releases you can add something like this.

ps. very very great library!!!

![image](https://github.com/user-attachments/assets/dd498069-3bbd-4676-95c8-5c01cdfd6478)
![image](https://github.com/user-attachments/assets/4084e667-f32f-44b2-930a-e2733c1b1e1f)

photonstorm:
Hiya - how old is this device/set-up you're testing? Because `setTransform` has allowed you to pass a matrix object to it for _years_ now! See https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/setTransform#browser_compatibility

rafa-fie:
yeees, that is pretty old browser, from 2017 probably.
Using your link... 
Chrome (not sure if Chrominium has same numbers) support that from version 68. Unfortunately Sketchup started using Chrominium > 68 after 2021 ;/

Full situation:
We are developing plugin for SketchUp (app for architects) and using phaser to have some kind of PDF viewer/editor.
SketchUp has own browser based on CEF.

![obraz](https://github.com/user-attachments/assets/b81b2f77-7057-4640-a9ee-438d0ed2c158)

Of course we wish that everybody use the newest version of SketchUp, but reality is not like that :P
[...]

PR #6970, opened 2024-11-29
add backward compatibility for setTransform function #6965

rafa-fie:

* Fixes a bug

Describe the changes below:

add backward compatibility for setTransform function #6965


photonstorm:
👍

PR #6974, opened 2024-12-04
Loading retry improvements

pavle-goloskokovic:
This PR improves an existing feature:

- Implemented retries with exponential backoff which is a technique that retries loading with an exponentially increasing delay time, up to a maximum retry count
- Increased default `maxRetries` value to `5`, which together with exponential backoff increases possible loading error recovery time to ~3 seconds compared to previous less then 100ms
- Added `FILE_LOAD_RETRY` event to Loader Plugin to enable listening to retry occurrences
- `FILE_LOAD_ERROR` now also emits Event that resulted form an error as an argument to provide more details to user regarding error cause
- Moved `maxRetries` setting from `FileConfig` to `XHRSettingsObject` provided to any Loader Plugin method call, to make it possible to override value per file, since there is no way to define properties on `fileConfig` directly
- Few other fixes along the way

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -53,8 +53,6 @@ module.exports = [
                     });
                 }
             }
-        ],
-
-        devtool: 'source-map'
+        ]
     }
 ];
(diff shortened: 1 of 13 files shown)


photonstorm:
Thanks for originally opening this PR. I'm going to close it because we're moving to Phaser v4. If you think this feature is still valid for v4 you're welcome to resubmit it once v4 is live.

PR #6976, opened 2024-12-04
Fix Static Physics Group refresh

samme:
This PR

* Fixes a bug (v4.0.0)

`refresh()` was not affecting any bodies.


photonstorm:
👍

PR #6978, opened 2024-12-09
Prevent matterjs stalls when returning to inactive tabs

### LINKED ISSUE #6977: Tab stalls when using matterjs and moving away from the tab for a while, probably Engine.update related
## Version

* Phaser Version: v3.85.0

* Operating system: Windows 11 Pro 22631.4460

* Browser: Chrome

## Description

I noticed that my phaser game was crashing when I went away from it for a while and came back to the tab.

At https://ubershmekel.github.io/bell-curve-balls/ there are a lot of balls, so the simulation step is a bit heavier than usual.

Phaser-Matter-away-problem

## Example Test Code

* https://phaser.io/examples/v3.85.0/physics/matterjs/view/pachinko?display_mode=full_screen works more slowly. You have to create a few balls then be away from the tab for 20 minutes before you see the delays. Code at https://phaser.io/examples/v3.85.0/physics/matterjs/view/pachinko
* https://ubershmekel.github.io/bell-curve-balls/ works pretty quickly, just click it once to let the balls roll, then go to another tab and wait 1-2 minutes and when you come back to this tab it will be multiple seconds until you can play again. Code at https://github.com/ubershmekel/bell-curve-balls

## Additional Information

Options I see:

* Engine.update should be called while in the background
* Do not call Engine.update to catch up when the tab becomes active, just resume the game from where it was when the tab became inactive.
* Create some "simulation is catching up spinner UI" for when the tab becomes active again.

The relevant code is at:

* This is the function that gets called multiple times during the performance freeze: https://github.com/phaserjs/phaser/blob/97e3e67ac679863e30d4225d2b29a49a6f0de3e9/src/physics/matter-js/lib/core/Engine.js#L84
* This is the function that takes a ton of time to complete because of all the `Engine.update` calls: https://github.com/phaserjs/phaser/blob/97e3e67ac679863e30d4225d2b29a49a6f0de3e9/src/physics/matter-js/World.js#L1268

ubershmekel:

* Fixes a bug

Describe the changes below:

Fixes https://github.com/phaserjs/phaser/issues/6977

A more complete PR would include adding this to the type `Phaser.Types.Physics.Matter.MatterRunnerConfig`

This change limits the amount of time matterjs will be simulating after a tab is inactive for a long time. The number 100ms seems a reasonable maximum amount of time to wait for the game loop to resume.

Another issue with this PR is I don't understand why does src/physics/matter-js/World.js not use `Runner.create` as defined in src/physics/matter-js/lib/core/Runner.js but I'd rather someone that knows what's going on give an opinion on that.

### DIFF
--- a/src/physics/matter-js/World.js
+++ b/src/physics/matter-js/World.js
@@ -179,7 +179,8 @@ var World = new Class({
             isFixed: GetFastValue(runnerConfig, 'isFixed', false),
             delta: delta,
             deltaMin: deltaMin,
-            deltaMax: deltaMax
+            deltaMax: deltaMax,
+            maxFrameTime: 100
         };
 
         /**


photonstorm:
Thank you for spending time opening this, however we've merged in a different PR that will fix this issue in a slightly better way.

PR #6991, opened 2024-12-20
Fail gracefully when a texture isn't created in `addBase64()`

samme:
This PR

* Fixes a bug

In `addBase64()`, `onload()` didn't handle a null result from `TextureManager#create()`, giving an error:

> TypeError: null is not an object (evaluating 'texture.source')

I changed it so `onload()` exits early without an error.


photonstorm:
👍

samme:
I guess this could go in v3.88 too.

photonstorm:
Yeah I already added it there :)

PR #6993, opened 2024-12-24
Applying config background color immediately

pavle-goloskokovic:
This PR makes an improvement/fixes a bug:

Currently background color is rendered on first render call and not immediately upon renderer creation.

This causes a brief but noticeable "flash" of black canvas which lasts from game creation to first render call.

This PR introduces logic to apply background color on renderer initialization, both for Canvas and WebGL.

It also fixes a bug for Canvas renderer where `transparent` setting was not being copied from game to renderer config, but was being used.

### DIFF
--- a/src/renderer/canvas/CanvasRenderer.js
+++ b/src/renderer/canvas/CanvasRenderer.js
@@ -14,6 +14,7 @@ var Events = require('../events');
 var GetBlendModes = require('./utils/GetBlendModes');
 var ScaleEvents = require('../../scale/events');
 var TextureEvents = require('../../textures/events');
+var GameEvents = require('../../core/events');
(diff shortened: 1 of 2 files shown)


photonstorm:
You're the first person to notice this in 10 years 😅 I don't mind merging this, but I'd question what's going on in the game code to make it take > 60ms between the addition of the canvas to the dom, to the first render step.

pavle-goloskokovic:
Took me 10 years to notice 😅

I have a boot scene which applies some global settings and starts preload scene from its create method. I guess it takes 3-4 ticks to get from canvas creation to preload scene rendering, taking ~50-65ms of black canvas, which I noticed while working on preloading animation :)

PR #6994, opened 2024-12-24
Mouse events for unlocking Web Audio

pavle-goloskokovic:
This PR makes an improvement:

Added `mousedown` and `mouseup` event listeners for unlocking Web Audio. Both events occur before existing `click` event, allowing for earlier audio unlocking on devices that use mouse.

### DIFF
--- a/src/sound/webaudio/WebAudioSoundManager.js
+++ b/src/sound/webaudio/WebAudioSoundManager.js
@@ -81,13 +81,16 @@ var WebAudioSoundManager = new Class({
 
         BaseSoundManager.call(this, game);
 
-        if (this.locked && game.isBooted)
+        if (this.locked)
         {
-            this.unlock();
-        }
-        else
-        {
-            game.events.once(GameEvents.BOOT, this.unlock, this);
+            if (game.isBooted)
+            {
+                this.unlock();
+            }
+            else
+            {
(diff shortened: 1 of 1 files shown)


photonstorm:
If we add those I don't see the need to keep the `click` event at all. Can anyone think of one?

pavle-goloskokovic:
> If we add those I don't see the need to keep the `click` event at all. Can anyone think of one?

Technically we don't need `mouseup` and `touchend` if `mousedown` and `touchstart` work.

But it depends on browser vendors which user gestures they perceive as valid for resuming audio context. And we don't have control over it or insight into specific browser implementations.

`click` also occurs on mobile, where not all `touchstart` and `touchend` are valid user gestures (like when part of a swipe or a scroll).

I would say that we are safe to remove `click` listener, but would keep it in to cover some potential edge cases, and have audio unlocked as soon as possible.

photonstorm:
`mouseup` and `touchend` are there because the event may have started outside of the canvas but completes while over it.

pavle-goloskokovic:
Makes sense. I removed `click` handlers and updated `locked` condition to only depend on audio context state. Also fixed condition for invoking `unlock` method.

photonstorm:
👍

PR #7009, opened 2025-01-15
Prevent hanging timer events with repeat

### LINKED ISSUE #7004: A TimerEvent with a delay of 0 and repeat of -1 causes an infinite loop
## Version

* Phaser Version: 3.86.0

* Operating system: Windows

* Browser: Chrome

## Description

Setting the delay property to 0 and the repeat property to -1 in an TimerEvent causes an infinite loop.

## Example Test Code

Note - the following code will cause your browser tab to lock up.
```
export default class TimerEventDelayIssue extends Phaser.Scene {
  create() {
    this.time.addEvent({
      delay: 0,
      callback: () => console.log('timer event callback'),
      repeat: -1
    });
  }
}
```

## Additional Information

Prior to Phaser ~3.60, using a delay value of 0 seemed to be a way to "allow this to run every frame." Post Phaser 3.60, there seems to be logic to make sure the callback gets called (on average) once every `delay` value. So the callback might get called multiple times per frame. This happens [here](https://github.com/phaserjs/phaser/blob/v3.86.0/src/time/Clock.js#L400) in the Clock.js file. There is some logic in the [TimerEvent `reset` function](https://github.com/phaserjs/phaser/blob/v3.86.0/src/time/TimerEvent.js#L193) to check for delay of 0 and loop/repeats, but it doesn't check for `repeat === -1`. I think it may make more sense for the check to be in `Clock.js`, in the `update` loop - something like this:
```
// Very short delay
if (remainder >= event.delay && event.delay > 0)
{
  while ((remainder >= event.delay) && (event.repeatCount > 0))
  {
    if (event.callback)
    {
      event.callback.apply(event.callbackScope, event.args);
    }

    remainder -= event.delay;
    event.repeatCount--;
  }
}
```
This allows you to have repeating frame based events just like you could in Phaser 3.55.2, while also protecting against delays that are 0 to prevent infinite loops. The check in `TimerEvent.js` could then be removed since it would be a valid config again.

samme:
This PR

* Fixes a bug

I improved the check in `TimerEvent#reset()` and added a check in `Clock#addEvent()`.

Fixes #7004.

### DIFF
--- a/src/time/Clock.js
+++ b/src/time/Clock.js
@@ -197,6 +197,11 @@ var Clock = new Class({
             event.elapsed = event.startAt;
             event.hasDispatched = false;
             event.repeatCount = (event.repeat === -1 || event.loop) ? 999999999999 : event.repeat;
+
+            if (event.delay <= 0 && event.repeatCount > 0)
+            {
+                throw new Error('TimerEvent infinite loop created via zero delay');
+            }
         }
         else
         {
--- a/src/time/TimerEvent.js
+++ b/src/time/TimerEvent.js
@@ -188,7 +188,7 @@ var TimerEvent = new Class({
         this.hasDispatched = false;
         this.repeatCount = (this.repeat === -1 || this.loop) ? 999999999999 : this.repeat;
 
-        if (this.delay === 0 && (this.repeat > 0 || this.loop))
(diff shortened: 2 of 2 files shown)


zekeatchan:
Thanks @samme for the PR. It has been merged with the `master` branch.

PR #7012, opened 2025-01-21
Update TilemapLayerWebGLRenderer.js

ospira:
This PR

* Fixes a bug(?)

Describe the changes below:

This is a draft. The default [WebGLPipeline.setTexture2D()](https://github.com/ospira/phaser/blob/hotfix-TilemapLayerWebGLRenderer/src/renderer/webgl/WebGLPipeline.js#L1982) (and most of the other WebGL pipelines that inherit from it) take only 1 arg, but is called in TilemapLayerWebGLRenderer.js with two.

This may actually be designed to be flexible if you set the pipeline to Light2D, which I believe is the only pipeline that has an overwriting implementation of [setTexture2D()](https://github.com/ospira/phaser/blob/hotfix-TilemapLayerWebGLRenderer/src/renderer/webgl/pipelines/LightPipeline.js#L263)... as I realized when I saw the same two arg call in [TileSpriteWebGLRenderer](https://github.com/ospira/phaser/blob/hotfix-TilemapLayerWebGLRenderer/src/gameobjects/tilesprite/TileSpriteWebGLRenderer.js#L41) and [TextWebGLRenderer](https://github.com/ospira/phaser/blob/hotfix-TilemapLayerWebGLRenderer/src/gameobjects/text/TextWebGLRenderer.js#L38). 

If that is the case then this PR can be ignored/closed, just note it makes it not play nice with TypeScript if you are not using/expecting Light2D as your pipeline, but that's fine as I am aware that would be out of scope for a PR of this nature.

Thanks.

### DIFF
--- a/src/tilemaps/TilemapLayerWebGLRenderer.js
+++ b/src/tilemaps/TilemapLayerWebGLRenderer.js
@@ -72,7 +72,7 @@ var TilemapLayerWebGLRenderer = function (renderer, src, camera)
 
         var texture = tileset.glTexture;
 
-        var textureUnit = pipeline.setTexture2D(texture, src);
+        var textureUnit = pipeline.setTexture2D(texture);
 
         var frameWidth = tileWidth;
         var frameHeight = tileHeight;


ospira:
Confirmed intentional to accommodate Light2D pipeline

Issue #7014, opened 2025-01-23
Layers's `setToTop()` method throws an exception: `this.getDisplayList is not a function`


leha-games:
## Version
* Phaser Version: 3.87.0
* Operating system: Windows 10

## Description

Calling setToTop() method of `Phaser.GameObjects.Layer` throws an exceptions: `Uncaught TypeError: this.getDisplayList is not a function`

## Example Test Code

```js
this.add.layer().setToTop();
```

## Additional Information

The `getDisplayList()` method appears to be missing from the `Phaser.GameObjects.Layer` class.

zekeatchan:
Hi @leha-games. Thanks for submitting this issue. @rexrainbow has submitted a PR to fix this and we've merged it to the `master` branch. It will be part of the next release. Do test it out and let us know if you encounter any issues.

PR #7015, opened 2025-01-23
Add getDisplayList method to layer game object

rexrainbow:

* Fixes a bug

Describe the changes below:

Layer game object does not extend from GameObject class, thus it does not have getDisplayList method. Copy this method from GameObject class. #7014

### DIFF
--- a/src/gameobjects/layer/Layer.js
+++ b/src/gameobjects/layer/Layer.js
@@ -919,6 +919,38 @@ var Layer = new Class({
     },
 
     /**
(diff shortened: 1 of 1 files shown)


zekeatchan:
Hi @rexrainbow. Thanks for submitting this PR. We have merged and pushed it to the `master` branch. It will be part of the next release.

rexrainbow:
The root cause is Layer game object is not extended from GameObject class, so these methods must copy from GameObject.

Issue #7016, opened 2025-01-24
Phaser 3.87 doc issue deriving on unexpected visual @ emitters under CANVAS rendering


urueda:
* Phaser Version: 3.87.0
> Rendering: Phaser.CANVAS (WEBGL w/out issues)

* Sandbox: https://phaser.io/sandbox/1hQHhXvU
> Issue happens using onUpdate w/out onEmit at emitter configuration
> Expected: emitter particles should only display at green circle (emitting spot)
> Unexpected: emitter particles displaying at red circle (out of emitting spot)

* Doc: https://docs.phaser.io/api-documentation/class/gameobjects-particles-particleemitter#custom-onemit-and-onupdate-callbacks
> You can provide **either one** or both functions. The onEmit is called at the start of the particles life and defines the value of the property on birth. The onUpdate function is called every time the Particle Emitter updates until the particle dies.

zekeatchan:
Hi @urueda. Thanks for submitting this issue. @samme has submitted a PR to fix this and we've merged it to the master branch. It will be part of the next release. Do test it out and let us know if you encounter any issues.

urueda:
> Hi [@urueda](https://github.com/urueda). Thanks for submitting this issue. [@samme](https://github.com/samme) has submitted a PR to fix this and we've merged it to the master branch. It will be part of the next release. Do test it out and let us know if you encounter any issues.

As the issue is exposed @ [Sandbox](https://phaser.io/sandbox/1hQHhXvU) ... might I suggest that the sandbox also includes the master branch version ? Probably that would benefit everyone, specially when testing is a concern. Thanks

UPDATE: tested through the **Dev Build** @ [Sandbox](https://phaser.io/sandbox/1hQHhXvU) and issue is fixed. Should also be merged into Phaser 4 as Beta versions got the issue too at the sandbox example.

PR #7017, opened 2025-01-26
Fix EmitterOp `defaultEmit()`, `current` value for custom callbacks

samme:
This PR

* Fixes a bug (2)

1. EmitterOp `defaultEmit()` always returned `undefined`, causing particle problems if you gave only an `onUpdate` callback (#7016).
2. If you configured an EmitterOp with `onEmit` or `onUpdate`, the op's `current` value would be incorrect (an object) until the first emit.


zekeatchan:
Hi @samme. Thanks for submitting this PR. We have merged and pushed it to the `master` branch. It will be part of the next release.

PR #7053, opened 2025-02-20
Fix imageCollections default Tileset values

Snoturky:
* Fixes a bug

Describe the changes below:

A recent change (a763d77) prevents the default values for Tileset's tileProperties and tileData from being set because it passes `null` instead of `undefined` for those [constructor args](https://github.com/phaserjs/phaser/blob/b73965652a715932f2f52f6bd53367edca9427c2/src/tilemaps/Tileset.js#L35).

This restores those default values, eliminating an `Uncaught TypeError: Cannot read properties of null` error [here](https://github.com/phaserjs/phaser/blob/b73965652a715932f2f52f6bd53367edca9427c2/src/tilemaps/Tileset.js#L214) when using ImageCollections.

I rolled back my phaser version from 3.88 to 3.87 (before the above changes) and the error was gone.

### DIFF
--- a/src/tilemaps/parsers/tiled/BuildTilesetIndex.js
+++ b/src/tilemaps/parsers/tiled/BuildTilesetIndex.js
@@ -35,7 +35,7 @@ var BuildTilesetIndex = function (mapData)
                 y: image.height - mapData.tileHeight
             };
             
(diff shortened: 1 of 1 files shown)


zekeatchan:
Hi @Snoturky. Thanks for submitting this PR. We have merged this and pushed it to the `master` branch. It will be part of the next release. Do test it out and let us know if you encounter any issues.

PR #7056, opened 2025-02-21
Throw an error for missing DOM Container

samme:
This PR

* Adds a new feature

Previously Phaser would silently abort `add.dom()` if `game.domContainer` was empty. Now it throws an error instead:

> No DOM Container


photonstorm:
👍

PR #7079, opened 2025-03-14
Fix missing types for Filters component

SBCGames:
* Updates the Documentation
* Fixes a bug

Small update in doc for GameObject.js. In Phaser 4.0.0 it newly uses Filter component, but was missing  @extends tag for this component in doc.

Therefore, there was missing Filters interface implementation in GameObject class in generated Typescript defs. And it was not possible to use enableFilters() in Typescript code.

(I hope this is the right way how to fix the missing type defs...)

### DIFF
--- a/src/gameobjects/GameObject.js
+++ b/src/gameobjects/GameObject.js
@@ -21,6 +21,7 @@ var SceneEvents = require('../scene/events');
  * @class GameObject
  * @memberof Phaser.GameObjects
  * @extends Phaser.Events.EventEmitter
+ * @extends Phaser.GameObjects.Components.Filters
  * @constructor
  * @since 3.0.0
  *


BenjaminDRichards (commented):
It's correct, but there should be more. GameObject also extends the RenderSteps component. Looking at e.g. `Image`, you can gather up the mixins and document them just below the `@since` tag.

BenjaminDRichards:
I'm rusty with the PR review options; I think I should have clicked "request changes". Let me know if you just want me to do it myself, or you can implement it - either way is fine!

SBCGames:
I am total PR beginner, so, if the change is small, better do it on your side. BUT:

I didn't add RenderStep component, because while it is missing in the doc of GameObject, it is available in TS defs - for eaxmple in Sprite (and in defs it is directly under GameObject, so available for all GameObjects). I am not sure, where did it get there from?

BenjaminDRichards:
I'm at a bit of a loss myself, but I did check around and fix another couple of missing type documentations in Layer. So this was a productive issue! I'll close this PR but the changes are implemented.

PR #7082, opened 2025-03-21
TypeScript: Missing property 'cache' in FileConfig

raaaahman:
This PR:

* Fixes a bug

Describe the changes below:

- Add the missing 'cache' property in the FileConfig type definition.

Reasoning:

I didn't make this property optional to force user being conscious that a non cached file will be of no use in the majority of cases.


photonstorm:
👍

PR #7085, opened 2025-03-25
Fix bounds calculation in getFrameBounds

jjcapellan:

* Fixes a bug

Describe the changes below:
The range of the loop in *Texture#getFrameBounds* has been adjusted to not include the last frame (__BASE), as it caused an incorrect calculation of the bounds (allways {x:0, y:0, w: textureWidth, h: textureHight}).  
You can see the problem in this code: https://phaser.io/sandbox/tqnwvEda . If the texture comes from an atlas frame, the framebounds of the texture should be the same as the bounds of the atlas frame.


photonstorm:
Thanks, this is merged into 4.0.0.

PR #7091, opened 2025-03-31
Fix Camera rotateTo() and change destination argument

### LINKED ISSUE #7090: Camera `rotateTo()` timing is wrong
## Version

* Phaser Version: 3.24.1 and later, maybe since 3.23.0

## Description

The camera `rotateTo()` effect's apparent timing and easing don't match the description.

## Example Test Code

https://phaser.io/sandbox/Zbcuwnsp

Click to **rotateTo** button and compare `rotateToEffect.progress` and `rotateToEffect.current`. `progress`  looks correct for the given duration but `current` changes too quickly and nonlinearly.

samme:
This PR

* Adds a new feature (breaking change)
* Fixes a bug

I started fixing #7090 and ended up changing the destination argument.

### Updates (breaking change)

In Phaser 3's [Camera.rotateTo()](https://docs.phaser.io/api-documentation/class/cameras-scene2d-camera#rotateto), when the `shortestPath` argument was `false` (the default value), the sign of the `radians` argument indicated the direction of rotation and the absolute value of the `radians` argument indicated the destination angle. In Phaser 4, the `radians` argument is now called `angle` and represents the destination angle as a signed value. You can use the new functions `Phaser.Math.Angle.GetClockwiseDistance()` or `Phaser.Math.Angle.GetCounterClockwiseDistance()` to calculate rotation distances in a specific direction.

### Bug Fixes

- Fixes #7090

### DIFF
--- a/src/cameras/2d/Camera.js
+++ b/src/cameras/2d/Camera.js
@@ -442,27 +442,26 @@ var Camera = new Class({
     },
 
     /**
-     * This effect will rotate the Camera so that the viewport finishes at the given angle in radians,
-     * over the duration and with the ease specified.
+     * Rotate the Camera to the given angle over the duration and with the ease specified.
      *
      * @method Phaser.Cameras.Scene2D.Camera#rotateTo
      * @since 3.23.0
      *
-     * @param {number} radians - The destination angle in radians to rotate the Camera viewport to. If the angle is positive then the rotation is clockwise else anticlockwise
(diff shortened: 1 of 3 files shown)


zekeatchan:
Hi @samme. Thanks for the PR. We have merged it with the `v4.0.0` branch. It will be part of the next release. Do test it out and let us know if you encounter any issues.

PR #7092, opened 2025-03-31
Add angle distance functions

samme:
This PR

* Adds a new feature

After #7091 I thought there should be a way to make a rotation in one direction or another.

## New Features

- `Phaser.Math.Angle.GetClockwiseDistance()` gets the shortest nonnegative angular distance between two angles.
- `Phaser.Math.Angle.GetCounterClockwiseDistance()` gets the shortest nonpositive angular distance between two angles.
- `Phaser.Math.Angle.GetShortestDistance()` gets the shortest signed angular distance between two angles. (This is like `Phaser.Math.Angle.ShortestBetween()` but in radians.)

## Tests

- [Tests](https://codepen.io/samme/pen/mydvKbO?editors=0110 "Phaser 4 angle distance functions")


zekeatchan:
Hi @samme. Thanks for the PR. We have merged it with the `v4.0.0` and also added it to the `master` branch. It will be part of the next release. Do test it out and let us know if you encounter any issues.

PR #7095, opened 2025-04-02
Update ArcadeColliderType.js

ospira:
This PR

* Updates the Documentation
* Fixes a bug

Describe the changes below:

`ArcadeColliderType` typedef/jsdoc is mildly incorrect. The array versions of this type do not need to be of a single type (see collider code [here](https://github.com/phaserjs/phaser/blob/master/src/physics/arcade/World.js#L1816) for example) but rather if it's an array that array can be a mix of any of the various single types, as https://github.com/phaserjs/phaser/blob/master/src/physics/arcade/World.js#L1934 is going to test each member of the array anyway. Note, this still works if you pass an array of a single type, it just doesn't force you to do that, as it isn't actually necessary.

### DIFF
--- a/src/physics/arcade/typedefs/ArcadeColliderType.js
+++ b/src/physics/arcade/typedefs/ArcadeColliderType.js
@@ -1,6 +1,6 @@
 /**
  * An Arcade Physics Collider Type.
  *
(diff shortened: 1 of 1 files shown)


ospira:
note in the diff also some of the individual types were not carried over to the array types prior to this change, i believe that is also incorrect. the array type is now simply an array that can have any of the preceding types of ArcadeColliderType as a member...

photonstorm:
Thanks, this has been merged to the 4.0.0 branch.

PR #7107, opened 2025-04-09
Fix JSDocs return type of `Phaser.Tweens.Tween#getValue`

DayKev:
This PR:
* Updates the Documentation

Describe the changes below:

The return type of `Phaser.Tweens.Tween#getValue` is changed from `number` to `number|null` to reflect its actual return types.


photonstorm:
👍

PR #7108, opened 2025-04-09
Fix audio files not loading from Base64 data URIs

bagyoni:

* Fixes a bug

Describe the changes below:

When sending an XHR with `responseType: 'arraybuffer'`, the `response` field will get populated with an `ArrayBuffer`. The fake XHR code path doesn't do this, so audio files, which expect this field, can't be loaded from base64.
The fix uses [this](https://stackoverflow.com/a/41106346) workaround to do the conversion. Tested with a few jsfxr outputs and found no issues.


photonstorm:
Note: Merged to v4, needs back-porting to v3.

photonstorm:
Already merged in v4.

PR #7117, opened 2025-04-22
Add missing thisValue for forEach

TadejZupancic:

* Fixes a bug

Describe the changes below:

this value was accessed inside the functions in forEach, but the value for this was not provided.
I noticed that in some places var _this = this approach is used for that and in some places this is added as the second parameter of forEach. I added this as the second parameter of forEach to fix that. Maybe it would be good to make it the same everywhere.


photonstorm:
👍

PR #7118, opened 2025-04-25
Added "external" to FiltersInternalExternal type definition.

OuttaBounds:
Added "external" to FiltersInternalExternal type definition, as it was missing before.


photonstorm:
Thanks - I've fixed this in the source files, so the defs will be correct on the next build.

PR #7123, opened 2025-04-26
Fix createFromTiles to handle multiple tilesets when using sprite sheets

### LINKED ISSUE #7122: TileMap.createFromTiles doesn't properly handle multiple tilesets
## Version

* Phaser Version: 3.88.2

## Description
Tilemap.createFromTiles with `useSpriteSheet` specified in the SpriteConfig assigns the frame number just using the current tile index without taking into account that the tile index doesn't always start from 1 for all tilesets other than the very first one.

## Example Test Code

https://stackblitz.com/~/github.com/vikerman/my-phaser-game

The sprites created from the second tileset are all wrong
Image

To test the fix locally comment lines 221-230 and uncomment lines 233-242 in Game.ts in the StackBlitz .

Sprites are properly created after the fix

Image

## Additional Information

The line at

https://github.com/phaserjs/phaser/blob/c62baca677399e098ce21bb49757974c4742895b/src/tilemaps/components/CreateFromTiles.js#L86

                config.frame = tile.index - 1;

needs to be changed to

                 config.frame = tile.index - tile.tileset.firstgid;

Tested the fix locally by creating a secondary copy of createFromTiles in my project and applying the above fix.

vikerman:
This PR 

* Fixes a bug

Describe the changes below:

Fixes #7122.

Properly use the starting index of the tileset while calculating the frame to be used for the sprite from the tileset image.

### DIFF
--- a/src/tilemaps/components/CreateFromTiles.js
+++ b/src/tilemaps/components/CreateFromTiles.js
@@ -83,7 +83,7 @@ var CreateFromTiles = function (indexes, replacements, spriteConfig, scene, came
             if (config.hasOwnProperty('useSpriteSheet'))
             {
                 config.key = tile.tileset.image;
-                config.frame = tile.index - 1;
+                config.frame = tile.index - tile.tileset.firstgid;
             }
 
             sprites.push(scene.make.sprite(config));


photonstorm:
👍

PR #7126, opened 2025-04-29
Fix: prevent gamepad buttons from sending bogus down events on init

LexVocoder:

* Fixes a bug

Describe the changes below:

Phaser is always initializing Gamepad buttons as not being pressed. This creates a problem when reading gamepad inputs in one scene, which starts another, and then reading gamepad inputs in the second scene.  If the player holds the button down for a tiny, tiny bit in the first scene, the second scene sees a bogus `down` event.

The fix introduces a new, boolean parameter to gamepad Button's constructor, `isPressed`.  When gamepad is initializing its buttons, it checks the current value for the button.  If it's greater or equal to 0.5, Gamepad considers the Button pressed and passes true for this parameter.  Otherwise, it passes false.  Button then uses the isPressed parameter to initialize its official `pressed` property.  By initializing with the current pressed state of the gamepad Button, the fix eliminates the bogus event at scene start.

[This Phaselet](https://phaser.io/sandbox/HgqW9szf) tests for the bug.  Grab a gamepad and follow the directions.

### DIFF
--- a/src/input/gamepad/Button.js
+++ b/src/input/gamepad/Button.js
@@ -19,12 +19,13 @@ var Events = require('./events');
  *
  * @param {Phaser.Input.Gamepad.Gamepad} pad - A reference to the Gamepad that this Button belongs to.
  * @param {number} index - The index of this Button.
+ * @param {boolean} isPressed - Whether or not the button is already being pressed at creation time.  This prevents the Button from emitting spurious 'down' events at first update.
  */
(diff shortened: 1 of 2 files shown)


LexVocoder:
I used 0.5, because threshold is not available at time of Button construction.

photonstorm:
Thanks, this fix has been merged into the 4.0.0 branch and will be part of that release.

PR #7136, opened 2025-05-12
Docs/v4.0.0/Arcade Typescript

ospira:
This PR

* Updates the Documentation
* Fixes a bug(?)

Describe the changes below:

Delete ArcadeCollider.js (duplicate type) and some other fixes, but would appreciate guidance on the following question before leaving draft mode:

Should it be allowed to pass just a `GameObject` into the collider? Shouldn't it be `GameObjectWithBody`? Without a body GameObject ends up being passed to the collider ok, but I believe it will never collide, and just silently dead end... for example if one GameObject has a body but the other does not, it will just dead end in this block: https://github.com/phaserjs/phaser/blob/v4.0.0/src/physics/arcade/World.js#L1921-L1935

This might be ok, but also I would think it could also be thought of as a bug... and something the type checker should prevent, because why are you passing a gameobject into a collider without a body? Asking this as a discussion though because a). I may have missed something and b). it also might be better to keep this API how it is and simply emit a warning in this case instead of changing the API/type def. I have changed it here so you can see the diff, but if I'm wrong about this happy to defer. It just seems to me not warning the user that their gameobjects are missing bodies could in this case lead to poor devx...

### DIFF
--- a/src/physics/arcade/World.js
+++ b/src/physics/arcade/World.js
@@ -946,7 +946,6 @@ var World = new Class({
         this._elapsed += delta;
 
         //  Update all active bodies
-        var body;
         var bodies = this.bodies;
 
         //  Will a step happen this frame?
--- a/src/physics/arcade/World.js
+++ b/src/physics/arcade/World.js
@@ -1972,8 +1972,8 @@ var World = new Class({
      * @method Phaser.Physics.Arcade.World#canCollide
      * @since 3.70.0
      *
-     * @param {Phaser.Types.Physics.Arcade.ArcadeCollider} body1 - The first body to check.
-     * @param {Phaser.Types.Physics.Arcade.ArcadeCollider} body2 - The second body to check.
+     * @param {Phaser.Types.Physics.Arcade.ArcadeColliderType} body1 - The first body to check.
+     * @param {Phaser.Types.Physics.Arcade.ArcadeColliderType} body2 - The second body to check.
      *
      * @return {boolean} True if the two bodies will collide, otherwise false.
      */
--- a/src/physics/arcade/typedefs/ArcadeCollider.js
+++ b/src/physics/arcade/typedefs/ArcadeCollider.js
@@ -1,6 +1,0 @@
-/**
- * An Arcade Physics Collider Type.
- *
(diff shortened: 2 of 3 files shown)


samme:
> Should it be allowed to pass just a GameObject into the collider? Shouldn't it be GameObjectWithBody?

`GameObjectWithBody` is more correct but it's possible this change would frustrate some authors who were adding physics bodies correctly (for Phaser) but not typing them correctly (for TypeScript). On the whole it might be worth it though.

> it also might be better to keep this API how it is and simply emit a warning in this case

As I recall you can give an array of game objects to a collider, later destroy one of those game objects (removing the body) **without** removing it from the array, and the collider will just skip the missing body without throwing an error. So from one perspective that's a good feature. But it is confusing if you forgot to add physics bodies then see the collider doing nothing at all.

ospira:
> > Should it be allowed to pass just a GameObject into the collider? Shouldn't it be GameObjectWithBody?
> 
> `GameObjectWithBody` is more correct but it's possible this change would frustrate some authors who were adding physics bodies correctly (for Phaser) but not typing them correctly (for TypeScript). On the whole it might be worth it though.
> 
> > it also might be better to keep this API how it is and simply emit a warning in this case
> 
> As I recall you can give an array of game objects to a collider, later destroy one of those game objects (removing the body) **without** removing it from the array, and the collider will just skip the missing body without throwing an error. So from one perspective that's a good feature. But it is confusing if you forgot to add physics bodies then see the collider doing nothing at all.

yep, the fact that it skips without causing an overt error makes it tempting to just leave it as it is, but on the other hand thinking about it more and considering your comments, it should actually be changed. the type isn't helpful if it doesn't make use of the more correct `GameObjectWithBody`... for those who already typed with Phaser it would be a small and hopefully worthwhile change - most of them probably won't have to update at all if they passed a body in correctly... might even catch some bugs out there... The points you made about runtime flexibility will of course persist, but I think if you are using TypeScript this is the kind of thing you want to be aware of. 

thanks for the feedback, going to mark out of draft now but feel free to give more feedback when you get a chance. I appreciate it and know you have bigger fish to fry atm. 🫡

photonstorm:
I think this is fine, so happy to merge.

PR #7145, opened 2025-05-19
Updated type of param in MatterPhysics.collision.collides to support null.

ixonstater:
**Nullable parameter type bugfix**

The Matter.collision.collides method accepts an optional third parameter to reuse existing collision pairs and reduce GC overhead.
Passing `null` to this method forces new collision pairs to be created and is supported per the Matter.js docs:
* https://brm.io/matter-js/docs/classes/Collision.html#method_collides

This change allows `null` to be passed here without a cumbersome cast eg. `null as any`.


photonstorm:
👍

PR #7154, opened 2025-05-28
Ability to have different radii for each corner of a rounded rectangle

### LINKED ISSUE #7152: Rounded rectangles have jagged edges?
## Version

* Phaser Version: 3.90.0

* Operating system: macOS 15.4.1

* Browser: Google Chrome

## Description

I'm not sure if this is a bug or if this is explicitly how its meant to behave, but I do find it strange that the rounded corners on rectangles set via `setRounded()` have jagged edges when the radius is about 9 or less. At 10 and above it starts to look a bit better, but still looks strange, especially when compared to rounded corners created via a graphics gameobject.

## Example Test Code

https://phaser.io/sandbox/XE4yedss

This sandbox compares a rectangle with rounded corners (located on the left side) with a graphics object with similar settings, located on the right side.

ChrisCPI:

* Adds a new feature
* ~~Fixes a bug~~

Describe the changes below:

EDIT: [This commit](https://github.com/phaserjs/phaser/commit/4ea4f95) fixes the issue of jagged edges. This PR has been updated to match that logic. Since now this PR won't directly fix this issue, any mention of it will be crossed out below.

---

In summary, this PR adds the ability to set different radii for each corner of a rectangle, ~~and no longer reduces the amount of segments to render (fix #7152).~~

I hope I followed all the rules for doing a pull request, please do let me know if I need to adjust anything. I'm especially not sure of how to properly handle the JSDoc comments.

### All of the changes (to the Rectangle class, unless otherwise noted)
- The properties `radiusTopLeft`, `radiusTopRight`, `radiusBottomLeft` and `radiusBottomRight` have been added. These define the radii for their respective corners.
- The `isRounded` property has been changed to a getter. Returns `true` if any of the new aforementioned radius properties are not equal to 0.
- The parameters for the `setRounded` method are changed to `topLeft`, `topRight`, `bottomLeft` and `bottomRight`, for the ability to set the radius of each corner. If ONLY `topLeft` is defined, then that radius is applied to all 4 corners (this prevents existing use cases of this method from breaking).
- The `radius` property now acts as a getter and a setter. As a getter, it returns `radiusTopLeft`. As a setter, it runs `setRounded`, passing in the value its being set to.
- `updateRoundedData` has been adjusted to accommodate setting individual radii of each corner and edge.
  - ~~When calculating the segments, it no longer divides the radius by 5. This fixes an issue that was causing the corners to look rough in the WebGL renderer.~~
- (RectangleCanvasRenderer) Updated the `DrawRoundedRect` function to support drawing individual radii for each corner.
  
  
image

### DIFF
--- a/src/gameobjects/shape/rectangle/Rectangle.js
+++ b/src/gameobjects/shape/rectangle/Rectangle.js
@@ -300,10 +300,10 @@ var Rectangle = new Class({
         var y = halfHeight;
 
         //  The number of segments is based on radius (more segments = larger radius)
-        var tlSegments = Math.max(1, Math.floor(tl));
-        var trSegments = Math.max(1, Math.floor(tr));
-        var blSegments = Math.max(1, Math.floor(bl));
-        var brSegments = Math.max(1, Math.floor(br));
+        var tlSegments = Math.max(4, Math.min(16, Math.ceil(tl / 2)));
+        var trSegments = Math.max(4, Math.min(16, Math.ceil(tr / 2)));
+        var blSegments = Math.max(4, Math.min(16, Math.ceil(bl / 2)));
+        var brSegments = Math.max(4, Math.min(16, Math.ceil(br / 2)));
 
         //  Create points going clockwise from top-left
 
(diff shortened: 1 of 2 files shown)


photonstorm:
Thanks for originally opening this PR. I'm going to close it because we're moving to Phaser v4. If you think this feature is still valid for v4 you're welcome to resubmit it once v4 is live.

PR #7163, opened 2025-06-11
Use first item of an array of tilesets when making a TilemapGPULayer

### LINKED ISSUE #7158: Uncaught TypeError: e.getAnimationDataIndexMap is not a function
## Version
* Phaser Version: 4.0.0-rc4
* Operating system: Arch Linux
* Browser: Brave / Chrome

## Description

If I pass multiple tilesets as array to create GPU layer then it breaks.

Here is exception:
```
4.0.0-rc.4.min.js:1 Uncaught TypeError: e.getAnimationDataIndexMap is not a function
    at initialize.generateLayerDataTexture (4.0.0-rc.4.min.js:1:1128591)
    at initialize.setTileset (4.0.0-rc.4.min.js:1:1128452)
    at new initialize (4.0.0-rc.4.min.js:1:1128243)
    at initialize.createLayer (4.0.0-rc.4.min.js:1:1117782)
    at MainScene.create (<anonymous>:18:13)
    at initialize.create (4.0.0-rc.4.min.js:1:1005143)
    at initialize.loadComplete (4.0.0-rc.4.min.js:1:1004532)
    at o.emit (4.0.0-rc.4.min.js:1:1643)
    at initialize.loadComplete (4.0.0-rc.4.min.js:1:523064)
    at initialize.fileProcessComplete (4.0.0-rc.4.min.js:1:522752)
```

## Example Test Code

https://phaser.io/sandbox/bciiLd65

ChrisCPI:
This PR

* Fixes a bug

Describe the changes below:

In the `createLayer` method on a tilemap, it says that a string or tileset (or an array of those) can be passed in for `tileset`. However, the TilemapGPULayer (according to its documentation) can only support one tileset. And so, when the `gpu` parameter was true, the GPU layer didn't expect to see an array, and so when passing in an array, it breaks.

Now, when `gpu` is set to true, it checks if `tileset` is an array. If it is, then it will only pass in the first item of that array when creating the layer. If the array has more than one item, it gives a warning.

Fix #7158

### DIFF
--- a/src/tilemaps/Tilemap.js
+++ b/src/tilemaps/Tilemap.js
@@ -590,7 +590,7 @@ var Tilemap = new Class({
      * @param {(string|string[]|Phaser.Tilemaps.Tileset|Phaser.Tilemaps.Tileset[])} tileset - The tileset, or an array of tilesets, used to render this layer. Can be a string or a Tileset object.
      * @param {number} [x=0] - The x position to place the layer in the world. If not specified, it will default to the layer offset from Tiled or 0.
      * @param {number} [y=0] - The y position to place the layer in the world. If not specified, it will default to the layer offset from Tiled or 0.
-     * @param {boolean} [gpu=false] - Create a TilemapGPULayer instead of a TilemapLayer. This option is WebGL-only. A TilemapGPULayer is less flexible, but can be much faster.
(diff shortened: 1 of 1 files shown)


BenjaminDRichards:
I tested this and it works exactly as intended. Let's get that merged!

PR #7167, opened 2025-06-16
NineSlice improvements

### LINKED ISSUE #7110: NineSlice setFrame() doesnt work when updateSize is false
## Version
* Phaser Version:
3.88.2 
* Operating system:
macOS 15.3.1
* Browser:
Google Chrome

## Description

When `setFrame()` is called on a NineSlice game object, and the `updateSize` parameter is set to `false`, the frame does not appear to update.

## Example Test Code

https://phaser.io/sandbox/3r6cWwMn

## Additional Information

In the above sandbox, there is one normal image and one nineslice, both with the same texture and frame. Both have the exact same interactive and listeners set to them, where the frame changes on `pointerover` and on `pointerout`. However, only the normal image appears to have its frame changed.

ChrisCPI:

* Fixes some bugs

Describe the changes below:
- Setting `_width` and `_height` in `setSlices` is now clamped to what it should be.
- The setters for `width` and `height` now run `setSize` (the `updateVertices` call has also been moved there, and `setSize` now sets `_width` and `_height` instead of calling the setters). This fixes an issue where the hit area for input wouldn't be updated when setting the width/height in this manner.
- `setFrame` will always run `updateUVs` (fix #7110). Before this it was only being run through `setSizeToFrame`, meaning the frame would not appear to change when the `updateSize` parameter was false.

### DIFF
--- a/src/gameobjects/nineslice/NineSlice.js
+++ b/src/gameobjects/nineslice/NineSlice.js
@@ -1020,6 +1020,10 @@ var NineSlice = new Class({
         {
             this.setSizeToFrame();
         }
+        else
+        {
+            this.updateUVs();
+        }
 
         if (this._originComponent && updateOrigin)
         {
@@ -1032,8 +1036,6 @@ var NineSlice = new Class({
                 this.updateDisplayOrigin();
             }
         }
-
-        this.updateUVs();
 
         return this;
     },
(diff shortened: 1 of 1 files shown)


photonstorm:
Thanks for originally opening this PR. I'm closing it because we're no longer updating v3 and preparing v4 for launch. If you think it's still a valid addition to v4, then you're welcome to resubmit against that once it is live.

PR #7168, opened 2025-06-17
Add support for rendering NineSlice via canvas

ChrisCPI:

* Adds a new feature

This PR adds the ability for NineSlice game objects to render via the canvas renderer, done by drawing 9 separate images (or 3, if it is set to be a three-slice) based on the width, height, leftWidth, rightWidth, topHeight and bottomHeight properties of the NineSlice. The vertices/quads are not used here.


photonstorm:
Thanks for originally opening this PR, but I'm going to close it as the Canvas renderer is now deprecated with the release of Phaser v4.

Issue #7172, opened 2025-06-22
Light positions are off when main camera setRotation is different from 0


FSDevelop:
## Version

* Phaser Version: v3.90.0
* Operating system: Windows 11
* Browser: Electron 33.2.1

## Description
Light positions become offset when the main camera rotation is set to any value other than 0.

In LightPipeline.js:200-202, the light position calculation applies scroll and zoom factors manually after using cameraMatrix.transformPoint(). This double-transformation causes incorrect positioning when the camera is rotated, since the camera matrix already handles rotation, scroll, and zoom together.

## Additional Information

Current problematic code:

  `cameraMatrix.transformPoint(light.x, light.y, tempVec2);`
  `this.set2f(lightName + 'position', tempVec2.x - (camera.scrollX * light.scrollFactorX * camera.zoom), height - (tempVec2.y - (camera.scrollY * light.scrollFactorY) * camera.zoom));`

Expected behavior: Lights should maintain their correct world positions regardless of camera rotation.

Actual behavior: Lights appear to drift/offset as the camera moves when camera.setRotation() is non-zero.        

Fix: Apply scroll factors before matrix transformation when camera is rotated, letting the camera matrix handle all transformations together.

XWILKINX:
Will this be fixed?

BenjaminDRichards:
> Will this be fixed?

Honestly, it's not a priority. Phaser 3 is unlikely to receive further first-party releases, and Phaser 4 rewrote the lighting code (fixing several other issues), so this class of problem won't recur.

That does mean it's safe to patch a fix into your Phaser 3 projects, as we're unlikely to change the code further. Or use Phaser 4, which I believe to be more stable and robust than Phaser 3 by this point, and is extremely close to release.

FSDevelop:
@BenjaminDRichards I'm already using Phaser4 because I know this is not an issue there, but people still using Phaser3 (still latest stable version) in their projects and don't have plans to migrate yet may experience this.

photonstorm:
There are no more updates for v3 planned from us, so they can either migrate, or we may consider a PR if we get enough of them to warrant a point release as a one-off bundle. Or monkey-patch it in.

PR #7173, opened 2025-06-22
Fix lights position when setting camera rotation to something different from 0

FSDevelop:

* Fix lights position when setting camera rotation to something different from 0

Describe the changes below:

Issue described here:
https://github.com/phaserjs/phaser/issues/7172

### DIFF
--- a/src/renderer/webgl/pipelines/LightPipeline.js
+++ b/src/renderer/webgl/pipelines/LightPipeline.js
@@ -197,9 +197,18 @@ var LightPipeline = new Class({
 
             var lightName = 'uLights[' + i + '].';
 
-            cameraMatrix.transformPoint(light.x, light.y, tempVec2);
-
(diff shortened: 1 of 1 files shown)


cavolarl (commented):
#7172 

Would it be possible to do it within the transformPoint logic instead?

FSDevelop:
I'll try @cavolarl

FSDevelop:
> #7172
> 
> Would it be possible to do it within the transformPoint logic instead?


@cavolarl 
I think the problem is that transformPoint is an operation that only does basic math transformations and doesn't know about scroll factors or camera data. The if/esle block I propose is mathematically necessary because rotated cameras need scroll applied before rotation (in world coordinates) while non rotated cameras can apply scroll after transformation (in screen coordinates with zoom). I can do it but it would break other systems that rely on it.

cavolarl:
That makes sense.

photonstorm:

FSDevelop:
No problem, thanks guys you are doing a great job! I love the library

PR #7181, opened 2025-07-11
Add ability to set scaleFactor of mask texture to preserve memory when using addMask filter

kimdanielarthur-cowlabs:
This PR 
* Adds a new feature

Description of feature
Addition of scaleFactor paramter to filters.addMask so that the size of the underlying texture used for masks can be controlled in order to preserve memory or work around maximum texture sizes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "phaser",
-  "version": "3.80.0-beta.3",
+  "version": "4.0.0-rc.4",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "phaser",
-      "version": "3.80.0-beta.3",
+      "version": "4.0.0-rc.4",
       "license": "MIT",
       "dependencies": {
         "eventemitter3": "^5.0.1"
--- a/src/filters/Mask.js
+++ b/src/filters/Mask.js
@@ -61,16 +61,27 @@ var Controller = require('./Controller');
  * @param {boolean} [invert=false] - Whether to invert the mask.
  * @param {Phaser.Cameras.Scene2D.Camera} [viewCamera] - The Camera to use when rendering the mask with a GameObject. If not specified, uses the scene's `main` camera.
  * @param {'local'|'world'} [viewTransform='world'] - The transform to use when rendering the mask with a GameObject. 'local' uses the GameObject's own properties. 'world' uses the GameObject's `parentContainer` value to compute a world position.
+ * @param {number} [scaleFactor=1] - Scale factor to apply to the underly mask texture. Can be used to balance memory usage and needed mask precision
  */
 var Mask = new Class({
     Extends: Controller,
 
-    initialize: function Mask (camera, mask, invert, viewCamera, viewTransform)
+    initialize: function Mask (camera, mask, invert, viewCamera, viewTransform, scaleFactor)
     {
         if (mask === undefined) { mask = '__WHITE'; }
         if (invert === undefined) { invert = false; }
+        if (scaleFactor === undefined) { scaleFactor = 1; }
 
         Controller.call(this, camera, 'FilterMask');
+
+        /**
+         * The scale factor to apply to the underlying mask texture.
+         *
+         * @name Phaser.Filters.Mask#scaleFactor
+         * @type {number}
+         * @since 4.0.0
+         */
+        this.scaleFactor = scaleFactor;
 
         /**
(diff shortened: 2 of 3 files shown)


BenjaminDRichards:
This is a useful feature, for the reasons stated. The implementation initially seems to work well.

### Problem

However, I'm concerned that there are situations where it would cause unexpected side effects. This implementation uses `scale` properties in the dynamic texture capture config, and these override the properties of the masking object. If that object had a non-default scale (i.e. not 1), this would ignore that value and produce unwanted results. (The override would not persist; it is only used during capture rendering.) This isn't common (a complex mask is probably rendered via several objects inside a vanilla `Container` with scale 1), but it could happen.

I've tested camera scroll, and it actually breaks the scale approach. The reasons can be summed up as "nested transforms are really complicated".

### Potential Approaches

I've experimented with adjusting the camera instead, but a camera transform is a bit of persistent data generated early in the rendering process, and it doesn't feel like regenerating it twice for this situation is the correct solution. Even if it is correct, we have to be very careful not to introduce corruption to other parts of the render pipeline.

We could also apply the scale factor solely to the `DynamicTexture` resolution, and rely on the user to scale the mask object when they set the scale factor. This is not as ergonomic, though.

### Necessity of Feature

We should also consider whether this is a significant saving. I think it is, but the reasons are worth investigating.

First, the `DynamicTexture` where the mask is rendered is persistent. This means that every mask in the game adds to the memory footprint.

Now, the mask isn't the only texture used during rendering. A Mask is a Filter, and filters use framebuffers to render the image in progress until it is finally drawn to the screen. Each framebuffer uses memory, and is subject to the same resolution limitations. However, filter framebuffers are drawn from a pool, so only a few chunks of memory are reserved as the game runs; they might be recycled by different masks.

Thus, a large number of masks could have a significant memory impact, but only from the Mask part, not the Filter part. The idea of scaling down the `DynamicTexture` is appropriate, because this is unique to Masks.

We don't need to worry about maximum texture sizes, as the mask texture and framebuffer are both set from the same DrawingContext size; if the mask texture would fail, so would the framebuffer, and then we'd have bigger problems. It's generally a good idea to be careful with filters on the internal list, as they are sized to an object, and can thus potentially be bigger than the screen; but again, this is a problem that affects all filters.

### Final Approach

After some R&D and experimenting with camera transform regeneration, I think that we should break this into two parts:

1. Adjust the size of the mask texture.
2. Adjust the rendered size of the mask object.

Of these, 2 is too complicated, because masks can be set up in several different ways, and most of them resist messing with the current camera. That leaves us with 1, simply adjusting the size of the mask texture. That addresses the core issue - oversized texture resources.

Part 2 can then be left to the developer, who knows more about their game - how cameras scroll etc - and can set up all aspects of the mask to be compatible with the scaled size. We would add documentation to the `sizeFactor` param/property to explain basic use. In most cases it would simply entail one extra line of code, something like `maskContainer.setScale(0.5)`.

Does this make sense? If so, I'll implement it with reference to this PR.

kimdanielarthur-cowlabs:
That is a great summary and the conclusion makes sense to me that the developer would have to manage sideeffects when overriding defaults.

As a note related to mask texture sizes: the current phaser code has the mask max texture size hardcoded to 4096x4096 rather than using max texture size as queried from opengl capabilities.

This results in games running at high resolution 4k/2xdpr unable to, for example, have any full screen clipping masks. It also becomes unpredictable to developer as its normal to assume the only limit would be max texture size?

See the Phaser.Renderer.WebGL.DrawingContextPool#get functionimage

BenjaminDRichards:
Well spotted. The 4096 value was a deliberate conservative limit, but as you note, we do actually have the device limit in `maxTextureSize`. Ordinarily I'd want to be careful about exceeding minimum spec, to avoid creating games that run on a developer's desktop but not on the target mobile devices; but these textures are temporary and internally generated, so they shouldn't pose a problem.

There are a couple of other places where textures are limited to 4096x4096, but these are "data textures" for high performance GPU Layer objects. I don't want to free those, because that _does_ introduce non-obvious compatibility problems. For example, it would be possible to use a tile map 5000 tiles wide on one device, but not on others.

I'll go ahead and implement both these changes, then close this out. Thanks for the input!

kimdanielarthur-cowlabs:
Yes makes perfect sense, nice one!

BenjaminDRichards:
I've pushed the changes with appropriate mentions. Those should be in the next release.

PR #7183, opened 2025-07-22
Add optional sortByY param to tilemap createFromObjects method

saintflow47:

* Adds a new feature

Describe the changes below:
Setting the optional sortByY attribute in the config sorts tilemap layer objects before creation by y position which leads to objects that are lower on screen being displayed on top of higher objects. Tiled uses this sort order by default.


photonstorm:
👍

PR #7192, opened 2025-08-28
Update JSDoc for `Phaser.Math.Wrap`

DayKev:
This PR:

* Updates the Documentation

Describe the changes below:

The doc now mentions that it wraps between `min` and `max - 1` instead of between `min` and `max`,
to match the behavior of the function.


photonstorm:

PR #7194, opened 2025-09-09
fixed memory leak on mobile pipeline with spine

dhashvir:

* Fixes a bug - https://github.com/phaserjs/phaser/issues/7193

Describe the changes below:

It only adds the attrib location if it is found. This prevents this array from growing infinitely.


photonstorm:

PR #7195, opened 2025-09-12
add build dir to .eslintignore

justin-calleja:
so that viewing built phaser.js in editor is not full or errors


photonstorm:
Thanks, this has been merged into the 4.0.0 branch

Issue #6642, opened 2023-10-04
SpinePlugin breaks generated images file path


rez23:
## Version

* Phaser Version: nodejs phaser@3.60.0
* Npm Version: '8.19.3',
* NodeJs Version: '19.2.0',
* TypeScript version: 5.2.2

* Operating system: Windows

* Browser: Chrome/Edge/Nodejs

The reference environment where I have encountered this issue is based on typescript and use nodejs versione of phaser 3

## Description
The issue affects the phaser path system. when you set ```baseUrl``` and ```path``` properties of the loader via ```setBaseUrl()``` and ```setPath()``` functions and try to load spine's atlases/Jsons using phaser's Spine plugin (inside this repository).

When the loader tries to load the images obtained from the parsing of the atlases/jsons files provided to the loader through ```spine()``` function, the loader appends twice baseUrl to path and the result is an error in the GET stage when the loader tries to charge the Spine's multilayered images.

## Example Test Code

You can reproduce (i suppose) the error with a code like:
```js
#inside game scene
this.load.setBaseUrl('assets');

this.load
        .setPath('spine')
        .spine('MAIN.animations', 'spine.json', 'spine.atlas');

```

Supposing that inside my spine files I have linked a spine.png, this statement will result in the trying of the loader to load this image from "assets/assets/spine/spine.png" instead of "assets/spine/spine.png" that is what should happen.

## Additional Information
I have circumscribed (and potentially fixed) this issue to this piece of code inside the onFileComplete() function present inside SpineFile,js in the spine plugin source code:
```js
// This start at line 159 of SpineFile,js
var config = this.config;
var loader = this.loader;

var currentBaseURL = loader.baseURL;
var currentPath = loader.path;
var currentPrefix = loader.prefix;

var baseURL = GetFastValue(config, 'baseURL', this.baseURL);
var path = GetFastValue(config, 'path', file.src.match(/^.*\//))[0]; //Here is the problem
var prefix = GetFastValue(config, 'prefix', this.prefix);
var textureXhrSettings = GetFastValue(config, 'textureXhrSettings');

loader.setBaseURL(baseURL);
loader.setPath(path);
loader.setPrefix(prefix);

```

I've a possible solution on [this fork](https://github.com/photonstorm/phaser/compare/master...rez23:phaser:spine_png_load_fix)

spayton:
[...]

PR #7210, opened 2025-10-26
fix(loader): recognize file:// as absolute URL in GetURL() to support offline environments

aomsir:
#### Summary
Under the **`file://` protocol**, `Phaser.Loader.GetURL()` does not treat `file://` URLs as absolute. When a `baseURL` is set, it
gets **prepended** to an already-absolute `file://` path, producing double-prefixed URLs like:

```
file:///Users/.../spine--demo.../file:///Users/.../assets/spine/spineboy-pma.png
```

**Important scope clarification:** In my environment, **_all other assets load correctly_** under `file://` + `setBaseURL(...)`.  
The problem manifests specifically when **Spine atlas-derived page images (the `.png` listed inside the `.atlas`)** are queued.  
Those page images sometimes end up with an **absolute `file://` in `file.url`**, and `GetURL()` then prepends `baseURL` again.

This PR adds `file://` to the absolute-URL check so `GetURL()` returns the original URL and does not concatenate `baseURL`.
This keeps atlas page images consistent with all other file types that already behave correctly in the same environment.

---

#### Why this matters
In our setup we **must** run under **`file://`** (no local HTTP server available) **and** we **must** use `this.load.setBaseURL(...)`
to keep asset resolution consistent across browsers. Without this patch, using both `file://` + `baseURL` breaks **Spine atlas page PNGs**,
while other assets remain fine. With this patch, Phaser treats `file://` just like `http(s)`, `blob:`, and `data:`.

---

#### Changes
```diff
// src/loader/GetURL.js

var GetURL = function (file, baseURL)
{
    if (!file.url)
    {
        return false;
    }

-   if (file.url.match(/^(?:blob:|data:|capacitor:\/\/|http:\/\/|https:\/\/|\/\/)/))
+   if (file.url.match(/^(?:blob:|data:|capacitor:\/\/|file:\/\/|http:\/\/|https:\/\/|\/\/)/))
    {
        return file.url;
    }
    else
    {
        return baseURL + file.url;
    }
};
```

---

#### Results

| Environment           | Case                                      | `file.url`                                 | `file.src`                                              | Outcome |
|-----------------------|-------------------------------------------|---------------------------------------------|---------------------------------------------------------|---------|
| HTTP / HTTPS          | any asset                                 | `assets/bg.png`                             | `http://localhost/.../assets/bg.png`                    | ✅ OK   |
| `file://` (before)    | **Spine atlas _page PNG_** (derived file) | `file:///Users/.../spineboy-pma.png`        | `file:///base/.../file:///Users/.../spineboy-pma.png`   | ❌ BAD  |
| `file://` (after)     | **Spine atlas _page PNG_** (derived file) | `file:///Users/.../spineboy-pma.png`        | `file:///Users/.../spineboy-pma.png`                    | ✅ OK   |
| `file://` (baseline)  | regular images / audio / json             | `assets/...`                                | `file:///base/.../assets/...`                           | ✅ OK   |

---

#### Notes
- No impact on `http(s)`, `blob:`, `data:`, or `capacitor://` environments.
- Patch is isolated to `GetURL()`; no other loader behavior is changed.
- Verified on Phaser 3.90.0 (macOS + Chrome) under `file://` with `setBaseURL(...)` required for consistent resolution.

### DIFF
--- a/src/loader/GetURL.js
+++ b/src/loader/GetURL.js
@@ -22,7 +22,7 @@ var GetURL = function (file, baseURL)
         return false;
     }
 
-    if (file.url.match(/^(?:blob:|data:|capacitor:\/\/|http:\/\/|https:\/\/|\/\/)/))
+    if (file.url.match(/^(?:blob:|data:|capacitor:\/\/|file:\/\/|http:\/\/|https:\/\/|\/\/)/))
     {
         return file.url;
     }


aomsir:
#### Additional context

I have reviewed [issue #6642](https://github.com/phaserjs/phaser/issues/6642) and confirmed that it **partially addresses** the `baseURL` + absolute path duplication issue for HTTP/HTTPS cases.

However, the **`file://` protocol is still affected**.  
Even with the fix from #6642 (`file.src` → `file.url`), when running under `file://` with a `baseURL` set, the loader still prepends `baseURL` to already-absolute `file://` URLs—most notably for **Spine atlas page PNGs**.  
This patch extends the absolute URL detection in `GetURL()` to include `file://`, ensuring parity between `file://` and other absolute schemes such as `http(s):`, `blob:`, and `data:`.

photonstorm:

PR #7253, opened 2026-02-26
Fix sub-pixel jitter after tile separation

### LINKED ISSUE #7252: Arcade Physics: tilemap tile separation leaves body.position at sub-pixel values, causing 1px visual oscillation
## Version

* Phaser Version: Phaser 3.90.0
* Browser: Chrome

## Description

When a sprite with Arcade physics collides with a tilemap layer, the tile separation code can leave body.position.y at fractional values (e.g., 63.9997 instead of 64). With pixelArt: true and integer tile sizes (16×16), this causes the sprite to visually oscillate 1px vertically each frame, it renders at Y=63 one frame and Y=64 the next.

The issue is specific to certain tile row boundaries and is reproducible regardless of room layout, gravity values, or body size. It affects both tilemap tile collision and separate static body collision when Phaser's separation algorithm is involved.

## Example Test Code

https://jsbin.com/zoxusuyeve

https://jsbin.com/zoxusuyeve/edit?html,js,output

## Additional Information

320×240 game, pixelArt: true, 16×16 tilemap, Arcade gravity {y: 800}
Sprite with body size ~12×14 standing on tiles at specific Y boundaries (e.g., row 4, pixelY=64)
Sprite visually vibrates 1px vertically while standing still

danipeck:
This PR:

* Fixes a bug

Describe the changes below:

Fixes #7252

ProcessTileSeparationX/Y can leave body.position at fractional values after subtracting the overlap, like 63.9997 instead of 64.

Tiles are at whole pixel coords so this shouldn't happen, but float math is gonna float math. In pixelArt games it shows up as sprites jittering 1px back and forth while just standing on a tile.

The fix: after the subtraction, if the result is within 0.01 of a whole number, snap it. Catches the float drift without messing up bodies that are supposed to be fractional (scaled sprites with non-integer dimensions, etc).

---

> [!NOTE]
> **Low Risk**
> Small, localized change to tile collision resolution; main risk is edge-case behavior changes for bodies that rely on tiny fractional positions near pixel boundaries.
> 
> **Overview**
> Fixes sub-pixel jitter after Arcade Physics tile separation by snapping `body.position.x`/`y` to the nearest integer when the post-separation value is *very close* to a whole pixel.
> 
> The new rounding is tolerance-based (`< 0.01`) so intentionally fractional positions (e.g., from scaled sprites / non-integer body sizes) are preserved, and the rest of the separation/bounce logic remains unchanged.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d01636b1b1e0b8b1713418349dd113457055e995. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/physics/arcade/tilemap/ProcessTileSeparationX.js
+++ b/src/physics/arcade/tilemap/ProcessTileSeparationX.js
@@ -27,6 +27,17 @@ var ProcessTileSeparationX = function (body, x)
     }
 
     body.position.x -= x;
+
+    // Float arithmetic can leave position at e.g. 63.9997, causing 1px
+    // jitter. Snap near-integer results but preserve intentionally
+    // fractional positions (non-integer body sizes from scaled sprites).
+    var rx = Math.round(body.position.x);
+
+    if (Math.abs(body.position.x - rx) < 0.01)
+    {
+        body.position.x = rx;
+    }
+
     body.updateCenter();
 
     if (body.bounce.x === 0)
--- a/src/physics/arcade/tilemap/ProcessTileSeparationY.js
+++ b/src/physics/arcade/tilemap/ProcessTileSeparationY.js
@@ -27,6 +27,17 @@ var ProcessTileSeparationY = function (body, y)
     }
 
     body.position.y -= y;
+
+    // Float arithmetic can leave position at e.g. 63.9997, causing 1px
+    // jitter. Snap near-integer results but preserve intentionally
+    // fractional positions (non-integer body sizes from scaled sprites).
+    var ry = Math.round(body.position.y);
+
+    if (Math.abs(body.position.y - ry) < 0.01)
+    {
+        body.position.y = ry;
+    }
+
     body.updateCenter();
 
     if (body.bounce.y === 0)


photonstorm:
My concern is that this fix causes the body to visually penetrate the tile:

image

danitpeck:
Good catch, snapping to the nearest can change the resolved position and cause overlap. I'll revisit and either make the snap direction-aware or avoid mutating the physics position directly

photonstorm:
Also worth adding this is likely solved via all the work we did around pixel art rendering in v4 (see https://github.com/phaserjs/phaser/blob/v4.0.0/docs/Phaser%204%20Pixel%20Art%20Guide/Phaser%204%20Pixel%20Art%20Guide.md)

BenjaminDRichards:
I've reviewed the test case in Phaser v4, and it looks like we no longer get flickering of this sort. Pixel art rendering got some upgrades which handle things more generally.

We _do_ get visual penetration, but I think that's intentional in this case, as the body is 14px tall on a 16px sprite.

I'm going to close this as no longer relevant to ongoing development of v4. But if you find floats floating buzzily in other situations - which they might, and you're clearly handling this stuff frequently - please file a new issue!

PR #7259, opened 2026-03-01
NineSlice: Implement tiled mode

skhoroshavin:
This PR adds a new feature.

Adds tiled rendering mode to the NineSlice game object via tileX and tileY constructor parameters. When enabled, the scalable regions of a nine-slice (or three-slice) are repeated across the object instead of being stretched. Each tile is still slightly stretched so that it remains visible in full, but this causes far less distortion than pure stretching. Of course, the texture should be seamless to avoid visible artifacts between tiles. The repeat counts are recalculated automatically when the object is resized, and the vertex/UV arrays are rebuilt as needed.

Examples:
* Without tiling: 
load-dialog

* With tiling: 
image

Tested on a personal project with screenshot-based UI tests - without enabling tiled rendering (and by default it is disabled), the output was pixel-perfect compared to the v4.0.0 branch.

Note that this is still a rough implementation. Ideally, tiled rendering would be done by implementing a proper grid where quad vertices are shared between adjacent quads rather than repeated per-quad, but I believe that is out of scope for this PR.

---

> [!NOTE]
> **Medium Risk**
> Changes core NineSlice vertex/UV generation and dynamically resizes the underlying vertex array, which could impact rendering correctness and performance across different sizes/frames (though tiling is opt-in via defaults).
> 
> **Overview**
> Adds an opt-in *tiled* mode for `NineSlice` via new `tileX`/`tileY` parameters, allowing the scalable regions to repeat rather than stretch.
> 
> To support this, `NineSlice` now calculates repeat counts on resize, rebuilds its `vertices` array when repeat counts change, and updates vertex positions/UVs using new row-emission helpers (`_updateVertexRow`, `_updateUVRow`). The factory/creator APIs and `NineSliceConfig`/TypeScript typings are updated to accept the new flags.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7b409b6c59df58899496e6d19de7f67a3e922b41. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/gameobjects/nineslice/NineSlice.js
+++ b/src/gameobjects/nineslice/NineSlice.js
@@ -121,6 +121,8 @@ var Vertex = require('./NineSliceVertex');
  * @param {number} [rightWidth=10] - The size of the right vertical column (B).
  * @param {number} [topHeight=0] - The size of the top horizontal row (C). Set to zero or undefined to create a 3 slice object.
  * @param {number} [bottomHeight=0] - The size of the bottom horizontal row (D). Set to zero or undefined to create a 3 slice object.
+ * @param {boolean} [tileX=false] - When enabled, the scalable horizontal regions are repeated across the object instead of being stretched. Each tile is still slightly stretched so that it remains visible in full, which may cause minor distortion but far less than pure stretching. The texture should be seamless to avoid visible artifacts between tiles.
+ * @param {boolean} [tileY=false] - When enabled, the scalable vertical regions are repeated across the object instead of being stretched. Each tile is still slightly stretched so that it remains visible in full, which may cause minor distortion but far less than pure stretching. The texture should be seamless to avoid visible artifacts between tiles.
  */
 var NineSlice = new Class({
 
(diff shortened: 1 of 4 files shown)


BenjaminDRichards:
This looks great! I've looked over the code and it all looks good. And I tested it, shrinking the NineSlice too far to see what happened (it worked fine).

I'd prefer a couple of tweaks to docstrings:

- `NineSliceFactory` should document the new `tileX` and `tileY` parameters
- `NineSlice#tileX`/`tileY` say they're readonly, like the pre-existing slice properties. This isn't strictly true; these can all be modified, but they won't do anything until `updateVertices()` is called. It's unlikely that anyone will ever need this, but I feel it should be documented:
  - Remove "Do not modify directly" from `tileX`/`tileY`
  - Add documentation to `updateVertices` to include these properties.

We could also imagine other use cases, like allowing more stretch in the tiling slices, or alternating flip on tiles (which is complicated because it has a 50% chance of breaking texture seams at one end). But this is so niche that I think it'll just be needless clutter.

But these docstring changes are minor, and I can do them myself. So I'll go ahead and merge this, and add those tweaks while I'm processing it.

PR #7273, opened 2026-04-05
fix(physics): update local body variable after creating MatterTileBody

### LINKED ISSUE #7267: MatterTileBody crashes with "Cannot read properties of null (reading 'inertia')" on flipped tiles with collision shapes
## Version

* Phaser Version: 3.90.0
* Operating system: macOS
* Browser: Chromium

## Description

When `Matter.World.convertTilemapLayer()` processes a tile that has `flipX` or
`flipY` set, it crashes with:

```
TypeError: Cannot read properties of null (reading 'inertia')
```

The crash occurs in `Body.scale()`, called from the `MatterTileBody` constructor.
The root cause is a scope bug: the constructor declares a local `body` variable
initialised to `null`, then creates the actual body via `setFromTileCollision()`
or `setFromTileRectangle()` — both of which set `this.body` but **never update
the local `body` variable**. Immediately after, the flip check reads from the
stale local:

```js
// MatterTileBody constructor (~line 99)
var body = GetFastValue(options, 'body', null);  // body = null

if (!body) {
    // ...
    this.setFromTileCollision(options);  // sets this.body, NOT body
}

if (tile.flipX || tile.flipY) {
    Body.scale(body, scaleX, scaleY, rotationPoint);  // body is still null → crash
}
```

Expected behaviour: flipped tiles with collision shapes should have their bodies
scaled/flipped correctly, not crash.

Affected file: `src/physics/matter-js/MatterTileBody.js`

## Example Test Code

Self-contained reproduction using Vitest + Playwright Chromium (no external
assets required — tilemap JSON is inline and the texture is generated):

```ts
import { describe, it, expect, afterEach } from 'vitest'
import Phaser from 'phaser'

const FLIPPED_HORIZONTALLY = 0x80000000

const TILEMAP_WITH_FLIPPED_TILE = {
  width: 2, height: 1, tilewidth: 16, tileheight: 16,
  orientation: 'orthogonal', renderorder: 'right-down',
  infinite: false, nextlayerid: 2, nextobjectid: 1,
  layers: [{
    id: 1, name: 'Ground', type: 'tilelayer',
    width: 2, height: 1, x: 0, y: 0,
    visible: true, opacity: 1, properties: [],
    data: [
      1,                          // normal tile
      1 | FLIPPED_HORIZONTALLY,   // same tile with flipX
    ],
  }],
  tilesets: [{
    firstgid: 1, name: 'stub',
    tilewidth: 16, tileheight: 16,
    spacing: 0, margin: 0, columns: 1, tilecount: 1,
    image: 'stub', imagewidth: 16, imageheight: 16,
    tiles: [{
      id: 0,
      objectgroup: {
        draworder: 'index', name: '', opacity: 1,
        type: 'objectgroup', visible: true, x: 0, y: 0,
        objects: [{
          id: 1, x: 0, y: 0, width: 0, height: 0,
          rotation: 0, visible: true,
          polygon: [
            { x: 0,  y: 0  },
            { x: 16, y: 0  },
            { x: 16, y: 16 },
            { x: 0,  y: 16 },
          ],
        }],
      },
    }],
  }],
}

class FlippedTileScene extends Phaser.Scene {
  error: Error | null = null

  constructor() { super({ key: 'FlippedTileScene' }) }

  preload() {
    this.textures.generate('stub', { data: ['1'], pixelWidth: 16 })
    // Format 1 = Phaser.Tilemaps.Formats.TILED_JSON
    this.cache.tilemap.add('flipped-map', { format: 1, data: TILEMAP_WITH_FLIPPED_TILE })
  }

  create() {
    const map = this.make.tilemap({ key: 'flipped-map' })
    const tileset = map.addTilesetImage('stub', 'stub')!
    const layer = map.createLayer('Ground', tileset)!
    layer.setCollisionFromCollisionGroup()
    try {
      this.matter.world.convertTilemapLayer(layer)  // crashes here
    } catch (e) {
      this.error = e as Error
    }
  }
}

let game: Phaser.Game | null = null
afterEach(() => { game?.destroy(true); game = null })

describe('Phaser bug: MatterTileBody crashes on flipped tiles', () => {
  it('throws TypeError on convertTilemapLayer when a colliding tile has flipX', async () => {
    game = new Phaser.Game({
      type: Phaser.CANVAS,
      width: 32, height: 16,
      physics: { default: 'matter', matter: { gravity: { x: 0, y: 0 } } },
      scene: [FlippedTileScene],
    })

    const scene = await new Promise<FlippedTileScene>((resolve) => {
      const check = setInterval(() => {
        const s = game!.scene.getScene('FlippedTileScene') as FlippedTileScene
        if (s && game!.scene.isActive('FlippedTileScene')) {
          clearInterval(check)
          resolve(s)
        }
      }, 50)
    })

    // Bug: crashes with "Cannot read properties of null (reading 'inertia')"
    expect(scene.error).not.toBeNull()
    expect(scene.error?.message).toMatch(/null|inertia/i)
  })
})
```

## Additional Information

**Proposed fix** — add one line to update the local `body` variable after the
body is created from tile collision shapes:

```js
if (!body) {
    var collisionGroup = tile.getCollisionGroup();
    var collisionObjects = GetFastValue(collisionGroup, 'objects', []);

    if (collisionObjects.length > 0) {
        this.setFromTileCollision(options);
    } else {
        this.setFromTileRectangle(options);
    }

    body = this.body;  // <-- add this line
}
```

**Workaround** (used until the fix lands): clear the collision flag on flipped
tiles before calling `convertTilemapLayer` so they are skipped by the converter:

```js
layer.setCollisionFromCollisionGroup()
layer.forEachTile(tile => {
    if (tile.collides && (tile.flipX || tile.flipY)) tile.setCollision(false)
})
this.matter.world.convertTilemapLayer(layer)
```

This avoids the crash at the cost of losing collision on flipped tiles.

I'd be happy to submit a pr, please let me know - I just want confirmation that I am not off-base about this issue at hand.

Thank you.

cyphercodes:
## Description

Fixes a scope bug in MatterTileBody where the local 'body' variable remained null after creating the body via setFromTileCollision() or setFromTileRectangle(). This caused a crash when processing flipped tiles because Body.scale() was called with null.

## Bug

When Matter.World.convertTilemapLayer() processes a tile that has flipX or flipY set, it crashes with:

## Root Cause

The constructor declares a local body variable initialised to null, then creates the actual body via setFromTileCollision() or setFromTileRectangle() — both of which set this.body but never update the local body variable. Immediately after, the flip check reads from the stale local.

## Fix

Add one line to update the local body variable after the body is created:

Fixes #7267

---

> [!NOTE]
> **Low Risk**
> Low risk, single-line fix in `MatterTileBody` constructor to prevent `null` body usage when scaling flipped tiles.
> 
> **Overview**
> Fixes a constructor scoping bug in `MatterTileBody` where the local `body` variable could remain `null` after creating a body from tile collision shapes or a tile rectangle.
> 
> After generating the body via `setFromTileCollision()` / `setFromTileRectangle()`, it now reassigns `body = this.body` so the subsequent flip handling (`Body.scale` for `flipX`/`flipY`) operates on the created body instead of crashing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 415a5fc38feff75c168fe61dfb3b45d0ef9754d8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/physics/matter-js/MatterTileBody.js
+++ b/src/physics/matter-js/MatterTileBody.js
@@ -113,6 +113,8 @@ var MatterTileBody = new Class({
             {
                 this.setFromTileRectangle(options);
             }
+
+            body = this.body;
         }
         else
         {


photonstorm:
Thanks, this has been merged into 4.0.0

PR #7278, opened 2026-04-12
Fix pointer position when canvas has CSS transforms (e.g. rotate)

### LINKED ISSUE #7175: Pointer position incorrect of canvas is rotated
## Version

* Phaser Version: 3.80, 4.0.0-rc.4

## Description

If the canvas is rotated, then the pointer positions will be calculated incorrectly. In the example below, the pointer should change to the hand cursor if and only if the pointer is over the green square, but it is not the case here.

## Example Test Code

```html
<!DOCTYPE` html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>My First Phaser Game</title>
    <style>
        html, body {
            width: 100%;
            height: 100%;
            margin: 0;
            padding: 0;
            background-color: #333;
        }

        #gameHolder {
            position: absolute;
            top: 100px;
            background: red;
            width: 600px;
            height: 500px;
            transform: rotate(90deg);
        }
    </style>
    <script src="https://cdn.jsdelivr.net/npm/phaser@4.0.0-rc.4/dist/phaser.min.js"></script>
    <script>

    class Example extends Phaser.Scene {
        create () {
            const rect = this.add.rectangle(50, 50, 300, 300, 0x00FF00, 1).setOrigin(0);

            rect.setInteractive({cursor:'pointer'});
            rect.on('pointermove', () => console.log('move'));
        }
    }

    const config = {
        type: Phaser.AUTO,
        parent: 'gameHolder',
        width:500,
        height:400,
        scene: Example
    };

    </script>
</head>
<body>
    <div id="gameHolder" ></div>
</body>
</html>
```

roy7:
## Summary

- Fixes incorrect pointer/input coordinates when the canvas or a parent element has a CSS `transform` (e.g. `rotate(90deg)` for forcing landscape in portrait-locked webviews)
- Adds `transformXY` method to ScaleManager that accounts for CSS rotation/scale/skew via `DOMMatrix`
- Updates `InputManager.transformPointer` to use the combined XY transform instead of separate `transformX`/`transformY` calls
- Zero overhead when no CSS transforms are present — falls back to the existing math

## Detail

Closes #7175.

The existing `transformX` and `transformY` methods convert pointer coordinates independently:

```js
transformX: function (pageX) {
    return (pageX - this.canvasBounds.left) * this.displayScale.x;
}
```

This breaks when the canvas has a CSS rotation because the X and Y axes become coupled — a screen-space X offset maps to both game X and Y. The simple subtract-and-scale approach can't account for this.

The fix adds `getInverseCSSTransform()` which walks the DOM tree from the canvas to the document root, composing any CSS `transform` values found along the way. It extracts the rotation/scale/skew portion (translation is already handled by `getBoundingClientRect`) and caches its inverse. The new `transformXY` method uses this inverse to convert screen-space pointer offsets back to the canvas's local coordinate system before scaling to game coordinates.

The inverse matrix is only recomputed in `updateBounds()` (on resize/orientation change), not on every pointer event.

## Test plan

- [x] Existing test suite passes (23,304 tests pass, 0 new failures)
- [x] Verified pointer accuracy with `transform: rotate(90deg)` on canvas parent (reproduction from #7175)
- [x] Verified no regression for non-transformed canvas (standard setup)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Touches core pointer coordinate transformation logic in `ScaleManager`/`InputManager`, which could impact all mouse/touch input; mitigation is the cached inverse matrix and explicit fallback to the prior math when no CSS transforms are present.
> 
> **Overview**
> Fixes incorrect pointer coordinates when the canvas (or an ancestor) has CSS transforms like rotation/skew by introducing a coupled `ScaleManager.transformXY` path.
> 
> `ScaleManager` now caches an inverse accumulated CSS transform matrix (via `DOMMatrix`) during `updateBounds()` and uses it to convert `(pageX,pageY)` into game-space coordinates; `InputManager.transformPointer` switches from separate `transformX`/`transformY` calls to this new combined transform, falling back to the existing offset+scale behavior when no CSS transforms are detected.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0625d1482cc8dae5ec3336480d94d0ebb3c5db86. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/input/InputManager.js
+++ b/src/input/InputManager.js
@@ -1034,9 +1034,10 @@ var InputManager = new Class({
         p1.x = p0.x;
         p1.y = p0.y;
 
-        //  Translate coordinates
-        var x = this.scaleManager.transformX(pageX);
-        var y = this.scaleManager.transformY(pageY);
+        //  Translate coordinates using transformXY to handle CSS transforms (rotation, skew)
+        var transformed = this.scaleManager.transformXY(pageX, pageY, this._tempPoint);
+        var x = transformed.x;
+        var y = transformed.y;
 
         var a = pointer.smoothFactor;
 
--- a/src/scale/ScaleManager.js
+++ b/src/scale/ScaleManager.js
@@ -271,6 +271,18 @@ var ScaleManager = new Class({
         this.displayScale = new Vector2(1, 1);
 
         /**
+         * The cached inverse of any CSS transforms applied to the canvas or its ancestors.
+         * Used to correctly convert pointer coordinates when the canvas has been CSS rotated or skewed.
+         * `null` if no CSS transforms are detected or DOMMatrix is unavailable.
+         *
+         * @name Phaser.Scale.ScaleManager#_cssTransformInverse
+         * @type {?DOMMatrix}
+         * @private
+         * @since 4.0.0
+         */
+        this._cssTransformInverse = null;
+
+        /**
          * If set, the canvas sizes will be automatically passed through Math.floor.
          * This results in rounded pixel display values, which is important for performance on legacy
          * and low powered devices, but at the cost of not achieving a 'perfect' fit in some browser windows.
@@ -1255,6 +1267,8 @@ var ScaleManager = new Class({
         bounds.y = clientRect.top + (window.pageYOffset || 0) - (document.documentElement.clientTop || 0);
         bounds.width = clientRect.width;
         bounds.height = clientRect.height;
+
+        this._cssTransformInverse = this.getInverseCSSTransform();
     },
 
     /**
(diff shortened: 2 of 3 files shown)


roy7:
I was thinking about a PR for a different issue I was having locally, but it seems it might be working as intended so decided to just use my local workaround and not report it as a bug. (Phaser's useHandCursor changes the cursor for all interactive objects under the pointer regardless of mask visibility or depth ordering, so masked/hidden objects still show a hand cursor. But the docs are clear masking is purely for graphics and not used for interactions. So I worked around it by manually checking for cursor changes instead of relying on useHandCursor.)

However when exploring known issues I saw this one and figured I'd let Claude Code take a crack at it. I did test locally that this fixes the bug in the example HTML provided in the original issue, and it passed the test suite, so I'm sending it in for your review in case you think it is worthwhile.

Thanks for reviewing!

roy7:
Pushed a follow-up commit addressing a few gaps found on re-review:

- **CSS longhand properties**: `getComputedStyle().transform` doesn't include the individual `rotate:` / `scale:` properties, so transforms written that way weren't detected. They're now composed in spec order (rotate → scale → transform). `translate:` is skipped since translation never affects the linear part of the matrix.
- **Degenerate transforms**: a non-invertible transform like `scale(0)` now falls back to the standard path instead of yielding NaN coordinates.
- **Docs**: corrected `@since` to `4.NEXT`, and documented that only 2D transforms are supported (3D rotations/`perspective` can't be inverted to a flat 2D mapping).
- **Tests**: added unit tests for `transformXY` (90° rotation, scale, and parity with `transformX`/`transformY` when no transform is present).

PR #7304, opened 2026-05-16
Fix lint configuration and apply fixes

niklas-e:
Linting was broken after recent eslint version upgrade from v8 to v10. See attached issue for more details.

Describe the changes below:

- Installed missing dependencies required by our linting setup
- Migrated `.eslintrc.json` and `.eslintignore` to `eslint.config.mjs`
- Applied `npm run lintfix` and fixed few remaining ones manually

Fixes #7303

Issue #7298, opened 2026-05-07
Investigate alternative libraries for type generation than `jsdoc`


Bertie690:
## Moving away
It has come to my attention that the `jsdoc` CLI that the `tsgen` parser relies on has [more or less died off](https://github.com/jsdoc/jsdoc/issues/1703#issuecomment-1175903032) given its commit history has effectively stopped. 
With the only activity being dependency version bumps, it appears there is little hope of the project adding support for things like [tuples](https://github.com/jsdoc/jsdoc/issues/1703) or [negative numbers](https://github.com/jsdoc/jsdoc/issues/2131), both of which serve as major limitations towards documenting Phaser's own API. (For instance, we physically cannot type `RandomDataGenerator#sign` with the 2 values it is contractually obligated to return as `jsdoc` will crash when trying to parse the union `-1 | 1`).

Unless a major revival is launched in the near future (which it likely won't be), it may be wise to start thinking of alternate means of generating type documentation long-term. (I'd **strongly** advise against an in-house solution as the current one is ALREADY incomprehensible as is.)

I have created this issue as a "discussion board" of sorts to discuss various options for moving towards a more functional (and ideally better-maintained) solution for typegen that doesn't curb expressivity.

> [!NOTE]
> If the repository is ever converted to pure TypeScript OR plain JavaScript (by which I mean actual ESM classes and not "fake" class-building functions), `tsc` would likely be the best means of generating types for us.

niklas-e:
Good point! I've always wondered why projects have been declaring types in any other way than native TypeScript. It is just generally fragile. Anyways, I could contribute to Phaser TS migration. I think it doesn't even need to be all-in migration at first, because you can configure TypeScript to allow JavaScript modules.

SamiSaves:
Phaser to TypeScript sounds to me like the natural answer to this. Has there been any earlier discussion or decisions about TypeScript conversion?

Bertie690:
> Has there been any earlier discussion or decisions about TypeScript conversion?

Not to my knowledge.
#7241 did mention it in passing (insofar as it requested a typing change that is fundamentally incompatible with JSDoc), but no existing discussions have been launched in recent memory.

Bertie690:
[...]

PR #7308, opened 2026-05-24
Feat: foundation for TypeScript migration

niklas-e:
> Disclaimer: this contains the eslint config fix commit from #7304, to enable proper lint configuration for TS

## Introduction

This PR starts an incremental JavaScript to TypeScript migration path for Phaser. I acknowledge this will be an epic journey if we go on this path.

The motivation is related to #7298: the current declaration pipeline still depends heavily on `jsdoc`, which has known parser limitations and an uncertain maintenance future. Rather than proposing an all-at-once rewrite, this branch shows that Phaser can migrate individual modules to native TypeScript while keeping the existing source tree, build output, and generated `types/phaser.d.ts` workflow intact.

The migrated modules are intentionally small but varied, covering math helpers, a class, mixins, a struct, geometry, and a game object helper. They are now discovered automatically from migrated TypeScript source via JSDoc namespace tags, and the type generation step overlays authoritative TypeScript declarations for those migrated symbols into the existing Phaser declaration file.

## Pros

- Allows gradual migration from JS to TS without requiring the whole codebase to move at once
- Keeps existing JavaScript modules working alongside migrated TypeScript modules
- Uses `tsc` as the source of truth for migrated declarations, avoiding some `jsdoc` parser limitations
- Gives migrated code real compiler checking instead of relying only on JSDoc annotations
- Preserves the existing public `types/phaser.d.ts` output shape for consumers
- Adds validation around migrated symbols so missing or incorrect generated declarations are caught
- Provides a practical foundation for continuing module-by-module migration
- When whole migration is completed, .d.ts maps can be published which enables navigation to source definition

## Cons / Trade-offs

- The type generation pipeline becomes more complex while Phaser is in a hybrid JS/TS state
- Contributors need to understand both existing JSDoc-based type generation and the new TypeScript overlay path

### How distribution artifacts are stitched together

The published `phaser.d.ts` remains a single output file, but during the migration period it is assembled from `tsgen` output and an overlay of declarations emitted by `tsc` for migrated modules.

```mermaid
flowchart TD
    jsFiles["JSDoc (.js files)"] --> jsdoc["jsdoc parser"] --> dtsDom["dts-dom tree"]
    tsFiles["TypeScript (.ts files)"] --> discovery["auto-discovery (JSDoc tags)"] --> tsc["tsc --emitDeclarationOnly"]

    dtsDom --> overlay["MigratedOverlay<br/>(synthetic stubs + overlay)"]
    tsc --> overlay

    overlay --> publish["publish.ts<br/>(orchestrator)"]
    publish --> phaser["phaser.d.ts"]
```

### Phases of the migration

1. Start with the tsgen + overlay strategy shown above
2. Continue using `npm run ts` to update and validate the type declarations during the migration period. Once migrated modules reach 50% of source modules, it will also start warning that it may be time to consider a TS-first declaration pipeline. No concrete plan yet, but at that point most of `phaser.d.ts` will be generated by `tsc` and we should patch JSDoc definitions for the remaining JavaScript modules on top of that
3. Once the codebase is fully migrated to TypeScript, `tsgen` can be removed as declarations will be emitted by `tsc`

## Agent skill for migrating modules

I wrote and tested an agent skill to help with the migration. This should be a great aid for contributors. You can find it here: https://github.com/niklas-e/phaser-typescript

## TODO

- [ ] Remove dist artifacts before merging (included for easier testing within actual projects)
- [ ] Setup an umbrella issue and/or GH project to track the progress
- [ ] Ensure docs.phaser.io works as expected - did not find the sources for it so couldn't see how it is set up
- [ ] Write a contribution guide


src/structs/Map.ts
```diff
@@ -0,0 +1,331 @@
...
+ * Unlike a native JavaScript Map, it also maintains an internal object of entries for efficient
+ * keyed access and iteration. Supports filtering, merging, and contains/size operations.
+ * Used internally by various Phaser systems for managing collections.
+ *
+ * ```javascript
+ * var map = new Map([
+ *    [ 1, 'one' ],
+ *    [ 2, 'two' ],
+ *    [ 3, 'three' ]
+ * ]);
```
niklas-e, line 29: JSDoc version allowed numbers for keys, but it is confusing while using an object for storage internally. In JavaScript this will be true `obj[1] === obj["1"]`, thus causing potential runtime bugs. Therefore I set the signature to `Map<K extends string = string, V = unknown>`.

src/structs/Map.ts
```diff
@@ -0,0 +1,331 @@
...
+
+import { objectKeys } from '../utils/object/TypedObjectUtils.js';
+
+/**
+ * @param key - The key of the Map entry.
+ * @param entry - The value of the Map entry.
+ *
+ * @returns The callback result.
+ */
+type EachMapCallback<K extends string, V> = (key: K, entry: V) => boolean | void;
```
niklas-e, line 15: JSDoc declaration said this callback returns null when in reality it returns `entries[key]`, not null. The type was misleading.

tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
+    "compilerOptions": {
+        "strict": true,
+        "target": "ES2018",
```
niklas-e, line 4: What would be a comfortable target without dropping too much compatibility? I understand there's probably varying views, but my point of view is that as a framework Phaser could ship builds with quite modern versions and people who want to support really old stuff can use existing tools in the ecosystem to convert to their preferred ES version. --- Also good to note we should definitely set the target to at least ES2015 to not have the distributables include bunch of utility code for e.g. classes

src/utils/Mixin.ts
```diff
@@ -0,0 +1,119 @@
...
+  Base: TBase
+) => TBase & Constructor<InstanceType<TBase> & AddedByAll<TMixins>>
+{
+    // The reduce chain is type-safe at call sites via the overload signature;
+    // the implementation needs a single cast to bridge the generic gap.
+    return ((Base: Constructor) =>
+        mixins.reduce((Current, mixin) => mixin(Current), Base)
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    ) as any;
+}
```
niklas-e, line 119: There's multiple ways to implement mixins, but this composable implementation keeps the footprint same size as before.

For comparison the worst case scenario with [TS handbook](https://www.typescriptlang.org/docs/handbook/mixins.html) mixins in current codebase would be:

```
// Note: creates many anonymous classes
const SpriteBase =
  Alpha(
  BlendMode(
  Depth(
  Flip(
  GetBounds(
  Lighting(
  Mask(
  Origin(
  RenderNodes(
  ScrollFactor(
  Size(
  TextureCrop(
  Tint(
  Transform(
  Visible(
    GameObject
  )))))))))))))));

export class Sprite extends SpriteBase { ... }
```

But with the composable approach it would be:

```
// Note: creates only single class
const SpriteBase = composeMixins(
  Alpha,
  BlendMode,
  Depth,
  Flip,
  GetBounds,
  Lighting,
  Mask,
  Origin,
  RenderNodes,
  ScrollFactor,
  Size,
  TextureCrop,
  Tint,
  Transform,
  Visible)(GameObject);

export class Sprite extends SpriteBase { ... }
```

src/utils/object/TypedObjectUtils.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Type-safe wrapper around `Object.keys()` that preserves the key type.
+ *
+ * The built-in `Object.keys()` always returns `string[]`, which loses
+ * generic key information. This helper returns `K[]` instead, making it
+ * safe to use the result to index back into the same record without casts.
+ *
+ *
```
Bertie690, line 8: I will mention that such a type-preserving `Object.keys` method should probably have a disclaimer about it being _technically_ unsound due to structural typing allowing excess properties.
It's fine if the object in question is effectively nominal (i.e. we know it won't contain any extra keys), but ~~should~~ **MUST** be avoided for anything where structural typing could possibly become involved.

(Also, we should absolutely not export this thing.)

src/utils/Mixin.ts
```diff
@@ -0,0 +1,119 @@
+/**
+ * A generic constructor type. Used as a constraint for mixin base classes.
+ *
+ * The `any[]` for args is the standard TypeScript mixin pattern — required
+ * because `unknown[]` breaks assignability of concrete constructors like
+ * `new (scene: Scene, type: string) => T` to the generic constraint.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Constructor<T = object> = new (...args: any[]) => T;
```
Bertie690, line 9: **If** this ever gets needed somewhere else later down the line, we could potentially improve this further:
(Improvement copied straight from `type-fest`)
```ts
type Constructor<T, Arguments extends unknown[] = any[]> = new (...arguments_: Arguments) => T
```

(As an aside, it may be wise to look into adding `type-fest` or a similar library as a dev dependency - they've got some very useful type utilities with much better edge-case support than your average hand-rolled versions.)

src/utils/Mixin.ts
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Type guard: returns true when `value` is a getter/setter descriptor object.
+ */
+function isAccessorDescriptor (value: unknown): value is AccessorDescriptor
+{
+    if (value === null || typeof value !== 'object') { return false; }
+
+    const record = value as Record<string, unknown>;
+
+    return typeof record.get === 'function' || typeof record.set === 'function';
```
Bertie690, line 30:
```suggestion
    return typeof (value as Record<string, unknown>).get === 'function' || typeof (value as Record<string, unknown>).set === 'function';
```

IMO, we should generally avoid creating temp variables for the compiler's sake where possible (type assertions are made for these kinds of things)

photonstorm:
Just to mention that we are reading this, and are interested in doing it, but we have no capacity for such an undertaking for quite a while. At least until the end of August. Even if the community (thank you!) and let's face it, Claude, does the majority of the work, this change would have multiple tooling ramifications for us, well beyond simply updating the library and pushing to npm. Even without changing the API surface, this change would be profound.

Also, if we're going to go down this path, we will almost certainly do it as part of a v5.0 release, to give us the chance to break the API in significant ways (i.e., to avoid the use of mixins).

scripts/tsgen/bin/Parser.js
Bertie690, line 1: Man, we _really_ shouldn't track the generated `js` file and sourcemap looool

photonstorm: It makes literally no difference to do so.

Bertie690: It increases the number of touched files per changeset and makes cloning the repo take more time.

Not necessarily a massive issue _per se_, but still far from nothing.

It's effectively the same argument for including autogenerated IDE files inside `gitignore`, albeit on a much smaller scale.

photonstorm: It's 4 files with a combined size of less than 49KB, so about as close to nothing as you could get in the grand scheme of things.

tsconfig.json
DayKev: ES2022 seems to be fairly safe to target (even Chromebooks support it), though if you want to be extra safe ES2021 or ES2020 should be good.

ES2023 is where you'll start sacrificing some meaningful compatibility with actual in-use devices I think; we started getting reports from Chromebook users being unable to play when we tried to move from ES2022 to ES2023 (though I don't know if it was only a small subset of Chromebooks or the majority that didn't support it since we quickly dropped back to ES2022 with polyfills for the ES2023 features we wanted to use).

src/structs/Map.ts
```diff
@@ -0,0 +1,331 @@
...
+     * Returns a boolean indicating whether an element with the specified key exists or not.
+     *
+     * @since 3.0.0
+     *
+     * @param key - The key of the element to test for presence of in this Map.
+     * @returns Returns `true` if an element with the specified key exists in this Map, otherwise `false`.
+     */
+    has (key: K): boolean
+    {
+        return (this.entries.hasOwnProperty(key));
```
Bertie690, line 158: If we upgrade further, we could (and arguably SHOULD) use `Object.hasOwn`; this breaks if you happen to add a string `hasOwnProperty` to the map.

src/structs/Map.ts
```diff
@@ -0,0 +1,331 @@
...
+     * Adds all the elements in the given array to this Map.
+     *
+     * If the key already exists, the value will be replaced.
+     *
+     * @since 3.70.0
+     *
+     * @param elements - An array of key-value pairs to populate this Map with.
+     * @returns This Map object.
+     */
+    setAll (elements?: Array<[K, V]> | null): this
```
Bertie690, line 74: From a pure correctness standpoint, I'd rather not have functions be able to take arguments that do literally nothing for them.
Worst comes to worst, some nullish coaclescing works out equivalently while preventing dumb no-ops like `setAll(null)`.

(This principle can likely be extended to other functions as well.)

src/structs/Map.ts
```diff
@@ -0,0 +1,331 @@
...
+        this.entries = {} as Record<K, V>;
+        this.size = 0;
+
+        this.setAll(elements);
+    }
+
+    /**
+     * Adds all the elements in the given array to this Map.
+     *
+     * If the key already exists, the value will be replaced.
```
Bertie690, line 67:
```suggestion
     * If any of the keys already exist, their values will be replaced.
```

plurals

src/structs/Map.ts
```diff
@@ -0,0 +1,331 @@
...
+     * @since 3.0.0
+     */
+    size: number;
+
+    /**
+     * @param elements - An optional array of key-value pairs to populate this Map with.
+     */
+    constructor (elements?: Array<[K, V]> | null)
+    {
+        this.entries = {} as Record<K, V>;
```
Bertie690, line 58: This should be a `null`-prototype object if or when we are able to use `Object.hasOwn`

src/structs/Map.ts
```diff
@@ -0,0 +1,331 @@
...
+     * @since 3.0.0
+     *
+     * @returns This Map object.
+     */
+    clear (): this
+    {
+        for (const prop of objectKeys(this.entries))
+        {
+            delete this.entries[prop];
+        }
```
Bertie690, line 192: Could this just set the object to `{}`?
That'd be a lot faster if we aren't doing weird stuff.

src/structs/Map.ts
```diff
@@ -0,0 +1,331 @@
...
+ */
+export class Map<K extends string = string, V = unknown>
+{
+    /**
+     * The entries in this Map.
+     *
+     * @default {}
+     * @since 3.0.0
+     */
+    entries: Record<K, V>;
```
Bertie690, line 43: Do we even need to expose this? If you need direct access to the map's contents, that's almost certainly a sign that you don't want to be using a built-in utility struct.

src/gameobjects/components/Depth.ts
```diff
@@ -0,0 +1,242 @@
...
+
+    /**
+     * Private internal value. Holds the depth of the Game Object.
+     *
+     * @name Phaser.GameObjects.Components.Depth#_depth
+     * @private
+     * @default 0
+     * @since 3.0.0
+     */
+    _depth: number;
```
Bertie690, line 27: we can mark this as private in the TS file

and should

Bertie690 (commented):
Random things I noticed

src/structs/Map.ts
DayKev: According to the v3-to-v4 migration guide, `Phaser.Struct.Map` and `.Set` have been replaced with JS's built-in versions, so I don't think this file even needs to be ported to TS (and since it seems like the Phaser devs want to make a TS port a major version bump, it should definitely just be fully dropped if it's already unused in v4).

image

photonstorm: Yes, this.

PR #7311, opened 2026-05-31
add Tileset#animating

sylvainpolletvillard:

* Adds a new feature

Describe the changes below:

Adds a new field to Tileset class, Tileset.animating ; boolean ; whether the tileset animations should play or not. If false, animated tiles will stay on their first frame.

Phaser 4 supporting animated tilesets natively is very nice, but being able to propose an option to players to disable those background animations is important for accessibility reasons. I did not find a way to easily do that with current API so I added a field for that.


sylvainpolletvillard:
Just found the existing API I was looking for:
```
tilemap.layers.forEach(layer => {
   layer.tilemapLayer.setTimerPaused(true)
})
```

So now new feature is needed, I can close this

PR #7312, opened 2026-06-02
Add Cone Lights

FSDevelop:
I've been waiting for Phaser 4.0 to include on top of it the cone lights that I needed for the flashlights on my game. I wonder if this is something that may be of interest for phaser and other developers.

Screenshot 2026-06-02 213841


photonstorm:
The `if` statement in the shader is a little worrying (they are expensive per-pixel ops), but otherwise this PR is lovely and well put together. Could you please change `@since 4.1.0` to `@since 4.2.0` in your code, and remove the TypeScript defs from the commit, too. Then we can easily merge it in without having to modify it afterward.

FSDevelop:
Thanks @photonstorm , I made the changes. Related to the shader, I can remove the if to avoid the evaluation per pixel, but then the math would run anyways. A potential solution could be a variation, but perhaps is too much duplication?

photonstorm:
It should be fine, the condition is uniform-ish per light, so GPU divergence should be limited. We'll merge this for the 4.2.0 release, thanks.

PR #7319, opened 2026-06-19
Fix setSmoothPixelArt call

scobo:
This PR

* Fixes a bug

Describe the changes below:

Fixes a runtime issue when attempting to use `setSmoothPixelArt`:

> Uncaught ReferenceError: Phaser is not defined

The `setSmoothPixelArt` method calls `this.setFilter(Phaser.Textures.FilterMode.LINEAR)` but `Phaser` is no longer global in Phaser 4. This PR imports the `FilterMode` constant instead.


BenjaminDRichards:
Thanks for spotting this! We double checked our code, and found a couple of other places where a global namespace was accessed. We combined those into one big fix (https://github.com/phaserjs/phaser/commit/b9410323a98f019ed2c11c4088c015743051cd96), and credited you in the changelog.

I'll go ahead and close this PR, but we're using all its changes.

PR #7328, opened 2026-06-27
Docs: correct setSizeToFrame()

### LINKED ISSUE #7325: `Size.setSizeToFrame` should mention default behaviour w/o frame; make parameter type optional
## Version
* Phaser Version: 4.2.1
* Operating system: N/A
* Browser: N/A

## Description
The `setSizeToFrame` method from the `Size` mixin claims to accept a boolean for the `frame` parameter, while leaving no mention as to what passing a boolean would even _do_.

Moreover, attempting to pass `true` (a valid boolean) to the function will set the object's width/height to `undefined` and likely cause a crash the next time said fields are used.

To fix this, I propose 2 solutions (all of which can likely be done in tandem):
1. The function should document the fact that it defaults to the current frame if the `frame` parameter is omitted. 
2. The function's typing should be altered to change `boolean` to `undefined` and/or make the parameter entirely optional.

## Example Test Code
```ts
sprite.setSizeToFrame(false); // non obvious, uses current frame
sprite.setSizeToFrame(true); // non obvious and crashes next time sprite is used
sprite.setSizeToFrame(); // would be most ergonomic and works at runtime, but gives type errors
```

## Additional Information
N/A

samme:
This PR

* Updates the Documentation

Fixes #7325

### DIFF
--- a/src/gameobjects/components/Size.js
+++ b/src/gameobjects/components/Size.js
@@ -110,20 +110,21 @@ var Size = {
     },
 
     /**
-     * Sets the size of this Game Object to be that of the given Frame.
+     * Sets the size of this Game Object to be that of the given Frame or the current Frame.
      *
      * This will not change the size that the Game Object is rendered in-game.
      * For that you need to either set the scale of the Game Object (`setScale`) or call the
      * `setDisplaySize` method, which is the same thing as changing the scale but allows you
      * to do so by giving pixel values.
      *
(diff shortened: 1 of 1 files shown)


Bertie690:
You should probably regenerate `phaser.d.ts`

samme:
It will get rebuilt right before release. 👍

photonstorm:
Yup, never submit the built defs, please. It's pointless; our build step generates them.

